PR #54, opened 2020-07-09
Use blackbox_exporter with probe_dns_duration_seconds

mem:
The forked version of blackbox_exporter exposes a metric called
"probe_dns_duration_seconds", which is the time it takes for the DNS
check to run, from connection to query and getting results back.

Currently SM is relying on probe_dns_lookup_time_seconds, which is the
wrong metric for this check, as that is reporting the time it takes to
resolve the _target_, which is the DNS *server*. Since we normally pass
an IP address as the target, that resolution time is very low
(microseconds).

The change in blackbox_exporter is in this PR:

https://github.com/prometheus/blackbox_exporter/pull/662


woodsaj:
Given this change was merged upstream, looks like we should just vendor in `github.com/prometheus/blackbox_exporter@55c667ab4ce5bdd81ebf8b843851222aa1f7fab3` (current master)

PR #55, opened 2020-07-16
Validate IP addresses / hostnames in targets

mem:
Ping, DNS and TCP checks should accept only IP addresses or fully
qualified hostnames as targets.

### DIFF
--- a/pkg/pb/synthetic_monitoring/checks_extra.go
+++ b/pkg/pb/synthetic_monitoring/checks_extra.go
@@ -18,6 +18,7 @@ package synthetic_monitoring
 
 import (
 	"errors"
+	"net"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -141,6 +157,10 @@ func (c *Check) Validate() error {
 	if c.Settings.Tcp != nil {
 		settingsCount++
 
+		if err := validateHost(c.Target); err != nil {
```
woodsaj, line 160: for TCP, the target should be `host:port` , So we need to split into host and target first, then validate each.  ie port must be > 0 and <= 65535

mem: I completely missed that, thanks!

Fixed

PR #65, opened 2020-08-26
Update CircleCI config to build and deploy tagged commits

mem:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,13 @@ jobs:
       - image: circleci/golang:1.13.10
     steps:
       - checkout
+      - run: make build
+
+  deploy:
+    docker:
+      - image: circleci/golang:1.13.10
+    steps:
+      - checkout
       - setup_remote_docker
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,22 +35,72 @@ jobs:
...
           command: |
-            if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              docker login -u $DOCKER_USER -p "$DOCKER_PASS"
-              make docker-push
-              export GCS_KEY_DIR=$(pwd)/keys
-              mkdir -p $GCS_KEY_DIR
-              echo "$GCS_KEY" > $GCS_KEY_DIR/gcs-key.json
-              make publish-packages
-            fi
+            docker login -u $DOCKER_USER -p "$DOCKER_PASS"
```
brittenhouse, line 38: What was the reason we needed to remove the check for being in the master branch?  It seems this would push all branch builds to our package repo which may not be what we want.

PR #72, opened 2020-09-03
Move build from circleCI to drone

captncraig:
This moves all builds from circleCI to drone. On pull request does all lint and build steps.

On push to master it will do all that, as well as publishing docker image and packages.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -66,8 +66,9 @@ local vault_secret(name, vault_path, key) = {
       + {environment: {
           GCS_KEY:{from_secret: 'gcs_key'},
           GPG_PRIV_KEY:{from_secret: 'gpg_priv_key'},
+          PUBLISH_PROD_PKGS: "1",
         }}
-      ,
+      + masterOnly,
   ]),
 
   vault_secret('docker_username','infra/data/ci/docker_hub', 'username'),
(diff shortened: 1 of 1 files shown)


captncraig:
I tested the docker and package publishing. All credentials seem to be working.

We should probably disable the circleCI jobs before merging this.

.drone.jsonnet
```diff
@@ -0,0 +1,81 @@
...
+      './scripts/version',
+      './scripts/version > .tags', // save version in special file for docker plugin
+      'make build',
+    ]),
+    // We can't use 'make docker' without making this repo priveleged in drone
+    // so we will use the native docker plugin instead for security.
+    step('docker build',[],'plugins/docker')+{
+      settings:{
+        repo: repo,
+        dry_run: 'false',
```
captncraig, line 52: @mem my intention was that the docker build step could run on all prs and things, and the docker push step would only run on master. By setting dry_run:false here, you will be pushing to docker hub on each pr build and stuff. Is that what you intended?

mem: no, I misread the config and thought this section applied to master. Reverted.

But it did fail and I was expecting it not to. Am I missing something else?

mem: I found what I was missing, sorry about the noise!

PR #73, opened 2020-09-08
Feature: Implement test to check metric changes

mem:

### DIFF
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -37,10 +37,10 @@ var (
 	staleMarker float64 = math.Float64frombits(staleNaN)
 
 	probers = map[string]prober.ProbeFn{
-		"http": prober.ProbeHTTP,
-		"tcp":  prober.ProbeTCP,
-		"icmp": prober.ProbeICMP,
-		"dns":  prober.ProbeDNS,
+		ScraperTypeHTTP: prober.ProbeHTTP,
+		ScraperTypeTcp:  prober.ProbeTCP,
+		ScraperTypePing: prober.ProbeICMP,
+		ScraperTypeDNS:  prober.ProbeDNS,
 	}
 )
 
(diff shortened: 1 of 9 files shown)


mem:
Sorry about the large, noisy change.

The goal here is to have a test that fails if the BBE metrics change. This can easily go by unnoticed, so the approach here is to run the actual probe against local targets (set up in the test) and collect the resulting metrics.

This required some reorganization of the code in the scraper, so that it's possible to get a slice of `prometheus.MetricFamily` that can be used to present the metrics using the exposition format. It's very likely that this code can be improved, but I tried hard to not change functionality.

The changes also open the path towards pushing metadata to prometheus, so that the metric documentation can be presented to the user in Grafana.

internal/scraper/scraper.go
```diff
@@ -427,7 +383,116 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, erro
...
+		Name: "probe_duration_seconds",
+		Help: "Returns how long the probe took to complete in seconds",
+	})
+	smCheckInfo := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name:        "sm_check_info",
+		Help:        "Provides information about a single check configuration",
+		ConstLabels: checkInfoLabels,
+	})
+
+	registry.MustRegister(probeSuccessGauge)
```
electron0zero, line 441: as per docs 
> MustRegister panics if an error occurs

IIUC this will crash the agent process, do we want to panic in this case?

internal/scraper/scraper_test.go
```diff
@@ -0,0 +1,457 @@
...
+			},
+		},
+
+		"http_ssl": {
+			prober: prober.ProbeHTTP,
+			setup: func(*testing.T) (string, func()) {
+				httpSrv := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+					w.WriteHeader(http.StatusOK)
+				}))
+				// httpSrv.EnableHTTP2 = true
```
electron0zero, line 82: are we not checking for HTTP2?

internal/scraper/scraper_test.go
```diff
@@ -0,0 +1,457 @@
...
+							InsecureSkipVerify: true,
+						},
+					},
+				},
+			},
+		},
+
+		"dns": {
+			prober: prober.ProbeDNS,
+			setup: func(*testing.T) (string, func()) {
```
electron0zero, line 103: nit: can we move all this logic into a function, and call that instead of doing it in a anon func?

it would make it easy to read

internal/scraper/scraper_test.go
```diff
@@ -0,0 +1,457 @@
...
+					IPProtocolFallback: true,
+					QueryName:          "example.com",
+					TransportProtocol:  "udp",
+				},
+			},
+		},
+
+		"tcp": {
+			prober: prober.ProbeTCP,
+			setup: func(t *testing.T) (string, func()) {
```
electron0zero, line 134: nit: same here, let's move all this logic into a function.

internal/scraper/scraper_test.go
```diff
@@ -0,0 +1,457 @@
...
+				Timeout: 2 * time.Second,
+				TCP: bbeconfig.TCPProbe{
+					IPProtocol:         "ip4",
+					IPProtocolFallback: true,
+				},
+			},
+		},
+
+		"tcp_ssl": {
+			prober: prober.ProbeTCP,
```
electron0zero, line 172: nit: let's move this as well, and maybe we can reuse the tcp setup method

internal/scraper/scraper.go
```diff
@@ -37,10 +37,10 @@ var (
 	staleMarker float64 = math.Float64frombits(staleNaN)
 
 	probers = map[string]prober.ProbeFn{
-		"http": prober.ProbeHTTP,
-		"tcp":  prober.ProbeTCP,
-		"icmp": prober.ProbeICMP,
-		"dns":  prober.ProbeDNS,
+		ScraperTypeHTTP: prober.ProbeHTTP,
```
electron0zero, line 40: nice :clap:

electron0zero (commented):
love this change, will ease our BBE upgrades :clap:

internal/scraper/scraper.go
mem: I think it was like that before. The reasoning probably is that it won't panic because we know that that metric can be added: basically that the metric hasn't been registered before, that the labels are consistent, etc. If it ever panics, we will catch it during testing, never during production.

internal/scraper/scraper_test.go
mem: The resulting metrics are the same. `probe_http_version` changes value, but the metric itself is the same.

internal/scraper/scraper_test.go
mem: OK, done.

internal/scraper/scraper_test.go
mem: done

internal/scraper/scraper_test.go
mem: done

electron0zero (approved):
lgtm :clap: :shipit:

PR #79, opened 2020-09-21
Feature: report probe version to API

mem:
For tracking and notification purposes, the probe is going to be
reporting its version (as computed at buildtime), commit and buildstamp.

The API can use this information to notify users of old probes, changes
in metrics and required upgrades.


electron0zero (approved):
lgtm :100: :shipit:

PR #100, opened 2020-11-25
Fix: validate check and probe labels

### LINKED ISSUE #99: agent accepts invalid label names
Prometheus is very strict about the labels it accepts, basically they should match `/^[a-zA-Z][a-zA-Z0-9]*$/`.

We are not strict enough with the labels we are accepting, and that causes Prometheus to reject them. This should be fixed both in the backend code and the front end plugin.

mem:
We are not validating the labels that we are going to apply to the
metrics, and we eventually hit a call to MustRegister that causes the
probes to crash with bad metrics.

Add validation code for labels, that gets called from checks and probes.
This is going to cause the check to be rejected if it makes it to the
probe, but the intention is that it's rejected in the API or the
frontend, where it can be handled more gracefully. The API is going to
inherit this validation.

Also add TestCheckValidate to make sure we are calling these validations
for incoming checks.

Partially fixes #99 (because of required API and frontend fixes).

### DIFF
--- a/pkg/pb/synthetic_monitoring/checks_extra.go
+++ b/pkg/pb/synthetic_monitoring/checks_extra.go
@@ -30,9 +30,12 @@ var (
 	ErrInvalidCheckTarget     = errors.New("invalid check target")
 	ErrInvalidCheckJob        = errors.New("invalid check job")
 	ErrInvalidCheckFrequency  = errors.New("invalid check frequency")
+	ErrInvalidCheckTimeout    = errors.New("invalid check timeout")
 	ErrInvalidCheckLabelName  = errors.New("invalid check label name")
 	ErrTooManyCheckLabels     = errors.New("too many check labels")
 	ErrInvalidCheckLabelValue = errors.New("invalid check label value")
+	ErrInvalidLabelName       = errors.New("invalid label name")
+	ErrInvalidLabelValue      = errors.New("invalid label value")
 
 	ErrInvalidCheckSettings = errors.New("invalid check settings")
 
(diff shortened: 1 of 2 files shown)


pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -97,20 +100,16 @@ func (c *Check) Validate() error {
 
 	// timeout must be in [1, 10] seconds
 	if c.Timeout < 1*1000 || c.Timeout > 10*1000 {
-		return ErrInvalidCheckFrequency
+		return ErrInvalidCheckTimeout
```
electron0zero, line 103: we should also validate that check frequency is grater then check timeout.

If we allow check frequency to be less then timeout, we would end up running concurrent checks for a target.

issue: https://github.com/grafana/synthetic-monitoring-api/issues/95

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -236,6 +231,27 @@ func (p *Probe) Validate() error {
...
+	if len(l.Name) == 0 {
+		return ErrInvalidLabelName
+	}
+
+	// This bit is lifted from Prometheus code, except that
+	// Prometheus accepts /^[a-zA-Z_][a-zA-Z0-9_]*$/ and we accept
+	// /^[a-zA-Z0-9_]+$/ because these names are going to be
+	// prefixed with "label_".
+	for _, b := range l.Name {
+		if !((b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || (b >= '0' && b <= '9')) {
```
electron0zero, line 244: curious to know why we didn't use https://golang.org/pkg/regexp/ here.

pkg/pb/synthetic_monitoring/checks_extra_test.go
```diff
@@ -281,6 +352,57 @@ func TestValidateHostPort(t *testing.T) {
 	}
 }
 
+func TestValidateLabel(t *testing.T) {
```
electron0zero, line 355: :rocket:

pkg/pb/synthetic_monitoring/checks_extra.go
mem: thanks, I'll include that here

pkg/pb/synthetic_monitoring/checks_extra.go
mem: According to Prometheus code it's for performance reasons. I doubt it makes a difference here, they probably run this bit of code many times, we only do it when checks are added.

PR #104, opened 2020-12-15
Feature: add option to reduce the number of published metrics

mem:
Checks derive a number of histograms that can be used to easily identify
for example the 95 percentile for a metric. Add an option to make this
derived metrics optional. To prevent removing metrics that users might
alredy be using, make this opt-out by defaulting to all metrics.


electron0zero (approved):
lgtm :shipit:

PR #107, opened 2020-12-17
Communicate tenant info in changes

mem:
When then tenant remote information changes, the API will communicate
this to the probes in the changes stream.

The GetTenant service is still needed, because there might be changes
happening outside the API (for example, the user deleted a token), so
the probes still need to be able to query the tenant information on
their own.

Changing the name of the protocol buffer message is backwards
compatible.


mem:
I verified this is working with the existing API

mem:
> I verified this is working with the existing API

Correction: I verified this before the very last change.

internal/pusher/pusher.go
```diff
@@ -208,37 +222,51 @@ func (p *Publisher) getClient(ctx context.Context, tenantId int64, newClient boo
...
 		Metrics: mClient,
 		Events:  eClient,
 	}
+
 	p.clientsMutex.Lock()
-	p.clients[tenantId] = clients
+	p.clients[tenant.Id] = clients
 	p.clientsMutex.Unlock()
 
+	p.logger.Debug().Interface("tenant", tenant).Msg("updated client")
```
electron0zero, line 268: let's log tenantId as well

electron0zero (approved):
few nits, rest LGTM :+1:

PR #110, opened 2021-01-14
Add alert sensitivity to check info metric

rdubrock:


mem (commented):
Looks good, thanks.

Are we happy with the `alertSensitivity` name?

rdubrock:
> Are we happy with the `alertSensitivity` name?

I'm not super jazzed about it, because I think the user expectation for alert levels is more oriented around "severity", but I think what we actually have control over is sensitivity, not severity. Definitely open to other ideas.

PR #122, opened 2021-02-19
Migrate to Go 1.16, update tools

mem:
There are enough improvements from 1.14 to 1.16 to make it worth
upgrading. Update the tools (golangci-lint, gotestsum, gosec), too.

Worth repeating: the version listed in go.mod does NOT cause the
compiler to error out if it's older than the version listed, but it will
error if the code uses features from that version (and the compiler is
older). Right now the code is not using any newer versions.


electron0zero (approved):
lgtm :+1:

PR #125, opened 2021-02-23
Cache and queue tenant requests in the agent

mem:
To avoid spamming the API server with requests for tenant information,
cache it in the agent for a short time. This is also causing the
requests to be queued, since GetTenant is issued with the lock held.

### DIFF
--- a/internal/pusher/tenant_manager.go
+++ b/internal/pusher/tenant_manager.go
@@ -1,0 +1,120 @@
+package pusher
+
+import (
+	"context"
+	"sync"
+	"time"
+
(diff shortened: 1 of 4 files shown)


internal/pusher/tenant_manager.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if found && info.validUntil.After(now) {
+		return info.tenant, nil
+	}
+
+	// either not found or too old
+
+	tm.tenantMutex.Lock()
+	info, found = tm.tenants[in.Id]
+	defer tm.tenantMutex.Unlock()
```
mem, line 45: this bit needs some thinking...

PR #126, opened 2021-02-24
Feature: add method to report check type

mem:
The checktype is determined by the settings field, as exactly one field
can be set. Even so, it's useful to be able to ask for the check type
without reimplementing logic that knows about this quirk.


electron0zero (approved):
lgtm :+1:

PR #127, opened 2021-02-26
Feature: provide number of active series per check type

mem:
Add a function to obtain the number of active series generated by a
check based on its configuration. The number is obtained from the
actual metrics emitted by each check type.


electron0zero (approved):
lgtm :+1:

PR #129, opened 2021-03-03
Fix: provide check type along with class info

mem:
In order to make it possible to associate accounting classes with check
types, extend the information returned by GetAccountingClassInfo to
include that detail.


electron0zero (approved):
lgtm :+1:

PR #133, opened 2021-03-18
Update blackbox_exporter to version 0.18.0

mem:
This brings in a new metric:

```
  # HELP probe_icmp_reply_hop_limit Replied packet hop limit (TTL for ipv4)
  # TYPE probe_icmp_reply_hop_limit gauge
  probe_icmp_reply_hop_limit 64
```


internal/scraper/testdata/ping.txt
```diff
@@ -1,14 +1,17 @@
 # HELP probe_dns_lookup_time_seconds Returns the time taken for probe dns lookup in seconds
 # TYPE probe_dns_lookup_time_seconds gauge
-probe_dns_lookup_time_seconds 1.4305e-05
+probe_dns_lookup_time_seconds 1.742e-06
```
samjewell, line 3: These changes are making this PR (and the git history) a bit more noisy. Do we need all of these changes, or could we omit them? It'd be great if we could reset these, and just keep the new additions on lines 12-14.

samjewell (approved):
Looks good to me, nice work 👏

internal/scraper/testdata/ping.txt
mem: I would rather not hand-edit the generated files.

samjewell: It's just a matter of selectively staging changes, no? `git add -p`? I think it's worth it, if you believe that code (and the git log) is read more times than it is written 🤔 .

pkg/accounting/data.go
```diff
@@ -8,8 +8,8 @@ var activeSeriesByCheckType = map[string]int{
 	"http_basic":     34,
 	"http_ssl":       122,
 	"http_ssl_basic": 38,
-	"ping":           81,
-	"ping_basic":     25,
+	"ping":           82,
+	"ping_basic":     26,
```
electron0zero, line 12: we need to update these numbers in plugin, https://github.com/grafana/synthetic-monitoring-app/blob/9113ffb972b6a22bf5218d2db5e18a7708a6f4e4/src/checkUsageCalc.ts#L16

electron0zero (approved):
need to update values in plugin, rest lgtm :+1:

PR #140, opened 2021-03-25
Fix: check that there are no duplicate label names

### LINKED ISSUE #139: Checks validation should not allow duplicate label names
Duplicate label names in a check will cause issues with Loki, we should validate them to make sure there are no duplicates.

mem:
Validate that user-defined labels are not using duplicated names, e.g.
label_a = 1 and label_a = 2.

Fixes #139


rdubrock (approved):

electron0zero (approved):
lgtm :+1:

PR #143, opened 2021-04-02
Feature: validate HTTP headers

mem:
Validate the headers specified in the HTTP settings.


rdubrock (approved):
🦕

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -23,6 +23,8 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+
+	"golang.org/x/net/http/httpguts"
```
electron0zero, line 27: looking at package docs and noticed.

> This package is shared by the standard library (which vendors it) and x/net/http2. It comes with no API stability promise. 

let's surface the fact that any update in httpguts can be a breaking change.

pkg/pb/synthetic_monitoring/checks_extra_test.go
```diff
@@ -728,6 +728,55 @@ func TestValidateLabel(t *testing.T) {
...
+			},
+			expectError: false,
+		},
+		"empty": {
+			input: HttpSettings{
+				Headers: []string{""},
+			},
+			expectError: true,
+		},
+		"no colon": {
```
electron0zero, line 758: let's add a tests for:
- only colon with space around it (`" : "`)
- unicode (emojis or non english text) in header values (`header: यूनिकोड टेक्स्ट`)

> यूनिकोड टेक्स्ट says `Unicode text` in Hindi language

electron0zero (approved):

pkg/pb/synthetic_monitoring/checks_extra.go
mem: We have quite a number of packages that fit that fit that description. The whole space under `golang.org/x` for example, as well as BBE.

PR #144, opened 2021-04-02
Add compression option

mem:


internal/scraper/scraper.go
```diff
@@ -824,26 +828,50 @@ func httpSettingsToBBEModule(ctx context.Context, logger zerolog.Logger, setting
...
+	for _, str := range settings.FailIfBodyNotMatchesRegexp {
+		re, err := bbeconfig.NewRegexp(str)
+		if err != nil {
+			return m, err
+		}
 
-	m.HTTP.FailIfBodyNotMatchesRegexp = make([]string, len(settings.FailIfBodyNotMatchesRegexp))
-	copy(m.HTTP.FailIfBodyNotMatchesRegexp, settings.FailIfBodyNotMatchesRegexp)
+		m.HTTP.FailIfBodyNotMatchesRegexp = append(m.HTTP.FailIfBodyNotMatchesRegexp, re)
+	}
```
mem, line 849: this is here because the new BBE version switched the type of `FailIfBodyNotMatchesRegexp`.

pkg/pb/synthetic_monitoring/checks_extra_test.go
```diff
@@ -801,6 +804,119 @@ func TestHttpSettingsValidate(t *testing.T) {
...
+
+	testcases := map[string]struct {
+		unserialized testStruct
+		serialized   []byte
+	}{
+		"none": {
+			unserialized: testStruct{
+				Compression: CompressionAlgorithm_none,
+			},
+			serialized: []byte(`{}`),
```
rdubrock, line 820: Just to be sure, this should evaluate to no decrompression if there is no `compression` key _or_ `compression: none` _or_ `compression: ""`, right?

rdubrock (approved):

pkg/pb/synthetic_monitoring/checks_extra_test.go
mem: @rdubrock that's right.

PR #147, opened 2021-04-21
Add MTR traceroute

rdubrock:
This PR adds a `traceroute` prober to the agent. It is using the https://github.com/tonobo/mtr under the hood (a go implementation of [mtr](https://en.wikipedia.org/wiki/MTR_(software))). MTR sends a burst of pings for each TTL, which allows us to produce more useful metrics. 

Things to look for:
- Traceroute is the first non-BBE prober type in the agent, so many of the types had to be wrapped in order to use the shared utlities in `scraper.go`.
- `traceroute.go` has the actual traceroute scraping implementation
-  Both timeout and frequency values are larger for traceroute checks than the other check types, since traceroutes take longer and generate a lot more requests (I took a guess at the numbers)

Things left to do:

- [ ] Performance/load testing

- [ ] Submit a PR to mtr in order to bind requests to a random integer instead of the process id. Using the process ID can cause collisions when listening for packets since we are running repeatedly from the same process

- [x] Abstract away validation logic to reduce the number of `if` statements in `checks_extra.go`


go.mod
```diff
@@ -49,3 +51,5 @@ replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.0+inc
 // 78d2af792bab is the commit tagged as v12.0.0.
 
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+
+replace github.com/tonobo/mtr v0.1.1-0.20210410081413-c8f9fe0b48b9 => ../mtr
```
rdubrock, line 55: I'll need to point this back to the actual library. I had a local copy running to experiment with changing the mtr library

internal/scraper/traceroute.go
```diff
@@ -0,0 +1,72 @@
...
+var (
+	COUNT            = 5
+	TIMEOUT          = 800 * time.Millisecond
+	INTERVAL         = 100 * time.Millisecond
+	HOP_SLEEP        = time.Nanosecond
+	MAX_HOPS         = 64
+	MAX_UNKNOWN_HOPS = 30
+	RING_BUFFER_SIZE = 50
+	PTR_LOOKUP       = false
+	srcAddr          = ""
```
samjewell, line 22: uppercase??

internal/scraper/scraper.go
```diff
@@ -52,7 +69,7 @@ type Scraper struct {
 	logger        zerolog.Logger
 	check         sm.Check
 	probe         sm.Probe
-	bbeModule     *bbeconfig.Module
+	bbeModule     *ConfigModule
```
samjewell, line 72:
```suggestion
	module     *ConfigModule
```
? Maybe this is no longer a BBE Module

internal/scraper/scraper.go
```diff
@@ -314,12 +331,15 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, erro
...
 	}
 
+	s.logger.Info().Msg("Collecting data")
 	// set up logger to capture check logs
 	logs := bytes.Buffer{}
 	bl := kitlog.NewLogfmtLogger(&logs)
 	sl := kitlog.With(bl, "ts", kitlog.DefaultTimestampUTC, "target", target)
 
-	success, mfs, err := getProbeMetrics(ctx, prober, target, s.bbeModule, s.buildCheckInfoLabels(), s.summaries, s.histograms, sl, s.check.BasicMetricsOnly)
+	success, mfs, err := getProbeMetrics(ctx, prober, target, s.bbeModule, s.buildCheckInfoLabels(), s.summaries, s.histograms, sl, s.check.BasicMetricsOnly, s.probe.Name)
```
samjewell, line 340: Shall we look into this more, as one of the first things we do? Maybe we can reverse this change, if the probe-name is already available to the node-graph / datasource somewhere/somehow?

internal/scraper/scraper.go
```diff
@@ -314,12 +331,15 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, erro
...
+
+	s.logger.Info().Msg("Collected metrics")
```
samjewell, line 342: ✂️

pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -192,10 +192,18 @@ message Check {
...
+	DnsSettings  dns  						= 3 [(gogoproto.jsontag) = "dns,omitempty"];
+	TcpSettings  tcp  						= 4 [(gogoproto.jsontag) = "tcp,omitempty"];
+	TracerouteSettings traceroute = 5 [(gogoproto.jsontag) = "traceroute,omitempty"];
+}
+
+message TracerouteSettings {
+	int64 maxHops 	 		 = 1 [(gogoproto.jsontag) = "maxHops"]; 
+	int64 maxUnknownHops = 2 [(gogoproto.jsontag) = "maxUnknownHops"];
+	bool  ptrLookup 		 = 3 [(gogoproto.jsontag) = "ptrLookup"];
+	int64 hopTimeout 		 = 4 [(gogoproto.jsontag) = "hopTimeout"];
```
samjewell, line 206: Could we add some code comments inline, at the end of each line here? Like we do for `PingSettings` below 🙏

samjewell: These config options could be removed from this PR and go into a later PR potentially. In order to make this PR smaller, and easier to work on.

internal/scraper/testdata/ping.txt
```diff
@@ -1,14 +1,14 @@
...
-probe_duration_seconds 0.000129361
+probe_duration_seconds 0.000974644
 # HELP probe_icmp_duration_seconds Duration of icmp request by phase
 # TYPE probe_icmp_duration_seconds gauge
-probe_icmp_duration_seconds{phase="resolve"} 1.742e-06
-probe_icmp_duration_seconds{phase="rtt"} 6.0107e-05
-probe_icmp_duration_seconds{phase="setup"} 2.5008e-05
+probe_icmp_duration_seconds{phase="resolve"} 4.1071e-05
+probe_icmp_duration_seconds{phase="rtt"} 0.00031317
+probe_icmp_duration_seconds{phase="setup"} 0.000340032
```
samjewell, line 11: Argg - these diffs kinda irritate me! I guess I'm wondering if we could exclude all these diffs from this PR, to make it easier to review. I know that puts a greater burden on the coder constructing the commits/PR though, so I'm ok if you want to keep them.

pkg/accounting/data.go
```diff
@@ -2,16 +2,18 @@
...
+	"http_ssl":         122,
+	"http_ssl_basic":   38,
+	"ping":             82,
+	"ping_basic":       26,
+	"tcp":              37,
+	"tcp_basic":        23,
+	"tcp_ssl":          41,
+	"tcp_ssl_basic":    27,
+	"traceroute":       20,
+	"traceroute_basic": 20,
```
samjewell, line 18: Does this make sense? If these are the same, it feels to me like we don't actually have a "basic" version of this check... 🤔

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -147,14 +150,22 @@ func (c *Check) Validate() error {
 	}
 
 	// frequency must be in [1, 120] seconds
-	if c.Frequency < 1*1000 || c.Frequency > 120*1000 {
+	if c.Type() != CheckTypeTraceroute && (c.Frequency < 1*1000 || c.Frequency > 120*1000) {
```
samjewell, line 153: So does Traceroute not have a check frequency setting at all? How often will the traceroute check be run?

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -147,14 +150,22 @@ func (c *Check) Validate() error {
...
+		return ErrInvalidCheckTimeout
+	}
+
+	if c.Type() != CheckTypeTraceroute && c.Timeout > 10*1000 {
+		return ErrInvalidCheckTimeout
+	}
+
+	if c.Timeout < 1*1000 || c.Timeout > c.Frequency {
 		return ErrInvalidCheckTimeout
 	}
```
samjewell, line 170: A couple of questions on this part:
- Could we add a few more comments to explain why you've chosen to set up the validations this way?
- I'm wondering whether we could extract the complexity out here somehow, and simplify (compose?) the code a bit more here? I'm thinking we could have a `validate_traceroute` function, separate from a `validate_bbe_check` function (or something similar), in order to have fewer places where we are doing `c.Type() == CheckTypeTraceroute` or `c.Type() =! CheckTypeTraceroute`

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -180,6 +191,10 @@ func (c *Check) Validate() error {
 		settingsCount++
 	}
 
+	if c.Settings.Traceroute != nil {
+		settingsCount++
+	}
```
samjewell, line 196: This is nice 👏 🎖️

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -220,6 +235,15 @@ func (c *Check) Validate() error {
...
 		}
+
+	case CheckTypeTraceroute:
+		if err := validateHost(c.Target); err != nil {
+			return ErrInvalidHostname
+		}
+
+		if err := c.Settings.Traceroute.Validate(); err != nil {
+			return err
+		}
```
samjewell, line 246: I'm wondering whether this code can move up a few lines in this file, so that it lives as close to `case CheckTypePing:` as possible. Ie. just above or below lines 203-210. That way it's easier to eyeball any differences between the two, and hopefully see that they are consistent. 🙏

pkg/pb/synthetic_monitoring/checks_extra_test.go
```diff
@@ -212,6 +212,19 @@ func TestCheckType(t *testing.T) {
...
+				Job:       "job",
+				Frequency: 60000,
+				Timeout:   30000,
+				Probes:    []int64{1},
+				Settings: CheckSettings{
+					Traceroute: &TracerouteSettings{},
+				},
+			},
+			expected: CheckTypeTraceroute,
+		},
```
samjewell, line 227: Again could we move this either directly above or below Ping? So we can eyeball more easily that they are the same, or similar?

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -147,14 +150,22 @@ func (c *Check) Validate() error {
...
 		return ErrInvalidCheckFrequency
 	}
 
 	// timeout must be in [1, 10] seconds, and it must be less than
 	// frequency (otherwise we can end up running overlapping
 	// checks)
-	if c.Timeout < 1*1000 || c.Timeout > 10*1000 || c.Timeout > c.Frequency {
+	if c.Type() == CheckTypeTraceroute && c.Timeout > 10*3000 {
+		return ErrInvalidCheckTimeout
+	}
```
samjewell, line 162: Is this validation tested at all? I'm wondering whether we could add a test for this.
I'm not too concerned though - somehow it seems we only test for `c.Timeout > c.Frequency` in the test suite so far, and have no tests for `> 10*1000` or `< 1*1000`, as far as I can tell.

internal/scraper/traceroute.go
```diff
@@ -0,0 +1,72 @@
...
+	totalPacketsSent := float64(0)
+	success := false
+	for _, hop := range m.Statistic {
+		totalPacketsLost += float64(hop.Lost)
+		totalPacketsSent += float64(hop.Sent)
+		avgElapsedTime := time.Duration(hop.Avg()) * time.Millisecond
+		if hop.Target == m.Address {
+			success = true
+		}
+		logger.Log("Level", "info", "Destination", m.Address, "Host", hop.Target, "TTL", hop.TTL, "ElapsedTime", avgElapsedTime, "LossPercent", hop.Loss(), "Sent", hop.Sent, "TraceID", traceID, "Probe", probeName)
```
samjewell, line 52: Here too, I think we concluded that we can remove `"Probe", probeName`

internal/scraper/scraper_test.go
```diff
@@ -169,18 +194,20 @@ func TestValidateMetrics(t *testing.T) {
...
+func verifyProberMetrics(t *testing.T, name string, prober ProbeFn, setup func(t *testing.T) (string, func()), config ConfigModule, basicMetricsOnly bool) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
 	summaries := make(map[uint64]prometheus.Summary)
 	histograms := make(map[uint64]prometheus.Histogram)
-	logger := &testLogger{w: ioutil.Discard}
+	// logger := &testLogger{w: ioutil.Discard}
+
+	logger := kitlog.NewJSONLogger(kitlog.NewSyncWriter(os.Stdout))
```
samjewell, line 205: Can you explain what's happening here 🙏

internal/scraper/scraper.go
```diff
@@ -314,12 +331,15 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, erro
 		target = addCacheBustParam(s.target, s.check.Settings.Http.CacheBustingQueryParamName, s.probe.Name)
 	}
 
+	s.logger.Info().Msg("Collecting data")
```
samjewell, line 334: ✂️

internal/scraper/config.go
```diff
@@ -0,0 +1,11 @@
+package scraper
+
+import (
+	"github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
+	bbeconfig "github.com/prometheus/blackbox_exporter/config"
+)
+
+type ConfigModule struct {
+	bbeconfig.Module
+	Traceroute synthetic_monitoring.TracerouteSettings `yaml:"traceroute,omitempty"`
```
samjewell, line 10: We discussed possibly changing this as follows, and making it anonymous, in order to promote the fields within `TracerouteSettings` up to `ConfigModule`.
```suggestion
      synthetic_monitoring.TracerouteSettings `yaml:"traceroute,omitempty"`
```
But we choose not to do this, in order to retain more similarity with how the checks are run within BlackBoxExporter.

samjewell (commented):
This is looking pretty nice - great work putting this all together, I'm very impressed 🎉 

Do you think you could add a PR description pls?
Also, how did the MTR library compare in the end, to the Traceroute library we were using before? How did it compare on the fronts we were considering:
- Concurrency
- Bug with any returning ping being mistaken for the ping that was sent
- was there another factor?

rdubrock:
> Does this make sense? If these are the same, it feels to me like we don't actually have a "basic" version of this check... 🤔  

I think the basic version is the one we have, probably. I'm thinking we just don't have a "basic" toggle/setting for v1, and as we come up with more use cases that can be supported by more metrics we can add it in

internal/scraper/testdata/ping.txt
rdubrock: Yeah it's a good point, there isn't really isn't any utility in updating the numbers, I think it's just the lines that we care about

pkg/accounting/data.go
rdubrock: I think the basic version is the one we have, probably. I'm thinking we just don't have a "basic" toggle/setting for v1, and as we come up with more use cases that can be supported by more metrics we can add it in

pkg/pb/synthetic_monitoring/checks_extra.go
rdubrock: It does, but I sort of grafted on into a different place. The current frequency/timeout settings are the same for all probe types because the frequency/timeout settings make sense for all of them (1-10 second timeout, 10 second to 2 minute frequency). Traceroute, however, usually takes between 5-30 seconds, so we need to have a different range for those settings

pkg/pb/synthetic_monitoring/checks_extra.go
rdubrock: 🤔 This is a good idea

internal/scraper/scraper_test.go
rdubrock: Yes! I was having trouble with the tests so I wanted to see the log output (the test logger is just discarding them, I think). I forgot to revert the change.

pkg/pb/synthetic_monitoring/checks.proto
rdubrock: I added the comments but left the config for now. Totally open to pulling them out, though

rdubrock:
> Also, how did the MTR library compare in the end, to the Traceroute library we were using before? How did it compare on the fronts we were considering:


> 
>     * Concurrency
> 
MTR is concurrent, which is good. It does a discovery pass to find the host addresses for all the TTL values, and then a second pass where is throws a burst of pings at each host to collect metrics. I don't think that overall it's much faster than the previous traceroute, but we're getting quite a bit more data out of it. 

>
>     * Bug with any returning ping being mistaken for the ping that was sent
> 

There is a bug (or more just a change that we need to make to accommodate our use case). The previous traceroute library naively listened to _all_ ICMP packets, which was a problem not only for traceroute executions colliding with each other, but also for colliding with other ping probes. The MTR library binds pings to the process ID, which makes sense for a CLI tool, but doesn't quite work for us because we run traceroutes many times from one process. This can be fixed with a one line change.

>
>     * was there another factor?
>

Only additional factor is that there is value add from being able to get a more robust set of stats around packet loss, latency, etc with the burst that MTR provides.

pkg/pb/synthetic_monitoring/checks_extra.go
rdubrock: I pulled out timeout and frequency validations into their own functions as a first step (also because this function was running into a "cyclomatic complexity" linting rule)

internal/scraper/traceroute.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		targets := strings.Join(hop.Targets, ",")
+		hosts += targets
+		err := level.Info(logger).Log("Level", "info", "Destination", m.Address, "Hosts", targets, "TTL", hop.TTL, "ElapsedTime", avgElapsedTime, "LossPercent", hop.Loss(), "Sent", hop.Sent, "TraceID", traceID)
+		if err != nil {
+			continue
+		}
+	}
+
+	traceHash := fnv.New32()
```
rdubrock, line 76: This hashing doesn't seem to actually work in it's current state. I think the string input isn't quite right.

Issue #117, opened 2021-02-04
User Agent String should indentify synthetic-monitoring-agent


electron0zero:
synthetic-monitoring-agent should identify itself with appropriate user agent string to make it easy for operators to identify where the traffic is coming from.

also see: https://github.com/prometheus/blackbox_exporter/pull/557

PR #150, opened 2021-04-26
Set user agent in outgoing HTTP requests

### LINKED ISSUE #142: User agent string should reflect the currently running version
The user agent is configured [here](https://github.com/grafana/synthetic-monitoring-agent/blob/69fe1d9fe8149dc9a25ef8d27211dceaea247d97/internal/pusher/pusher.go#L22)

It should include the current version, so we know what agents are currently running

mem:
The user agent identifies the synthetic monitoring agent version, OS,
architecture, buildstamp and commit hash, and it points to the synthetic
monitoring product page on grafana.com so that people can more easily
identify requests as coming from Synthetic Monitoring.

The more explicit user agent version is also passed to Prometheus so
that it's easier to identify issues.

Closes #142 #117

### DIFF
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -20,6 +20,7 @@ import (
 	"github.com/go-logfmt/logfmt"
 	"github.com/grafana/loki/pkg/logproto"
 	"github.com/grafana/synthetic-monitoring-agent/internal/pusher"
+	"github.com/grafana/synthetic-monitoring-agent/internal/version"
 	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
 	"github.com/mmcloughlin/geohash"
 	bbeconfig "github.com/prometheus/blackbox_exporter/config"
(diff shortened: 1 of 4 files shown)


internal/version/version.go
```diff
@@ -1,9 +1,19 @@
 package version
 
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"runtime/debug"
+)
+
+const smHomepage = "https://grafana.com/docs/grafana-cloud/synthetic-monitoring/"
```
electron0zero, line 10: nit: let's link to this repo because grafana.com paths can change without notice.

```suggestion
const smHomepage = "https://github.com/grafana/synthetic-monitoring-agent"
```

internal/scraper/scraper.go
```diff
@@ -804,16 +805,25 @@ func httpSettingsToBBEModule(ctx context.Context, logger zerolog.Logger, setting
...
 	for _, header := range settings.Headers {
 		parts := strings.SplitN(header, ":", 2)
+
 		var value string
 		if len(parts) == 2 {
 			value = strings.TrimLeft(parts[1], " ")
 		}
-		m.HTTP.Headers[parts[0]] = value
+
+		if strings.ToTitle(parts[0]) == userAgentHeader {
```
electron0zero, line 822: bit confused about what is going on here, can we add a comment about it?

internal/version/version.go
```diff
@@ -17,3 +27,17 @@ func Commit() string {
...
+}
+
+func buildUserAgentStr() string {
+	program := "unknown"
+
+	if info, ok := debug.ReadBuildInfo(); ok {
+		program = filepath.Base(info.Path)
+	}
+
+	return fmt.Sprintf("%s/%s (%s %s; %s; %s; +%s)", program, version, runtime.GOOS, runtime.GOARCH, commit, buildstamp, smHomepage)
```
electron0zero, line 42: 👏🏼 👏🏼

internal/scraper/scraper.go
mem: Headers are case insensitive (https://tools.ietf.org/html/rfc7230#section-3.2) so we need to fold them. `strings.EqualFold` folds both strings all the time, so it's a little more expensive, so it's better to have one string in the correct case and just fold the other one. I'll change this to use lowercase and maybe that's less obscure.

electron0zero (approved):
lgtm :+1:

PR #171, opened 2021-05-20
Feature: add a features flag on the command line

mem:
By passing `--features foo,bar,baz` it's possible to indicate that a
particular agent is willing to accept running new features while they
are being developed.

No features have been defined yet.


electron0zero (approved):
lgtm :+1:

PR #184, opened 2021-06-21
Feature: add release script

mem:
Add a script capturing the steps necessary to create a release.


samjewell (commented):
Thanks for tagging me, this is very neat. Looks like it depends on us using `Fix:` and `Feature:` etc. as prefixes in our PR description titles, so I'll try to do that going forwards.

Issue #196, opened 2021-08-03
Histogram metrics are missing +Inf bucket


mem:
Sometime ago we changed the way histograms are created, and it seems we dropped the +Inf bucket and we didn't notice.

The +Inf bucket is present in the testdata used to control the published metrics do not change, but the conversion from Go's client metrics to the protobuf version that is pushed to Prometheus comes at a later step.

PR #197, opened 2021-08-04
Fix: add +Inf bucket to histograms

mem:
At some point the implementation changed and dropped the +Inf bucket and
nobody noticed (expect me, because I clearly remember fixing this a
while ago).

Fixes: #196

### DIFF
--- a/internal/scraper/scraper_test.go
+++ b/internal/scraper/scraper_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/prometheus/prompb"
 	"github.com/rs/zerolog"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


internal/scraper/scraper_test.go
```diff
@@ -492,6 +493,284 @@ func setupTCPServerWithSSL(t *testing.T) (string, func()) {
...
+			require.Equal(t, len(tc.labels), len(actual.Labels))
+			require.Equal(t, tc.labels, actual.Labels)
+			require.Equal(t, 1, len(actual.Samples))
+			require.Equal(t, tc.t.UnixNano()/1e6, actual.Samples[0].Timestamp)
+			require.Equal(t, tc.value, actual.Samples[0].Value)
+		})
+	}
+}
+
+func TestAppendDtoToTimeseries(t *testing.T) {
```
samjewell, line 541: @mem I think there's a typo here. Feels like it should read `TestAppendToTimeseries`

samjewell: oh ok, my bad - thanks for your reply. So can I ask - what is "Dto"? What does it mean? Thanks!

mem: Data Transfer Object. In this case it's the Prometheus representation of the metric that is transmitted between processes.

PR #198, opened 2021-08-05
Only build/publish when tagging a release

samjewell:
and not when merging to master. As far as I'm aware this has two
benefits:
1. We only publish packages/docker images for other people to use when
   we are also using them, so we won't be the last to know when we break
   something.
2. We avoid a race condition, where two build/publish processes are
   happening concurrently when we tag a new release.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -49,7 +49,6 @@ steps:
       from_secret: docker_username
   when:
     ref:
-    - refs/heads/main
     - refs/tags/v*.*.*
 
 - name: package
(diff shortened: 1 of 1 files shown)


electron0zero (approved):
lgtm :+1:

.drone/drone.yml
```diff
@@ -49,7 +49,6 @@ steps:
       from_secret: docker_username
   when:
     ref:
-    - refs/heads/main
```
mem, line 52: this is fine... now we need a way to push builds from main to docker so that we can use them in staging...

use a different image name?

mem (approved):

.drone/drone.yml
samjewell: Good suggestion - thanks for the review, and for creating that github issue 👍

PR #203, opened 2021-08-17
Update Drone signature and add README

julienduchesne:
Builds are currently not running because the signature is out of date


.drone/drone.jsonnet
```diff
@@ -21,7 +21,6 @@ local pipeline(name, steps=[]) = {
 local releaseOnly = {
   when: {
     ref+: [
-      'refs/heads/main',
```
julienduchesne, line 24: This was modified in the `.drone.yml` file directly

mem (approved):

PR #212, opened 2021-08-30
Remove loki dependency

mem:


internal/pkg/logproto/README
```diff
@@ -0,0 +1,6 @@
+In order to avoid pulling in a lot of problematic dependencies, the
+contents of this directory have been copied verbatim from
```
electron0zero, line 2: we might want to add a script to update this dir to keep it in sync with upstream.

updates will become a routine operation to keep up with loki.

electron0zero (approved):
one nit comment, rest lgtm 👍🏼

rdubrock (approved):
Is there a background/an issue on why we need to remove loki as a dependency?

mem:
> Is there a background/an issue on why we need to remove loki as a dependency?

What we need from Loki is the protocol buffer serialization / deserialization code.

We are pulling everything with all the dependencies, and there are a couple down the chain that require ugly replace hacks in go.mod, and we end up propagating that problem down to whatever imports synthetic-monitoring-agent.

internal/pkg/logproto/README
mem: Thinking about this ... not sure.

There are some small changes in Loki version 2, but they are related to the server side of things, not really the client.

In principle I would prefer to NOT provide a path for upgrading to a newer version without having to carefully consider what that change entails.

PR #213, opened 2021-08-31
Fix: update fpm to 1.13.1

mem:
They fixed the bug that caused us to pin to version 1.11.0 (empty if
branch causing a shell syntax error). Pin it to a newer version that
works with whatever changed in ffi.


samjewell (commented):
Looks good to me- thanks for the fix 😎🤩

PR #219, opened 2021-09-13
add features to startup command

rdubrock:
We should have the ability to pass feature flags into agents running in VMs. I _think_ this is all that requires.


electron0zero (approved):
lgtm

samjewell (approved):
Looks fine 👍

PR #221, opened 2021-09-16
Feature: disconnect agent from API on signal

mem:
If the agent receives a SIGUSR1 signal, it will disconnect from the API,
but it will continue running checks. This makes it possible for a new
version of the agent to connect to the API. When that happens, it's
possible to kill the running agent (SIGTERM or otherwise).

This should provide for a smoother upgrade process.

### DIFF
--- a/internal/checks/checks.go
+++ b/internal/checks/checks.go
@@ -5,8 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
+	"os/signal"
 	"strconv"
 	"sync"
+	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/grafana/synthetic-monitoring-agent/internal/feature"
@@ -24,8 +28,9 @@ import (
 )
 
 var (
-	errNotAuthorized    = errors.New("probe not authorized")
-	errTransportClosing = errors.New("transport closing")
+	errNotAuthorized     = errors.New("probe not authorized")
+	errTransportClosing  = errors.New("transport closing")
+	errProbeUnregistered = errors.New("probe no longer registered")
 )
 
 // Updater represents a probe along with the collection of scrapers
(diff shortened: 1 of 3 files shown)


internal/checks/checks.go
```diff
@@ -279,18 +297,67 @@ func (c *Updater) loop(ctx context.Context) error {
...
+			return errProbeUnregistered
+		}
+
+		return grpcErrorHandler(action, err)
 	}
 
 	return nil
 }
 
+func installSignalHandler(ctx context.Context) (context.Context, *int32) {
```
electron0zero, line 340: nit: let's add docs on this func

internal/checks/checks.go
```diff
@@ -204,16 +209,29 @@ func (c *Updater) Run(ctx context.Context) error {
 				Msg("context cancelled, closing updater")
 			return nil
 
+		case errors.Is(err, errProbeUnregistered):
+			// Probe unregistered itself from the API, wait
+			// until attempting to reconnect again.
+			c.logger.Warn().
+				Str("connection_state", c.api.conn.GetState().String()).
+				Msg("unregistered probe in API, sleeping for 1 minute...")
```
electron0zero, line 217: thinking about reconnect here.

`errProbeUnregistered` means that we disconnected when `SIGUSR1` fired. i.e. we told probe to disconnect but keep running checks.

but if we reconnect back in a minute, `SIGUSR1` will `disconnect probe for a minute` and not `disconnect` forever.

I think one minute is reasonable time to reconnect back but maybe we should make it configurable.

internal/checks/checks_test.go
```diff
@@ -0,0 +1,95 @@
...
+			// verify that the signal context is done after
+			// receiving the signal, and that the signal is
+			// correctly reported as having fired.
+
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+			defer cancel()
+			sigCtx, signalFired := installSignalHandler(ctx)
+			require.NotNil(t, sigCtx)
+			require.NotNil(t, signalFired)
+			require.NoError(t, syscall.Kill(syscall.Getpid(), syscall.SIGUSR1))
```
electron0zero, line 25: let's document this signal in README

internal/checks/checks.go
mem: If we find that the timeout needs to be configured we can make it configurable once we get there.

1 minute should be enough for the other agent to connect and start running, but I know there have been instances where new agents take longer to do that. We need to experiment and figure out what's the appropriate strategy.

electron0zero (approved):
lgtm :+1:

PR #230, opened 2021-09-28
[cr] test coverage for NewProber for each probe type

jewbetcha:
Gets coverage for each prober to >90%

### DIFF
--- a/internal/prober/dns/dns_test.go
+++ b/internal/prober/dns/dns_test.go
@@ -15,8 +15,9 @@ func TestName(t *testing.T) {
 
 func TestNewProber(t *testing.T) {
 	testcases := map[string]struct {
-		input    sm.Check
-		expected Prober
+		input       sm.Check
+		expected    Prober
+		ExpectError bool
 	}{
 		"default": {
 			input: sm.Check{
@@ -38,6 +39,7 @@ func TestNewProber(t *testing.T) {
 					},
 				},
 			},
+			ExpectError: false,
 		},
 		"no-settings": {
 			input: sm.Check{
@@ -46,7 +48,8 @@ func TestNewProber(t *testing.T) {
 					Dns: nil,
 				},
 			},
-			expected: Prober{},
+			expected:    Prober{},
+			ExpectError: true,
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


internal/prober/http/http_test.go
```diff
@@ -13,6 +13,73 @@ import (
...
+				Settings: sm.CheckSettings{
+					Http: nil,
+				},
+			},
+			expected: Prober{},
+		},
+	}
+
+	for name, testcase := range testcases {
+		ctx := context.TODO()
```
mem, line 69: Conventionally `context.TODO()` means "I need a context here, I don't know what to do use right now, I will add one later". It's a sort of TODO note that is part of the code.

Usually you would use either contxt.Background() or context.WithTimeout, but for this particular test a timeout doesn't make sense (because the prober doesn't run).

internal/prober/traceroute/traceroute_test.go
```diff
@@ -1,13 +1,74 @@
...
+	for name, testcase := range testcases {
+		logger := zerolog.New(io.Discard)
+		t.Run(name, func(t *testing.T) {
+			actual, err := NewProber(testcase.input, logger)
+			require.Equal(t, &testcase.expected, &actual)
+			if name == "no-settings" {
+				require.NotNil(t, err)
+			} else {
+				require.Nil(t, err)
+			}
```
mem, line 67: and here, too

mem (changes requested):
Thanks, looks good!

Just a couple of small idiomatic fixes.

internal/prober/dns/dns_test.go
```diff
@@ -8,6 +8,61 @@ import (
...
+
+	for name, testcase := range testcases {
+		t.Run(name, func(t *testing.T) {
+			actual, err := NewProber(testcase.input)
+			require.Equal(t, &testcase.expected, &actual)
+			if name == "no-settings" {
+				require.NotNil(t, err)
+			} else {
+				require.Nil(t, err)
+			}
```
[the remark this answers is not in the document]
jewbetcha: Sounds good, thanks for explaining 👍

mem (approved):

PR #239, opened 2021-11-03
use ptr lookup in traceroute

rdubrock:
There were a few traceroute settings that were stubbed out but not supported yet. This updates the agent to use the user configured settings instead of hardcoded values. It also stubs out a `FirstHop` setting that isn't supported yet. 

Depends on https://github.com/grafana/mtr/pull/3/commits/89b054ac21be3659e7f9484ad83792db50866f63


pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -369,4 +369,5 @@ message TracerouteSettings {
 	int64 maxUnknownHops = 2 [(gogoproto.jsontag) = "maxUnknownHops"]; // Maximum hops probe that give no response before giving up
 	bool  ptrLookup 		 = 3 [(gogoproto.jsontag) = "ptrLookup"];      // Include reverse DNS lookup
 	int64 hopTimeout 		 = 4 [(gogoproto.jsontag) = "hopTimeout"];		 // Timeout for individual hop pings
+	int64 firstHop 		 	 = 5 [(gogoproto.jsontag) = "firstHop"];		   // TTL to start at
```
mem, line 372: this doesn't seem to be used?

mem (approved):

pkg/pb/synthetic_monitoring/checks.proto
rdubrock: you're right, I stubbed it out for now because it's something I'd like to add in the future, but maybe I'll just remove it until it's actually time to add it to prevent confusion down the line

PR #242, opened 2021-11-24
Feature: add /ready endpoint for readiness probe

electron0zero:
When deplying probes in kubernetes like envirements, we need a way to know if agent process is ready.

probe is considered ready when it is to able connect to SM API. 

You can use `/ready` endpoint to check if probe is ready (was able to connect to SM API).  `/ready` endpoint will return HTTP OK (200) with `ready`  when probe is ready.

**demo:**

https://user-images.githubusercontent.com/9503187/143260060-f49a1def-2884-41da-949f-f1eea1780c11.mp4

### DIFF
--- a/cmd/synthetic-monitoring-agent/http.go
+++ b/cmd/synthetic-monitoring-agent/http.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/pprof"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -23,6 +24,7 @@ type MuxOpts struct {
 		prometheus.Registerer
 		prometheus.Gatherer
 	}
+	isReady *readynessHandler
 }
 
 func NewMux(opts MuxOpts) *Mux {
@@ -39,6 +41,9 @@ func NewMux(opts MuxOpts) *Mux {
 	)
 
 	router.Handle("/metrics", promHandler)
+	isReady := &atomic.Value{}
+	isReady.Store(false)
+	router.Handle("/ready", opts.isReady)
 
 	// Register pprof handlers
 	router.HandleFunc("/debug/pprof/", pprof.Index)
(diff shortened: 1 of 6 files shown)


internal/checks/checks.go
```diff
@@ -307,6 +310,8 @@ func (c *Updater) loop(ctx context.Context) error {
 	c.logger.Info().Int64("probe id", c.probe.Id).Str("probe name", c.probe.Name).Msg("registered probe with synthetic-monitoring-api")
 
 	c.metrics.connectionStatus.Set(1)
+	// true indicates that probe was able to connect to API
+	c.IsReady.Store(true)
```
mem, line 314: while I agree that this does what we need, I would prefer to see a function here:

```
c.IsConnected(true)
defer c.IsConnected(false)
```

so what needs to be stored in the Updater struct is:

```
    IsConnected func(bool)
```

same for the options.

At the location where the options are passed, you can have:

```
IsConnected: func(v bool) { isReady.Set(v) }
```

this allows you to propagate the `connected` state without embedding too much information as to why you need to know that state elsewhere. For example, elsewhere you could decide that "is ready" means "has been connected for 1 second" or "it connected to the API once" (or whatever other criteria is appropriate).

electron0zero: updated

cmd/synthetic-monitoring-agent/http.go
```diff
@@ -112,3 +117,21 @@ func defaultHandler() http.Handler {
 		fmt.Fprintln(w, "hello, world!")
 	})
 }
+
+// readyHandler is a readiness probe.
+func readyHandler(isConnected, isReady *atomic.Value) http.HandlerFunc {
```
mem, line 122: I don't follow why you need two variables now.

If the intention is to set isReady to true and never back to false, it's enough to do something like `func (v bool) { if v { isReady.Store(v) } }` in the isConnected function.

mem (approved):

PR #248, opened 2021-12-17
Fix: Retry with backoff to get a new client

electron0zero:
This PR adds backoff in retries to ensure that we are not hitting the API continuously.

### DIFF
--- a/internal/pusher/pusher.go
+++ b/internal/pusher/pusher.go
@@ -21,6 +21,12 @@ import (
 
 const defaultBufferCapacity = 10 * 1024
 
+var (
+	maxRetries = 3
+	minBackoff = 30 * time.Millisecond
+	maxBackoff = 250 * time.Millisecond
+)
+
 var bufPool = sync.Pool{
 	New: func() interface{} {
 		buf := make([]byte, 0, defaultBufferCapacity)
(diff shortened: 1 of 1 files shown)


mem (changes requested):
Please look at internal/checks/checks.go. Having multiple slightly different backoff implementations in the code makes it harder to read.

internal/pusher/pusher.go
```diff
@@ -128,8 +142,14 @@ func (p *Publisher) publish(ctx context.Context, payload Payload) {
 				logger.Error().Err(err).Msg("publish events")
 				p.errorCounter.WithLabelValues("logs", tenantStr, strconv.Itoa(httpStatusCode)).Inc()
 				if hasStatusCode && httpStatusCode == http.StatusUnauthorized {
-					// Retry to get a new client, credentials might be stale.
+					// Retry with backoff to get a new client, credentials might be stale.
 					newClient = true
+
+					retries--
+					if retries > 0 {
+						time.Sleep(backoff)
```
electron0zero, line 150: after looking more into the code, maybe we need to think about this, and instead of putting in a retry with backoff, we can do a single retry and give up. 

we are already giving up when we get an error while calling `getClient`, so if we get a invalid client with `newClient: true`, we will probably not get a valid again on second try. 

`publish` method is called in a new goroutine, and caller spins up a new goroutines to publish data from each check run. So If we sleep for a lot of time, a bad check can lead to goroutines piling up.

PR #279, opened 2022-04-20
Fix: fix http status code parsing for publish

electron0zero:
We had a temporary fix where we were looking at error string to check if we were getting 401 from server. 

`sm_agent_publisher_push_errors_total` metric is still broken after the temporary fix, and it was reporting `0` as status for all errors.

In this change, I am fixing the status code parsing code, and getting rid of temporary fix.


thepalbi (approved):
LGTM

PR #284, opened 2022-04-25
Add tooling to configure agent in local k3d cluster

thepalbi:
This PR adds some tooling for deploying a newly built image of Synthetic Monitoring agent in a kubernetes cluster.
The configuration needed is minimal:
```
API_TOKEN=<probe authentication code>
SM_API_URL=<synthetic monitoring api url> 
```

Once deployed, the script will create 3 resources:
- Pod with grafana agent running
- Service, to encapsulate pods under a networking abstraction
- Ingress, to make the agent http server accessible from `http://sm-agent.k3d.localhost:9999`

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,14 +1,15 @@
 ### Local environment tools
-Follow the steps below to work with the local Synthetic Monitoring cluster, and add a development version of the SM agent in there.
+Follow the steps below to work with a local Synthetic Monitoring kubernetes cluster, and add a development version of the SM agent in there.
 
-First, setup a local cluster with [sm-dev](https://github.com/grafana/synthetic-monitoring-api/tree/main/scripts#sm-dev).
-Once configured, go into the local Grafana instance and configure the SM Plugin. Then:
+First, install the `Synthetic Monitoring plugin` in a `Grafana` instance, and follow these instructions on how to add a `Probe`.
 1. Go into the SM Plugin.
-2. Click `Probes`. Then `New`.
-3. Configure a `Probe Name`, and a `Region` name (`amer` for example). Click `Save`.
-4. Copy the `Probe Authentication Token` in a config file with the following format:
+2. First go to `Config`. You will notice under `Backend Address` a URL. This is the Synthetic Monitoring API URL. Save that for later, we will call it `<sm api url>`.
+4. After that, click`Probes`. Then `New`.
+5. Configure a `Probe Name`, and a `Region` name (`amer` for example). Click `Save`.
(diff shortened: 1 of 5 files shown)


scripts/README.md
```diff
@@ -0,0 +1,26 @@
+### Local environment tools
+Follow the steps below to work with the local Synthetic Monitoring cluster, and add a development version of the SM agent in there.
+
+First, setup a local cluster with [sm-dev](https://github.com/grafana/synthetic-monitoring-api/tree/main/scripts#sm-dev).
```
mem, line 4: This is fine, but since this repository is public, it probably makes sense to not assume that people have access to the sm-api repo.

Just leave it at something along the lines of "install the Synthetic Monitoring plugin in a Grafana instance and follow these instructions to add a new probe".

Note that the plugin reports the URL of the Synthetic Monitoring API that it's configured to use. That should be noted and used in the YAML below.

scripts/agent.yaml.tmpl
```diff
@@ -0,0 +1,54 @@
...
+    app: sm-agent
+    env: development
+spec:
+  containers:
+    - image: grafana/synthetic-monitoring-agent:latest
+      name: agent
+      imagePullPolicy: Never
+      args:
+        - -api-server-address
+        - sm-api:4031
```
mem, line 15: This bit here is correct if you have access to a local SM API server, but most users won't.

So it makes sense to make that a parameter, like the API_TOKEN, and maybe default to `sm-api:4031` if it's not provided.

scripts/sm-dev
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
+# Inspired by https://github.com/grafana/synthetic-monitoring-api/tree/main/scripts#sm-dev
```
mem, line 2: while correct, most people won't have access to this.

mem (changes requested):
Thanks, looks good!

Just don't make it depend on having access to a local sm-api installation, as most people won't have that.

scripts/README.md
thepalbi: Ohh you are right, didn't noticed that. Will do!

scripts/sm-dev
thepalbi: Erased!

scripts/agent.yaml.tmpl
thepalbi: Done. Added as a defaulted config.

scripts/README.md
thepalbi: Also inlined the `*.localhost` troubleshooting section.

thepalbi:
Adapted PR description as well

mem (approved):
Nice! Thanks!

PR #286, opened 2022-05-03
Revert Inadvertent Commit

captncraig:
I'm sorry, I was experimenting locally, and somehow pushed from the wrong window. This is to back out that change. Should not have had serious impact on anything, as it is mostly unreachable code.


thepalbi (approved):

thepalbi:
@grafana/synthetic-monitoring-dev should we add branch protection to main?

PR #287, opened 2022-05-05
Enable blocking IPs (via bbe fork)

captncraig:
This implements blocked IP addresses inside the agent code itself. It does this by injecting a custom dialer into blackbox_exporter that checks the forbidden cidrs before dialing. This is not merged into bbe yet, but hopefully they will take it. 

https://github.com/prometheus/blackbox_exporter/compare/master...captncraig:cmp_dialer?expand=1


electron0zero:
going to close this PR, We decided to use infrastrucure level tooling to restrict access of agent.

PR #318, opened 2022-06-24
Small POC for testing the HTTP prober

mem:
Ideally we would have a set of tests for the full spectrum of HTTP
options. With the testserver package it's possible to control the
responses from the test itself.


electron0zero (approved):
lgtm, love this approach.

PR #324, opened 2022-07-28
Fix: use buildkit when building docker image

mem:
9315139aa048099a89b0f54a53c1187e6719f5e0 added the ability to build
images for other architectures, and as part of that the Dockerfile was
changed to use BUILDPLATFORM and TARGETPLATFORM. These values are
provided by buildkit, not by regular docker builds. Set the
DOCKER_BUILDKIT environment variable to 1 so that the `docker build`
command uses buildkit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,7 +216,8 @@ help: ## Display this help.
 
 .PHONY: docker
 docker: build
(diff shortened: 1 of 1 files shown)


mem:
This has been dealt with in other ways.

PR #341, opened 2022-09-20
Make agent more resilient to network failures

adriansr:
It's been observed that the agent can hang if the API server disappears from the network after the initial connect without closing the grpc connection. This is due to all code paths waiting indefinitely for a reply to a network message from the API server.

This PR adds a timeout to the ping sent within the checks updater, and extends the error handling so that the agent can terminate if no response is received after 30s.

Added a test to reproduce the problem.


adriansr:
Superseded by #342

PR #351, opened 2022-10-18
grpc: Send known checks to API on connect

adriansr:
Updates the changes logic to send the existing checks to the API server so that a reduced set of probes can be returned.

This results in a reduced memory and network footprint on the API side when the connection is lost or the API restarted and all probes connect at the same time.


mem (approved):
This makes sense. I think we can marge this and it should work with the existing API.

PR #375, opened 2022-11-24
Fix: default to listening on localhost, not all interfaces

mem:


adriansr (approved):

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -39,7 +39,7 @@ func run(args []string, stdout io.Writer) error {
 		reportVersion     = flags.Bool("version", false, "report version and exit")
 		grpcApiServerAddr = flags.String("api-server-address", "localhost:4031", "GRPC API server address")
 		grpcInsecure      = flags.Bool("api-insecure", false, "Don't use TLS with connections to GRPC API")
-		httpListenAddr    = flags.String("listen-address", ":4050", "listen address")
+		httpListenAddr    = flags.String("listen-address", "127.0.0.1:4050", "listen address")
```
DanCech, line 42: suggest localhost here to be consistent with line 40, unless there's a reason to prefer 127.0.0.1

mem: ah, I didn't notice that. Thanks!

PR #425, opened 2023-04-01
Add multihttp check type

mem:
This supports the multihttp check type.

This shoud be used in tandem with sm-k6-runner (the address should be specified as a command line option).

### DIFF
--- a/internal/prober/multihttp/script.go
+++ b/internal/prober/multihttp/script.go
@@ -1,0 +1,266 @@
+package multihttp
+
+import (
+	"bytes"
+	"embed"
+	"fmt"
+	"net/url"
+	"strings"
+	"text/template"
+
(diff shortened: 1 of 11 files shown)


internal/prober/multihttp/script.go
```diff
@@ -0,0 +1,264 @@
...
+
+	switch assertion.Type {
+	case sm.MultiHttpEntryAssertionType_TEXT:
+		cond := assertionCondition(assertion.Condition)
+
+		switch assertion.Subject {
+		case sm.MultiHttpEntryAssertionSubjectVariant_RESPONSE_BODY:
+			cond.Name(&b, "body", assertion.Value)
+			b.WriteString(`': response => `)
+			cond.Render(&b, "response.body", assertion.Value)
```
mem, line 135: this needs to perform variable replacements.

if the input looks like `${foo}` it needs to be replaced by `vars['foo']` so that variables are used...

PR #442, opened 2023-05-10
Add MultiHttp to SM proto

mem:
This is just the proto and some validations, nothing else.


pkg/pb/synthetic_monitoring/checks.proto
mem, line 1: there's a lot of noise in this file because I had to reformat it to clean it up.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -553,6 +610,253 @@ func (s *K6Settings) Validate() error {
...
+	return nil
+}
+
+// Validate verifies that the MultiHttpEntryAssertion is valid.
+//
+// Because of the structure represents multiple orthogonal variants, this
+// function has to branch based on the type.
+//
+//nolint:gocyclo
+func (a *MultiHttpEntryAssertion) Validate() error {
```
rdubrock, line 701: Nice! I was thinking that this part was going to be painful in go, interesting to see how that works

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -553,6 +610,253 @@ func (s *K6Settings) Validate() error {
 	return nil
 }
 
+func (s *MultiHttpSettings) Validate() error {
+	if len(s.Entries) == 0 {
+		return ErrInvalidMultiHttpTargets
+	}
+
+	if err := validateCollection(s.Entries); err != nil {
```
rdubrock, line 618: Do we want a max quantity of entries? There is a max of 10 on the frontend I believe, but we can pick anything that makes sense.

rdubrock (approved):
looks good! One thing that would be cool at some point would be to generate typescript defs from the proto file, because we've duplicated all this stuff on the frontend

pkg/pb/synthetic_monitoring/checks_extra.go
mem: That's a good point. I'll add that. We will need to work out how to implement per-tenant limits.

pkg/pb/synthetic_monitoring/checks_extra.go
mem: I had to try a couple of times :-P

I'm not 100% happy with _not_ validating the expressions (e.g. json path or CSS selector) but I went and looked and those are implemented in javascript, not Go, so I don't want to deal with the headache of having to different implementations with different bugs.

pkg/pb/synthetic_monitoring/checks_extra.go
mem: Done

PR #454, opened 2023-05-31
Fix: parametrize the k6 runner

mem:
The k6 runner was left hard-coded because I forgot about it.

This is a really noisy change that is adding a parameter to specifying the runner that should be used. The goal is to be able to write tests that don't actually run k6 scripts, but instead use pre-defined responses to make it predictable. This also helps with the accounting package that needs to actually run the prober to figure out the number of metrics it produces.


go.sum
mem, line 1: I have no idea why this file is so noisy.

PR #487, opened 2023-08-02
feat: Experimental per-tenant batched publisher

adriansr:
This adds a new publisher (enable with `--publisher v2` flag):

- Uses a single pusher loop per-tenant and is able to submit the results of multiple check runs on a single request.
- Keeps the check results compressed in memory to reduce the memory impact, specially during network outages.
- Can react to various error conditions like rate-limits or out-of-quota.
- Is configurable in terms of memory use (max queued memory per tenant).

### DIFF
--- a/internal/pkg/prom/prom.go
+++ b/internal/pkg/prom/prom.go
@@ -18,7 +18,7 @@ import (
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
-	conntrack "github.com/mwitkow/go-conntrack"
+	"github.com/mwitkow/go-conntrack"
 	"github.com/pkg/errors"
 	"github.com/prometheus/prometheus/prompb"
 )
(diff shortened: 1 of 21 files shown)


internal/pkg/prom/prom.go
```diff
@@ -373,7 +394,6 @@ func (c *Client) Store(ctx context.Context, req []byte) error {
 	for k, v := range c.headers {
 		httpReq.Header.Set(k, v)
 	}
-	httpReq = httpReq.WithContext(ctx)
```
adriansr, line 376: This was a no-op, as the httpReq's context is reset a few lines below.

I think it's correct to ignore the input ctx, as we're using small timeouts anyway, and we don't want to cancel an in-flight push, but highlighting this just in case it was a mistake.

internal/pusher/v2/condition.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"sync"
+)
+
+// condition is a simple, channel-based condition variable.
+type condition struct {
+	mu sync.Mutex
+	ch chan struct{}
+}
```
mem, line 11: If I'm understanding this correctly, the mutex exists so that the channel can be initialized.

While I understand the appeal of this:

```
    var ready condition
```

which is "ready to use", it does pay a price, which is that every single access has to acquire the mutex.

Instead, you could do this:

```
type condition chan struct{}

func newCondition() condition {
	return make(condition, 1)
}

func (c condition) Signal() {
	select {
	case c <- struct{}{}:
	default:
	}
}

func (c condition) C() <-chan struct{} {
	return c
}
```

this does require to change the usage to this:

```
    var cond = newCondition()
```

or

```
type foo struct {
    cond condition
    ...
}

func newFoo() foo {
   return foo{
       cond: newCondition(),
   }
}

// ...

func main() {
    myFoo := newFoo()
}
```

mem (approved):

PR #492, opened 2023-08-28
Handle variables in multihttp request urls

rdubrock:


internal/prober/multihttp/script.go
```diff
@@ -17,26 +16,31 @@ import (
...
-	// should be valid. This function should never return an error.
-	u, _ := url.Parse(req.Url)
-
 	var buf strings.Builder
 
 	for _, field := range req.QueryFields {
 		if buf.Len() > 0 {
 			buf.WriteByte('&')
 		}
-		buf.WriteString(url.QueryEscape(field.Name))
```
rdubrock, line 30: I've been mulling this since yesterday, and I think query escaping (if it needs to happen), has to happen at runtime in the JS. I'm thinking this because

- If a variable is present in the URL, we don't want to escape the characters the signify a variable (`${}`)
- If a variable is present in the URL and resolves to a value that needs to be escaped, we'll need to do that after the variable has a value. So the output JS would need to be something like

```
const foo = "#(%#*)"
const url = "https://www.example.com/${foo}"

http.post(escapeString(url))
```

Instead of doing it ahead of time when building the script

internal/prober/multihttp/script.tmpl
```diff
@@ -67,19 +68,22 @@ function assertHeader(headers, name, matcher) {
...
 
 	{{ range .Entries }}
 	{{- $url := buildUrl .Request }}
+	{{- $urlString := buildUrlString .Request }}
 	{{- $method := .Request.Method.String }}
 	{{- $body := buildBody .Request.Body }}
 	{{- $headers := buildHeaders .Request.Headers .Request.Body }}
-	response = http.request('{{ $method }}', '{{ $url }}', {{ $body }}, {
+	url = {{ $url }}
+	response = http.request('{{ $method }}', url, {{ $body }}, {
```
rdubrock, line 81:
```suggestion
	response = http.request('{{ $method }}', url.toString(), {{ $body }}, {
```

internal/prober/multihttp/script.go
```diff
@@ -17,13 +18,37 @@ import (
...
+func buildUrl(req *sm.MultiHttpEntryRequest) string {
 	var buf strings.Builder
+	url := replaceVariablesInString(req.Url)
+	buf.WriteString("new URL(`")
+	buf.WriteString(url)
+	buf.WriteString("`)")
+	return buf.String()
+}
 
+func buildUrlString(req *sm.MultiHttpEntryRequest) string {
```
rdubrock, line 49: This exists only to provide something for `check` to print in a way that the plugin can query reliably

internal/prober/multihttp/script_test.go
```diff
@@ -10,52 +10,108 @@ import (
...
 				Url: "https://www.example.org/",
+			},
+			expected: "new URL(`https://www.example.org/`)",
+		},
+		"variable in url": {
+			request: sm.MultiHttpEntryRequest{
+				Url: "${variable}",
 				QueryFields: []*sm.QueryField{
 					{
-						Name:  "p&q",
```
rdubrock, line 53: The k6 URL class will escape invalid characters, so it doesn't need to be done here anymore

internal/prober/multihttp/script.go
```diff
@@ -17,13 +18,37 @@ import (
 //go:embed script.tmpl
 var templateFS embed.FS
 
-func buildUrl(req *sm.MultiHttpEntryRequest) string {
-	// If we are here, the request has already been validated, and the URL
-	// should be valid. This function should never return an error.
-	u, _ := url.Parse(req.Url)
+func replaceVariablesInString(s string) string {
+	re := regexp.MustCompile(`\$\{(.*?)\}`)
```
mem, line 22: the capture probably needs to be something like `[a-zA-Z_][a-zA-Z0-9_]*`

internal/prober/multihttp/script.go
```diff
@@ -17,13 +18,37 @@ import (
...
+	variableNames := re.FindAll([]byte(s), -1)
+	transformed := s
+	if len(variableNames) > 0 {
+		for _, variableName := range variableNames {
+			variableString := string(variableName)
+			strippedVarName := strings.Replace(variableString, "${", "", 1)
+			strippedVarName = strings.Replace(strippedVarName, "}", "", 1)
```
mem, line 29: the capture is the variable name, so if the input is `foo${bar}baz`, the capture is `bar`.

```
strings.ReplaceAll(input, "${" + name + "}", `${vars[` + name + `]}`)
```

should do the trick, no?

internal/prober/multihttp/script.go
mem: given some input like `foo${bar}baz${bar}` this will return `bar` twice.

This should take care of that:

```
slices.Sort(names)
names = slices.Compact(names)
```

internal/prober/multihttp/script.tmpl
```diff
@@ -67,19 +68,24 @@ function assertHeader(headers, name, matcher) {
...
 
 	{{ range .Entries }}
 	{{- $url := buildUrl .Request }}
+	{{- $urlString := buildUrlString .Request }}
+	{{- $queryParams := buildQueryParams .Request }}
 	{{- $method := .Request.Method.String }}
 	{{- $body := buildBody .Request.Body }}
 	{{- $headers := buildHeaders .Request.Headers .Request.Body }}
-	response = http.request('{{ $method }}', '{{ $url }}', {{ $body }}, {
+	url = {{ $url }}
```
mem, line 81: I know javascript is not my thing, but shouldn't this be `let url = {{ $url }}`?

internal/prober/multihttp/script.go
```diff
@@ -34,12 +59,29 @@ func buildUrl(req *sm.MultiHttpEntryRequest) string {
...
 	return template.JSEscapeString(u.String())
 }
 
+func buildQueryParams(req *sm.MultiHttpEntryRequest) string {
+	var buf strings.Builder
+	for i, field := range req.QueryFields {
+		if i > 0 {
+			buf.WriteRune('\t')
+		}
+		buf.WriteString("url.searchParams.append(`")
```
mem, line 73: the string `url` should be an argument to this function, so that the expectation that there's a variable called `url` in the context where this is being called is not so strong.

```
func buildQueryParams(url string, req ...)
```

internal/prober/multihttp/script.tmpl
mem: or possibly const?

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -286,6 +286,9 @@ func (c Check) validateTarget() error {
 	case CheckTypeMultiHttp:
 		// TODO(mem): checks MUST have a target, but in this case it's
 		// not true that the target must be a valid URL.
+		if strings.Contains(c.Target, "${") {
```
mem, line 289: why would the target of the check contain a replacement?

mem (changes requested):

internal/prober/multihttp/script.tmpl
rdubrock: It's declared on line 71 already, so this line is just overwriting the reference, not creating a new one. And that's also why we can't use `const`, because the thing that makes it `const` is that you can't reassign it's value

rdubrock: To be clear, I wouldn't actually write JS this way, but otherwise we'd need to give the url variable a unique name for each request

pkg/pb/synthetic_monitoring/checks_extra.go
rdubrock: I'm sure there's a better way to do this, but basically I want the URL to be validated _unless_ we detect a variable as part of the string. And if it _is_ part of the string, just bail out on validating altogether

internal/prober/multihttp/script.go
rdubrock: I don't understand why, but the capture is is actually `${bar}` right now

pkg/pb/synthetic_monitoring/checks_extra.go
rdubrock: If you mean from a user perspective, interacting with a rest api could end up needing a variable in the url. 

- Make a request for a list of products to `https://www.russstore.com/products/list`
- Extract an individual product from the list and save it to a variable
- Make a request for the specific product at `https://www.russtore.com/products/${productId}`

Something like that

internal/prober/multihttp/script.tmpl
```diff
@@ -67,19 +68,24 @@ function assertHeader(headers, name, matcher) {
 
 export default function() {
 	let response;
+	let url;
 	const vars = {};
 
 	{{ range .Entries }}
 	{{- $url := buildUrl .Request }}
+	{{- $urlString := buildUrlString .Request }}
```
mem, line 76: OK, I went over this several times, and I don't understand it.

buildURL is building a string that can be evaluated as Javascript, but the new version is _not_ including the query.

buildUrlString is building the URL _with_ the query added in.

`$url` is then fed to the request and `$urlString` is used in the assertions.

I see two optoins:

- the assertions generate metrics that use the URL as originally entered by the user
- the assertions generate metrics that use the effective URL that was used to make the request

rdubrock: > the assertions generate metrics that use the URL as originally entered by the user

That was sort of my thinking. The reason being:

- It means on the frontend I can query by URL predictably
- It will keep the metrics/logs for an individual step in grouped by the step. If the variable results in a different value every request, and therefore a different URL, there's no way for me to know "this assertion is the one that belongs to the third request", etc

mem (approved):

PR #550, opened 2023-10-30
Add an option to log multihttp responses

rdubrock:
One of the main points of feedback we've gotten is that it would be nice to have more visibility into what's happening during the request. Logging response bodies is tricky because the can be 1.) gigantic or 2.) contain sensitive information. However, it would be super useful for debugging 

I feel that by making logging response bodies an opt in option on a per request basis, and limiting to 1000 characters would give reasonable protection from both of these problems. 

Screenshot 2023-10-30 at 13 11 15


pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -462,9 +462,10 @@ message MultiHttpSettings {
 
 // MultiHttpEntry represents a single entry in a MultiHttp check.
 message MultiHttpEntry {
-  MultiHttpEntryRequest            request    = 1 [(gogoproto.jsontag) = "request,omitempty"];    // The request parameters.
-  repeated MultiHttpEntryAssertion assertions = 2 [(gogoproto.jsontag) = "checks,omitempty"];     // Zero or more assertions to be made on the response.
-  repeated MultiHttpEntryVariable  variables  = 3 [(gogoproto.jsontag) = "variables,omitempty"];  // Zero or more variables to be used in the request.
+  MultiHttpEntryRequest            request     = 1 [(gogoproto.jsontag) = "request,omitempty"];    // The request parameters.
+  repeated MultiHttpEntryAssertion assertions  = 2 [(gogoproto.jsontag) = "checks,omitempty"];     // Zero or more assertions to be made on the response.
+  repeated MultiHttpEntryVariable  variables   = 3 [(gogoproto.jsontag) = "variables,omitempty"];  // Zero or more variables to be used in the request.
+  bool                             logResponse = 4 [(gogoproto.jsontag) = "logResponse,omitempty"];  // Whether to log the response body in the output.
```
mem, line 468: Since there might be multiple requests, wouldn't it be better to call the field `LogResponseBodies` or `LogResponses`. I'm slightly in favor of the former, because it specifies what exactly will get added.

rdubrock: I was thinking this would be an option on a per request basis instead of all or nothing

rdubrock: Could be kind of a pain to specify for every request, though

mem: I thought about that.

Why would anyone want to have the body to be logged? Mostly for debugging. In that case, I imagine they want all the responses.

Why would anyone _not_ want a specific response logged? Because there's something sensitive in there that they don't want uploaded to Loki. In that case, why would they want the _other_ responses logged?

If this is for debugging (the test button), the frontend can rework the request so that it includes the option in each response.

Also, the frontend could have a "log all responses" button that causes all the individual options to be set.

rdubrock: Ok yeah I'm convinced. I'll rework it to be a global option

mem (approved):

PR #555, opened 2023-11-01
Fix: add a `name` label to metrics

mem:
By adding a `name` tag to each request, we get that in the result as a `name` label with the corresponding value.

Since the [k6
documentation](https://k6.io/docs/using-k6/http-requests/#url-grouping) says that setting `name` should not affect `url` but [in reality it does](https://github.com/grafana/k6/issues/3435), implement a nasty workaround to preserve the URL. If there's a tag with the special name `__raw_url__`, use that as the actual URL that was requested.

Modify the generated script so that we add both the `name` tag (set to a sequential number based on the position of the request in the entries list) and a `__raw_url__` tag (workaround). Then in the output extension, include `name` in the metric labels and convert `__raw_url__` to `url`.

### DIFF
--- a/xk6/sm/output.go
+++ b/xk6/sm/output.go
@@ -15,7 +15,10 @@ import (
 	"go.k6.io/k6/output"
 )
 
-const ExtensionName = "sm"
+const (
+	ExtensionName = "sm"
+	RawURLTagName = "__raw_url__"
+)
 
 func init() {
 	output.RegisterExtension(ExtensionName, New)
@@ -209,6 +212,7 @@ type targetId struct {
 	method   string
 	scenario string
 	group    string
+	name     string
 }
 
 type targetMetrics struct {
@@ -243,10 +247,11 @@ func newTargetMetricsCollection() targetMetricsCollection {
 
 func (collection targetMetricsCollection) Update(sample metrics.Sample, scenario, group string, tags map[string]string) {
 	key := targetId{
-		url:      tags["url"],
+		url:      getURL(tags),
 		method:   tags["method"],
 		scenario: scenario,
 		group:    group,
+		name:     tags["name"],
 	}
 
 	// the metrics for this target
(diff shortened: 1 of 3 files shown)


xk6/sm/output.go
```diff
@@ -15,7 +15,10 @@ import (
 	"go.k6.io/k6/output"
 )
 
-const ExtensionName = "sm"
+const (
+	ExtensionName = "sm"
+	RawURLTagName = "__raw_url__"
```
rdubrock, line 20: I think maybe the double underscores are forbidden in prom? When I run it I get an error that it's an invalid label name. If I change it to a random string it seems to work fine

rdubrock:
So in my experiments, this results in metrics that look like this:

```
probe_http_total_duration_seconds{probe=~".*", job="tests", instance="https://swapi.dev/api/planets/1", url="0", method="GET"}
```

The url is the index, and there is no `name` label getting sent. This works for me in the plugin, but I wonder whether we should keep the url as the url and add the index somewhere else? Could maybe be confusing for people querying the metrics directly

xk6/sm/output.go
mem: that's weird, I'm replacing that with `url`.

Am I missing some case where the replacement doesn't happen?

mem:
> So in my experiments, this results in metrics that look like this:
> 
> ```
> probe_http_total_duration_seconds{probe=~".*", job="tests", instance="https://swapi.dev/api/planets/1", url="0", method="GET"}
> ```
> 
> The url is the index, and there is no `name` label getting sent. This works for me in the plugin, but I wonder whether we should keep the url as the url and add the index somewhere else? Could maybe be confusing for people querying the metrics directly

I'm doing something wrong then.

The intention is for that to come out as

```
probe_http_total_duration_seconds{probe=~".*", job="tests", instance="https://swapi.dev/api/planets/1", url="https://swapi.dev/api/planets/1", name="0", method="GET"}
```

I'm using `name` instead of something else because that gives me compatibility with people in the future writing scripts that set `name`. We can document that `name` will be used to group metrics if desired.

xk6/sm/output.go
mem: this is weirder... if I run it "by hand" (which is what I have been doing all along), it's doing the right thing.

Why is it doing something when running from the agent?

mem: OK, we figured this out. It's working as expected.

mem:
> I'm doing something wrong then.

It's working as expected.

rdubrock (approved):

PR #570, opened 2023-11-21
examples/k8s: listen on all addresses to make health/readiness eps reachable

roobre:
Currently, the agent spawns its HTTP server listening in `localhost` only:
https://github.com/grafana/synthetic-monitoring-agent/blob/1a7c9f62ab0332e4292e791dfdf6f170d5304313/cmd/synthetic-monitoring-agent/main.go#L54

This setting is at odds with the example k8s deployment, which defines readiness and healthiness probes:
https://github.com/grafana/synthetic-monitoring-agent/blob/1a7c9f62ab0332e4292e791dfdf6f170d5304313/examples/kubernetes/deployment.yaml#L42-L49

By default, the kubelet will attempt to connect to the podIP, and will get a connection refused error as the HTTP server is listening in localhost only. As a result, the kubelet will periodically kill the agent pod as it think it is not responding.

This PR modifies the agent command so it listens on all addresses, making probes reachable by the kubelet.

### DIFF
--- a/examples/kubernetes/deployment.yaml
+++ b/examples/kubernetes/deployment.yaml
@@ -21,7 +21,7 @@ spec:
     spec:
       containers:
         - args:
-            - /usr/local/bin/synthetic-monitoring-agent --api-server-address=${API_SERVER} --api-token=${API_TOKEN} --verbose=true
+            - /usr/local/bin/synthetic-monitoring-agent --api-server-address=${API_SERVER} --api-token=${API_TOKEN} --verbose=true --listen-address=0.0.0.0:4050
           command:
             - sh
             - -c


examples/kubernetes/deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
     spec:
       containers:
         - args:
-            - /usr/local/bin/synthetic-monitoring-agent --api-server-address=${API_SERVER} --api-token=${API_TOKEN} --verbose=true
+            - /usr/local/bin/synthetic-monitoring-agent --api-server-address=${API_SERVER} --api-token=${API_TOKEN} --verbose=true --listen-address=0.0.0.0:4050
```
mem, line 24:
```suggestion
            # Please note that by listening on all interfaces (passing 0.0.0.0 to --listen-address) anyone with access
            # to the cluster might be able to access information about the agent that might not be available otherwise.
            # The HTTP server is used to present metrics as well as some control and debugging endpoints.
            - /usr/local/bin/synthetic-monitoring-agent --api-server-address=${API_SERVER} --api-token=${API_TOKEN} --verbose=true --listen-address=0.0.0.0:4050
```

roobre: I understand the reason behind adding a warning, but I'm not 100% sure about its shape. I'm afraid users might see it and read it as a "this default is dangerous for some reason and you should change it!". Then, if they do, the deployment breaks.

Do we think this information leak is dangerous enough to warn about it? In my experience as a k8s admin, I mostly expect pods to listen in `0.0.0.0` by default. If I'm worried about that, I should deploy some network policies to harden it.

roobre:
Closing in favor of https://github.com/grafana/synthetic-monitoring-agent/pull/723

PR #581, opened 2023-12-19
Add tenant label limits

ka3de:
This PR modifies the synthetic monitoring agent implementation so label limits applied when collecting metrics and logs data are based on the specific tenant limits instead of hardcoded defaults. The reason for this is because the label limits applied in Mimir and Loki are defined per tenant, which means that for some tenants the defaults currently in use in the agent are lower than the max cap they have available, truncating in that case the applied labels unnecessarily.

Updates grafana/synthetic-monitoring#5


pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -122,8 +122,6 @@ const (
 )
 
 const (
-	MaxMetricLabels        = 20  // Prometheus allows for 32 labels, but limit to 20.
-	MaxLogLabels           = 15  // Loki allows a maximum of 15 labels.
```
mem, line 126: This is technically a breaking change, because `synthetic_monitoring.MaxMetricLabels` is being removed.

I think you are correct in that `limits.MaxMetricLabels` reads better.

Since `limits` already imports this package, I think we can leave this here, and have something like:

```
   MaxMetricLabels = sm.MaxMetricLabels
```

in the `limits` package. We could add a deprecation notice here.

mem (commented):
this looks good, thanks!

the only thing I think needs fixing is the breaking of the API.

We have a handful of clients for it, and since pkg/pb/synthetic_monitoring is part of the public API, we cannot change that.

pkg/pb/synthetic_monitoring/checks_extra.go
mem: Actually no, we could deprecate it like this, but `limits` is an internal package, so it does not make sense.

ka3de: Right, I don't think it makes sense to expose the `TenantLimits` implementation in `pkg`, so probably your first suggestion is the best option, so keeping the constants in `pkg` and reference these from our internal implementation, with no actual need to deprecate it in this case.

Actually my initial plan was also to move all the other [limit constants](https://github.com/grafana/synthetic-monitoring-agent/blob/ac7b0c0778f6679a6c8763d772cb5e40b271e454/pkg/pb/synthetic_monitoring/checks_extra.go#L125-L131) to the `limits` package, but that was not possible due to an import cycle (as the `TenantManager`/`TenantProvider` that we use in the tenant limits implementation requires a `*sm.TenantInfo` as input). So keeping the limit constants where they currently are might be the best path forward.

mem (approved):

PR #596, opened 2024-01-31
Fix: make label names uniform

mem:
Our use of labels names in logs and metrics is not uniform and this is causing a lot of pain.

We have `foo_id`, `fooID` and other variations.

Pick one.

It's `fooId` from now on.

It's going to be painful to change everywhere else to match this, but it's better than dealing with the existing pain.

### DIFF
--- a/internal/checks/checks.go
+++ b/internal/checks/checks.go
@@ -612,7 +612,7 @@ func (c *Updater) handleCheckUpdateWithLock(ctx context.Context, check model.Che
 
 	scraper, found := c.scrapers[cid]
 	if !found {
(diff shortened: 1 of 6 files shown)


rdubrock (approved):

roobre:
So from now on we use camelCase for log labels, but we should keep using snake_case for prom labels, right?

mem:
> So from now on we use camelCase for log labels, but we should keep using snake_case for prom labels, right?

I didn't think that far, but sounds reasonable.

Basically I looked at what would introduce the least changes, and camel case for log labels won.

roobre (approved):

PR #601, opened 2024-02-05
add the check ID to the sm_check_info metric

rdubrock:
Not sure if this is OK. It would help me out a lot on the visualizing side to be able to extract the check ID from prometheus queries for building links and things like that.

### DIFF
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -647,6 +647,7 @@ func (s Scraper) buildCheckInfoLabels(userLabels []labelPair) map[string]string 
 		"region":     s.probe.Region,
 		"frequency":  strconv.FormatInt(s.check.Frequency, 10),
 		"geohash":    geohash.Encode(float64(s.probe.Latitude), float64(s.probe.Longitude)),
+		"check_id":   fmt.Sprint(s.check.Id),
 	}
 	if s.check.AlertSensitivity != "" && s.check.AlertSensitivity != "none" {
 		labels["alert_sensitivity"] = s.check.AlertSensitivity
--- a/internal/scraper/scraper_test.go
+++ b/internal/scraper/scraper_test.go
@@ -1045,6 +1045,7 @@ func (p testProber) Probe(ctx context.Context, target string, registry *promethe
 //nolint:gocyclo
 func TestScraperCollectData(t *testing.T) {
 	const (
+		checkId       = 1
 		checkName     = "check name"
 		checkTarget   = "target name"
 		frequency     = 2000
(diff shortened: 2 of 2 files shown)


mem:
> Not sure if this is OK. It would help me out a lot on the visualizing side to be able to extract the check ID from prometheus queries for building links and things like that.

While I understand why this is, there are three problems that I can see:

- we are exposing an internal ID. We are already doing plenty of this, but nevertheless...
- we are adding a label, and we already have some pressure on that resource, because sm_check_info is where we add user-defined labels

Those two could be ignored / negotiated.

The one that isn't as easy is that the way the existing queries are written, this could break them. This happened when I tried to add an attribution label to the metrics we generate.

The job/instance pair is unique and assigned to each check. The frontend has access to the list of all checks and their IDs. I know this isn't pretty, but it could build a job/instance <-> id mapping.

If you can describe the problem, maybe we can find a different solution...

rdubrock:
That all makes sense. There are ways I can hack around this on the frontend. The problem I'm bumping into is that I want to rework our navigation model, so that in order to get to the view for a specific check it would follow this convention:

```
/a/grafana-synthetic-monitoring-app/checks/<check id>
```

That way I can pluck the ID out of the URL and use it to find out the job, instance, etc. This works fine as is, the only problem is from the summary dashboard, I have to use dashboard data links to link from the summary to an individual check. The data has to be present in the response from a prometheus query. It's very possible I can issue another query to either a custom runtime datasource or the synthetics datasource itself, it's just messier. Or I can just build a custom table to replace the built in table that dashboards use. Another alternative would be to swap the URL convention to be:

```
/a/grafana-synthetic-monitoring-app/checks/<job>/<instance>
```

I don't love that, because weird things could happen when someone edits a check and changes the job/instance values. 

So all in all, including the ID in metrics is not a _have to have_ it's just a shortcut

mem:
can you have a redirect?

if the data link sends you to:

```
/a/grafana-synthetic-monitoring-app/checks/<job>/<instance>
```

then that generates a redirect to:

```
/a/grafana-synthetic-monitoring-app/checks/<check id>
```

That solves the edit issue.

The potential problem I see with HTTP checks is that some people are using really long URLs, but I don't think that's an actual problem.

rdubrock:
> can you have a redirect?

That's a good idea. The only thing there dealing with the navigation history then. We used to do a similar thing for dashboards, but it broke the back button (i.e. if you got redirected to a dashboard from the plugin, and then hit the back button, you'd just land on the redirect page which would shoot you right back to the dashboard). That's probably avoidable though

rdubrock:
I found a way to shoehorn the ID in there by including a query to the synthetic monitoring datasource and doing some transforms in the dashboard. So we're all good, I don't think I'll actually need this

PR #609, opened 2024-02-08
Update proto definition to send stats

mem:


pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -575,3 +578,18 @@ message GrpcSettings {
   bool                      tls             = 3 [(gogoproto.jsontag) = "tls,omitempty"];
   TLSConfig       tlsConfig           = 4 [(gogoproto.jsontag) = "tlsConfig,omitempty"];
 }
+
+// CheckStats provides information about the executed checks.
+message CheckStats {
+  string                      instance           = 1 [(gogoproto.jsontag) = "instance"];
```
mem, line 584: This is necessary to avoid mixing up stats from different agent instances. If these were regular Prometheus metrics, this would be added by the scraper.

pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -575,3 +578,18 @@ message GrpcSettings {
...
+  int32                       regionId           = 2 [(gogoproto.jsontag) = "region_id"];
```
mem, line 585: Strickly speaking this is not needed here, but it makes the work in the proxy easier, as the agent can split the stats by the tenant's region. This allows the proxy to just forward the messages instead of having to open them up and rewrite them.

ka3de:
I've been working on this here: https://github.com/grafana/synthetic-monitoring-agent/compare/main...feat/add-telemetry

pkg/pb/synthetic_monitoring/checks.proto
ka3de: This can make sense.
I see that based on your definition your idea is to, for every push, send the telemetry for all tenants that belong to a specific region. In my case my approach was to have one pusher loop per tenant to avoid sending all the agent's telemetry in a single request and be able to distribute the requests more evenly, as the loop time span has a jitter; and also we could log failed requests tight to a specific tenant. Let me know if you have some comments on the differences between the two approaches.

mem: > This can make sense. In that case we will have to send the local tenant ID as the tenantId in the message to avoid modifying that in the proxy also, is that correct? Could this create some "confusion"? As the approach seems to use global IDs in the agents, and delegate the translation to the proxy?

That's a good question.

The tenantId below would be a local ID, not a global one.

The regionId would be extracted from the tenant, and for agents that aren't running behind the proxy, this would be 0. The API can ignore this part of the message, because when the message arrives to it, it's already a local ID that it knows about.

PR #611, opened 2024-02-09
Feature: allow checks to run less often

mem:
This change allows a check to run less often by storing the produced metrics and pushing the last set that was produced by the scraper once every 2 minutes.

### DIFF
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -35,6 +35,8 @@ const (
 	CheckInfoMetricName    = "sm_check_info"
 	CheckInfoSource        = "synthetic-monitoring-agent"
 	maxLabelValueLength    = 2048 // this is the default value in Prometheus
+	maxPublishInterval     = 2 * time.Minute
+	minPublishGap          = 10 * time.Second
 )
 
 var (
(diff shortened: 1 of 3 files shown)


internal/scraper/scraper.go
```diff
@@ -220,74 +222,113 @@ func (s *Scraper) Run(ctx context.Context) {
...
-			for j := range ts.Samples {
-				ts.Samples[j] = staleSample
-			}
+	h.payload.streams = nil // do not republish logs
+
+	now := t.UnixMilli()
+	for i := range h.payload.ts {
+		ts := &h.payload.ts[i]
+		for j := range ts.Samples {
+			ts.Samples[j].Timestamp = now
```
ka3de, line 295: I might be misunderstanding something here, but, if I'm understanding this correctly, once we want to republish the previous published data, we iterate through all the previously collected time series, and for each sample in the time series we set the current time. That means that, if some of these time series collected for a check have more than one sample, when republishing all of them will have the same timestamp.

Is that correct? In such case, can that produce weird representations of the data? As I assume that the previously generated time series, probably have different timestamp values per sample..?

I wonder if we should define the sample timestamps based on the current time and also on the delta time between the samples for the "original payload", so basically maintaining the time difference between samples. Although for that we would have to, either define timestamps that are "in the future", or do it backwards, so the "oldest" sample for each time series will get the _timestamp=now_ and then we would recalculate the rest subtracting their respective deltas from _now_ back in time.

Let me know if I misunderstood something, or if I did not explain myself properly.

ka3de (commented):
That was nice! :clap: Made a couple of comments.

internal/scraper/scraper.go
mem: No, all the data in payload has the same timestamp (#L259, #L501, #L734, #L793). Queuing older data (or data with different timestamps) happens in the publisher, not in the scraper. The data collected in the scraper corresponds to a single point in time.

What you are saying is possible, but it doesn't happen here.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -138,6 +138,12 @@ const (
 	MaxMultiHttpTargets      = 10   // Max targets per multi-http check.
 	MaxMultiHttpAssertions   = 5    // Max assertions per multi-http target.
 	MaxMultiHttpVariables    = 5    // Max variables per multi-http target.
+
+	// Frequencies (in milliseconds)
+	MaxCheckFrequency      = 120 * 1000 // Maximum value for the check's frequency.
```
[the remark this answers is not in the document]
mem: that's correct

internal/scraper/scraper.go
ka3de: Got it! Thanks for the explanation.

ka3de (approved):

Issue #597, opened 2024-02-01
Agent should validate validations


mem:
It was reported that the agent is accepting regular expression validations with invalid regular expressions.

E.g. if fail_if_body_matches_regexp contains an invalid regular expression, the agent will consider the check to be valid.

That validation should be happening somewhere around here: https://github.com/grafana/synthetic-monitoring-agent/blob/main/pkg/pb/synthetic_monitoring/checks_extra.go#L574

PR #612, opened 2024-02-12
Fix: missing http check regex validations

The-9880:
Adds back regex validations on the fields of an HTTP check which use them, such as `fail_if_body_matches_regexp`.

It looks like the only other place these fields get validated is [here](https://github.com/grafana/synthetic-monitoring-agent/blob/a02485f1054df991c0e8289c6893650f6190ce77/internal/prober/http/http.go#L101) (converting `HttpSettings` to BBE config module). From [this line](https://github.com/grafana/synthetic-monitoring-agent/blob/a02485f1054df991c0e8289c6893650f6190ce77/internal/scraper/scraper.go#L136), I figure if the validation doesn't go through when creating the HTTP Probe, then a `Scraper` isn't created for that check/probe and it doesn't get executed (@mem suggested this in the original support issue).

Fixes https://github.com/grafana/synthetic-monitoring-agent/issues/597.

### DIFF
--- a/pkg/pb/synthetic_monitoring/checks_extra.go
+++ b/pkg/pb/synthetic_monitoring/checks_extra.go
@@ -25,6 +25,7 @@ import (
 	"mime"
 	"net"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -633,6 +638,34 @@ func (s *HttpSettings) Validate() error {
 		}
 	}
 
+	for _, reg := range s.FailIfBodyMatchesRegexp {
```
roobre, line 641: CI seems to be complaining about the increased cyclomatic complexity, probably because it sees to many `for` loops and gets scared. Perhaps we can `//nolint` this one out, or alternatively `append` all regexes together and return the same error in case any group is wrong.

The-9880: I'm inclined to `//nolint:gocyclo` this one since the specific error message seems nice to have.

I'll tinker around to see if there's an easy way to do it in a loop (maybe using a `map[error][]string`) but having the mix of `[]string` and `[]HeaderMatch` might complicate that.

roobre: That's a good point. I'd probably `nolint` it away instead of building a complex workaround 😃

roobre (approved):

PR #615, opened 2024-02-12
Feature: promote adhoc to permanent feature

mem:
This has caused a number of questions recently because the "test" functionality requires the "adhoc" feature to be enabled.

We have had this feature for a long time now and we are actively using it. Make it a permanent feature by removing the checks for whether or not it's enabled. Keep the feature flag in case someone is passing it on the command line to avoid breaking those set ups.

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -131,6 +131,8 @@ func run(args []string, stdout io.Writer) error {
 		Bool("pprof-enabled", *enablePProf).
 		Msg("starting")
 
+	notifyAboutDeprecatedFeatureFlags(features, zl)
+
 	if features.IsSet(feature.K6) {
 		newUri, err := validateK6URI(*k6URI)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ka3de (approved):
That makes sense.

nit: Should we remove the constants for `AdHoc` and `Traceroute`, which seem are no longer used?
https://github.com/grafana/synthetic-monitoring-agent/blob/e7bc645cd13675ae644affb05f483d77b2669530/internal/feature/feature.go#L13-L14

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -244,26 +244,24 @@ func run(args []string, stdout io.Writer) error {
...
-			K6Runner:       k6Runner,
-		})
-		if err != nil {
-			return fmt.Errorf("Cannot create ad-hoc checks handler: %w", err)
-		}
-
-		g.Go(func() error {
-			return adhocHandler.Run(ctx)
-		})
+	adhocHandler, err := adhoc.NewHandler(adhoc.HandlerOpts{
```
roobre, line 247: Should we print a simple warning saying it is now deprecated and has no effect? It can help as a reminder for operators to garbage-collect the flag.

mem: good point, thanks

PR #621, opened 2024-02-15
Add telemetry

ka3de:
This PR adds a telemetry subsystem responsible to collect telemetry data for check executions in the agent.

Currently the telemetry data covers executions and executions duration, and is organized per region, tenant and check class. Each region has an independent goroutine which periodically (by default every 5 min) sends the accumulated telemetry data to the proxy/API.

Updates: grafana/synthetic-monitoring#13

### DIFF
--- a/internal/telemetry/region_pusher.go
+++ b/internal/telemetry/region_pusher.go
@@ -74,6 +74,7 @@ LOOP:
 			p.logger.Debug().Msg("region pusher stopping")
 			m := p.next()
 			p.push(m)
+			ticker.Stop()
 			break LOOP
 		}
 	}
--- a/internal/telemetry/region_pusher.go
+++ b/internal/telemetry/region_pusher.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
+
+	prom "github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog"
 )
 
@@ -19,6 +21,31 @@ type RegionPusher struct {
 
 	telemetry   map[int64]map[sm.CheckClass]*sm.CheckClassTelemetry // Indexed by local tenant ID
 	telemetryMu sync.Mutex
+
+	metrics RegionMetrics
+}
+
+type RegionMetrics struct {
+	pushRequestsActive   prom.Gauge
+	pushRequestsDuration prom.Observer
+	pushRequestsTotal    prom.Counter
+	pushRequestsError    prom.Counter
+}
+
+// start handles region metrics before a push telemetry request.
+func (m *RegionMetrics) start() (start time.Time) {
+	m.pushRequestsActive.Inc()
+	m.pushRequestsTotal.Inc()
+	return time.Now()
+}
+
+// end handles region metrics after a push telemetry request.
+func (m *RegionMetrics) end(err error, start time.Time) {
+	m.pushRequestsActive.Dec()
+	m.pushRequestsDuration.Observe(time.Since(start).Seconds())
+	if err != nil {
+		m.pushRequestsError.Inc()
+	}
 }
 
 // NewRegionPusher builds a new RegionPusher.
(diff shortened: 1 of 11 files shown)


internal/telemetry/region_pusher.go
```diff
@@ -0,0 +1,147 @@
...
+		p.logger.Err(err).Msg("error pushing telemetry")
+		return
+	}
+	if r.Status.Code != sm.StatusCode_OK {
+		p.logger.Error().
+			Int32("statusCode", int32(r.Status.Code)).
+			Str("statusMessage", r.Status.Message).
+			Msg("error pushing telemetry")
+	}
+}
```
mem, line 147: this might benefit from some observability:

- keep track of the number of pushes
- keep track of the duration of those pushes
- keep track of the number of errors

internal/telemetry/region_pusher.go
```diff
@@ -0,0 +1,147 @@
...
+			p.logger.Debug().Msg("region pusher stopping")
+			m := p.next()
+			p.push(m)
+			break LOOP
+		}
+	}
+}
+
+// AddExecution adds a new execution to the tenant telemetry.
+func (p *RegionPusher) AddExecution(e Execution) {
```
mem, line 83: I almost want to ask for some observability here, too.

It should be enough to have a summary, to keep track of the number of calls to this function and their duration. I would prefer a (native) histogram.

The concern is that I suspect there might be a lot of contention for this lock, but it doesn't make sense to worry about that without some data. The happy path seems safe enough (O(1)).

internal/telemetry/ticker.go
```diff
@@ -0,0 +1,40 @@
...
+const (
+	jitterUpperBound = 60 // s
+)
+
+// ticker represents a ticker interface that
+// can be mocked for more reliable tests.
+type ticker interface {
+	C() <-chan time.Time
+}
+
```
mem, line 17: I'm really missing Stop() here.

mem: (in the interface, I mean)

internal/telemetry/region_pusher.go
```diff
@@ -0,0 +1,147 @@
...
+		select {
+		case <-ticker.C():
+			p.logger.Info().Msg("pushing telemetry")
+			m := p.next()
+			go p.push(m) // Avoid blocking
+		case <-ctx.Done():
+			p.logger.Debug().Msg("region pusher stopping")
+			m := p.next()
+			p.push(m)
+			break LOOP
```
mem, line 77: this is where I would like to see a call to ticker.Stop().

It seems a bit weird to stop the ticker here, but I don't see where else it can go given that it's an optional argument to NewRegionPusher and it can be created there if it's not provided.

mem (commented):
looks good to me at first pass.

internal/telemetry/ticker.go
ka3de: That is right, added it in [7fe148b](https://github.com/grafana/synthetic-monitoring-agent/pull/621/commits/7fe148b383301d34e30a47fcd0b48a159d411c00).

internal/telemetry/region_pusher.go
ka3de: Yeah, I think that's probably the best place, also I think it's better to call it at that point instead of through a `defer` at the beginning of the function, considering that it is initialized elsewhere. See [7fe148b](https://github.com/grafana/synthetic-monitoring-agent/pull/621/commits/7fe148b383301d34e30a47fcd0b48a159d411c00).

internal/telemetry/region_pusher.go
ka3de: Right, I hesitated to add it because this is being measured from the proxy and this one is executed within the same cluster and namespace, but it makes sense to measure it from the agent side. See [f622450](https://github.com/grafana/synthetic-monitoring-agent/pull/621/commits/f62245037d2b93878abc6418f3a4a6d87e96e4b6).

internal/telemetry/region_pusher.go
ka3de: That makes total sense, let's add it so we can measure and adjust if required.
See [7c78455](https://github.com/grafana/synthetic-monitoring-agent/pull/621/commits/7c78455a4aef339e3cd9031d8180f807a045d635). Even though we are using native histogram, I did set also the regular buckets defining custom values that seem sensible for our needs, as we are not measuring "network requests" here, so the default buckets are probably off for this case. Feedback is welcome in that aspect. Nevertheless we can adjust once we test this at scale.

mem: I need to validate that the set up in the POP clusters is actually scraping native histograms, I don't think that's the case. The Grafana Agent needs to have its configuration in flow mode in order to support that.

ka3de:
Rebased on top of main and resolved conflicts.

mem (approved):

PR #622, opened 2024-02-15
change the name of the k6 check type to scripted

rdubrock:


ka3de (approved):
LGTM.
I was reviewing where do we still have mentions to `k6` in the codebase, and saw that there are [some references](https://github.com/grafana/synthetic-monitoring-agent/blob/be5235192b4d2468c719079f8c73acfbfc050af1/cmd/synthetic-monitoring-agent/main.go#L136-L152) in `main` due to the _"k6 feature"_ with log messages such as "enabling k6 checks", but that applies for `multihttp` and `scripted`, so I think that probably still makes sense and the PR is good as is.

Issue #325, opened 2022-08-01
Add additional information to HTTP checks to make it easier to track


mem:
Currently we add a custom User-Agent header to make it easier for people on the receiving end of these checks to identify the source as Synthetic Monitoring.

To better handle abuse requests, we should add more information (possibly in the form of headers) to the request so that it's simpler and faster to track the issue down to specific users, e.g.

X-SM-Check-Id: 123
X-SM-Tenant-Id: 1000 (this is redundant)
X-SM-Region: prod-us-central-0

we need to balance this against adding too much information (increasing the payload), so maybe:

X-SM-Id: prod-us-central-0/1000/123

simonc6372:
I think adding a unique checkID, that includes a number for the probe ID (not a string) , shoudl be sufficiant. 
`X-SM-Id: nnnnnnnn-xxx `
were `nnnnnnnn` is the checId, and `xxx` is the probe ID. 

There should be an (internal only) API where the checkID can be translated into the tennant/customer details.

mem:
This needs to include the region ID (or use the "global" check ID).

And I think Simon is right, the global check ID (which allows us to identify the API instance) + the probe ID should be more than enough.

peterschretlen:
We would also need to consider this for k6-backed checks

peterschretlen:
Setting this back to triage for review - isn't this fixed by https://github.com/grafana/synthetic-monitoring-agent/pull/624 ?

peterschretlen:
Idea: could we put everything into the user agent string?

Pokom:
> Idea: could we put everything into the user agent string?

If we stuff everything into the user agent string, would that make it trickier for us to track versions of the agent being used?

PR #624, opened 2024-02-16
Add header to ease tracking HTTP requests

The-9880:
Addressing https://github.com/grafana/synthetic-monitoring-agent/issues/325.

The goal is to add a header `x-sm-id` with a value `checkId-probeId` so we can identify the owner of a check for anti-abuse purposes. This header gets added to`http.Prober.config`.

Happy to hear any suggestions on how to test this Go code better - testing `NewProber()` from `http_test.go` works to assert the resulting http config directly. but it'd be nicer to have a test server or something so I can add an HTTP check to a test `Updater` and validate the request headers sent.

### DIFF
--- a/internal/prober/http/http.go
+++ b/internal/prober/http/http.go
@@ -264,7 +264,7 @@ func augmentHttpHeaders(check *sm.Check, reservedHeaders http.Header) {
 	for _, header := range check.Settings.Http.Headers {
 		name, _ := strToHeaderNameValue(header)
 
-		_, present := reservedHeaders[strings.ToLower(name)]
+		_, present := reservedHeaders[http.CanonicalHeaderKey(name)]
 		if present {
 			continue // users can't override reserved headers with their own values
 		}
@@ -272,8 +272,15 @@ func augmentHttpHeaders(check *sm.Check, reservedHeaders http.Header) {
 		headers = append(headers, header)
 	}
 
-	for header, values := range reservedHeaders {
-		headers = append(headers, fmt.Sprintf("%s:%s", header, strings.Join(values, ",")))
+	for key, values := range reservedHeaders {
+		var b strings.Builder
+		for _, value := range values {
+			b.Reset()
+			b.WriteString(key)
+			b.WriteRune(':')
+			b.WriteString(value)
+			headers = append(headers, b.String())
+		}
 	}
 
 	check.Settings.Http.Headers = headers
--- a/internal/prober/http/http.go
+++ b/internal/prober/http/http.go
@@ -263,7 +263,7 @@ func convertOAuth2Config(ctx context.Context, cfg *sm.OAuth2Config, logger zerol
 // Overrides any user-provided headers with our own augmented values
 // for 'reserved' headers.
 func augmentHttpHeaders(check *sm.Check, additionalHeaders []string) {
-	filteredHeaders := []string{}
+	headers := []string{}
 	for _, header := range check.Settings.Http.Headers {
 		name, _ := strToHeaderNameValue(header)
 
@@ -271,11 +271,12 @@ func augmentHttpHeaders(check *sm.Check, additionalHeaders []string) {
 			continue // users can't override this header with their own value.
 		}
 
-		filteredHeaders = append(filteredHeaders, header)
+		headers = append(headers, header)
 	}
+	headers = append(headers, additionalHeaders...)
 
 	httpHeaders := &check.Settings.Http.Headers
-	*httpHeaders = append(filteredHeaders, additionalHeaders...)
+	*httpHeaders = headers
 }
 
 func buildHttpHeaders(headers []string) map[string]string {
--- a/internal/prober/http/http.go
+++ b/internal/prober/http/http.go
@@ -21,6 +21,8 @@ import (
 	"github.com/rs/zerolog"
 )
 
+const checkProbeIdHeader = "x-sm-id"
+
 var errUnsupportedCheck = errors.New("unsupported check")
 
 type Prober struct {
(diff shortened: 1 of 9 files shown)


The-9880:
Re: the header we're adding to HTTP request, do we think there might be a risk that some users are monitoring endpoints with strict header policies (e.g. reject requests with unexpected headers) and we could bust their SM check with this change?

Might be a contrived corner case for most SM cases but just wondering if this wants a config to disable it?

internal/prober/http/http.go
```diff
@@ -21,18 +21,26 @@ import (
...
+
 var errUnsupportedCheck = errors.New("unsupported check")
 
 type Prober struct {
 	config                     config.Module
 	cacheBustingQueryParamName string
 }
 
-func NewProber(ctx context.Context, check sm.Check, logger zerolog.Logger) (Prober, error) {
+func NewProber(ctx context.Context, check sm.Check, logger zerolog.Logger, checkProbeIdentifier string) (Prober, error) {
```
ka3de, line 33: Because we might want to add the `x-sm-id` header to other check types, for example `multihttp`, I wonder if it would make sense to move the constant and header definition to `internal/prober/prober.go` instead (or even in a separate `header.go` file)? In that case we could modify the `NewProber` constructor here to accept something more generic such as `reservedHeaders/systemHeaders http.Header`.

internal/prober/http/http.go
```diff
@@ -252,6 +260,25 @@ func convertOAuth2Config(ctx context.Context, cfg *sm.OAuth2Config, logger zerol
...
 }
 
+// Overrides any user-provided headers with our own augmented values
+// for 'reserved' headers.
+func augmentHttpHeaders(check *sm.Check, additionalHeaders []string) {
+	headers := []string{}
+	for _, header := range check.Settings.Http.Headers {
+		name, _ := strToHeaderNameValue(header)
+
+		if strings.ToLower(name) == checkProbeIdHeader {
```
ka3de, line 270: Following my previous comment, it feels odd to check for the specific `checkProbeIdHeader` here. Would it make sense to accept as input an `http.Header` field, which identifies the "reserved headers", and then, for any header set in the check that is already present in the reserved ones, we forbid it?

internal/prober/http/http.go
```diff
@@ -252,6 +260,25 @@ func convertOAuth2Config(ctx context.Context, cfg *sm.OAuth2Config, logger zerol
...
+			continue // users can't override this header with their own value.
+		}
+
+		headers = append(headers, header)
+	}
+	headers = append(headers, additionalHeaders...)
+
+	httpHeaders := &check.Settings.Http.Headers
+	*httpHeaders = headers
```
ka3de, line 279: I believe this should be equivalent to:
```suggestion
	check.Settings.Http.Headers = headers
```
As Golang [automatically](https://go.dev/tour/moretypes/4) dereferences it.

ka3de (commented):
Thanks for working on this @The-9880 !
Overall it looks good. I made a couple of comments on whether it makes sense to move the header definition "one level up" and accept a more generic argument in the HTTP probe implementation. Let me know what do you think.

internal/prober/http/http.go
The-9880: Liked this a lot -- was initially looking for a way to make the input more generic. Thanks for the reference to `net.http.Header`!

internal/prober/http/http.go
The-9880: Absolutely!

internal/prober/http/http.go
```diff
@@ -252,6 +257,28 @@ func convertOAuth2Config(ctx context.Context, cfg *sm.OAuth2Config, logger zerol
...
+		_, present := reservedHeaders[strings.ToLower(name)]
+		if present {
+			continue // users can't override reserved headers with their own values
+		}
+
+		headers = append(headers, header)
+	}
+
+	for header, values := range reservedHeaders {
+		headers = append(headers, fmt.Sprintf("%s:%s", header, strings.Join(values, ",")))
```
mem, line 276: note that it's OK to have the same header present multiple times

If reservedHeaders has multiple values for a single key, what should be done is add the header multiple times, e.g.

```
for key, values := range reservedHeaders {
   var b strings.Builder
   for _, value := range values {
       b.Reset()
       b.WriteString(key)
       b.WriteRune(':')
       b.WriteString(value)
       headers = append(headers, b.String())
   }
}
```

internal/prober/multihttp/multihttp.go
```diff
@@ -87,3 +93,27 @@ func settingsToModule(settings *sm.MultiHttpSettings) Module {
...
 	return m
 }
+
+// Overrides any user-provided headers with our own augmented values
+// for 'reserved' headers.
+func augmentHttpHeaders(check *sm.Check, reservedHeaders http.Header) {
+	updatedHeaders := []*sm.HttpHeader{}
+	for key, values := range reservedHeaders {
+		updatedHeaders = append(updatedHeaders, &sm.HttpHeader{Name: key, Value: strings.Join(values, ",")})
+	}
```
mem, line 103: the same comment about multiple values for the headers applies here

mem (commented):
It looks good, thanks.

I think there's some minor reorg to be made, but otherwise it's nice!

internal/scraper/scraper.go
```diff
@@ -114,13 +114,14 @@ func (d *probeData) Tenant() model.GlobalID {
...
 ) (*Scraper, error) {
+	checkProbeIdentifier := fmt.Sprintf("%d-%d", check.Id, probe.Id)
 	return NewWithOpts(ctx, check, ScraperOpts{
 		Probe:         probe,
 		Publisher:     publisher,
 		Logger:        logger,
 		ScrapeCounter: scrapeCounter,
 		ErrorCounter:  errorCounter,
-		ProbeFactory:  prober.NewProberFactory(k6runner),
+		ProbeFactory:  prober.NewProberFactory(k6runner, checkProbeIdentifier),
```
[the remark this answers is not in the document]
The-9880: Makes sense! Thanks for read here, I'd be happy to take a look into that issue later

mem (approved):

Issue #587, opened 2024-01-12
k6 checks - private probes can't access IPs in 10.0.0.0/8 private address range


peterschretlen:
When running k6 tests from the agent, we have hardcoded values for `--blacklist-ip` and `--block-hostnames`, and TODO to make these configurable. 

https://github.com/grafana/synthetic-monitoring-agent/blob/99a88d45475d021b7a1cd137f18fd4fdf7144d4b/internal/k6runner/k6runner.go#L409-L410

For private probes in particular we'd want to the option to allow private IP ranges like `10.0.0.0/8`

The-9880:
Fix merged, closing issue.

PR #633, opened 2024-02-26
Configure blacklist ip range for k6 runner

The-9880:
Addresses https://github.com/grafana/synthetic-monitoring-agent/issues/587: makes the blacklist IP range for the K6 runner configurable by flag.

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -208,14 +208,9 @@ func run(args []string, stdout io.Writer) error {
 	defer conn.Close()
 
 	var k6Runner k6runner.Runner
-
 	if features.IsSet(feature.K6) && len(*k6URI) > 0 {
 		if *k6BlacklistedIP != "" {
-			_, _, err := net.ParseCIDR(*k6BlacklistedIP)
-			if err != nil {
-				zl.Warn().Msg("Invalid CIDR IP provided for k6 blacklist; will use default config.")
-				*k6BlacklistedIP = ""
-			}
+			validateBlacklistIp(k6BlacklistedIP, zl)
 		}
 
 		k6Runner = k6runner.New(k6runner.RunnerOpts{
@@ -308,6 +303,14 @@ func newConnectionBackoff() *backoff.Backoff {
 		Max:    30 * time.Second,
 		Factor: math.Pow(30./2., 1./8.), // reach the target in ~ 8 steps
 		Jitter: true,
+	}
+}
+
+func validateBlacklistIp(ip *string, logger zerolog.Logger) {
+	_, _, err := net.ParseCIDR(*ip)
+	if err != nil {
+		logger.Warn().Msg("Invalid CIDR IP provided for k6 blacklist; will use default config.")
+		*ip = ""
 	}
 }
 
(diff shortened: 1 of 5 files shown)


internal/k6runner/k6runner.go
```diff
@@ -26,20 +26,33 @@ type Runner interface {
...
+
 		r = &LocalRunner{
-			k6path: uri,
-			logger: &logger,
-			fs:     afero.NewOsFs(),
+			k6path:        opts.Uri,
+			logger:        &logger,
+			fs:            afero.NewOsFs(),
+			blacklistedIP: blacklistIP,
 		}
```
mem, line 56: I feel this is going to start getting out of hand as we add support for the other TODO items in the way we call the local runner.

What do you think about:

```
r := LocalRunner{
    k6path: "k6",
    logger: &logger,
    fs: afero.NewOsFs(),
    blacklistedIP: blacklistIP,
}

r.withOpts(opts)
```

and then withOpts is something like:

```
func (r *LocalRunner) withOpts(opts RunnerOpts) {
    if opts.blacklistedIPs != "" {
        r.blacklistedIPs = opts.blacklistedIPs
    }
    // ...
}
```

in that way you have a list of default values, and you can override those with options.

The other alternative is that you go with functional options, which I don't like too much to be honest, but it's a common pattern.

mem (commented):
This looks good, thanks!

Please consider the comment I added.

internal/k6runner/k6runner.go
The-9880: I like this suggestion! I'm also not too a huge fan of functional options so this works for me

internal/k6runner/k6runner_test.go
```diff
@@ -23,18 +23,20 @@ import (
...
+	r3 := new("http://localhost:6565", "")
 	require.IsType(t, &HttpRunner{}, r3)
-	r4 := New("https://localhost:6565")
+	r4 := new("https://localhost:6565", "")
 	require.IsType(t, &HttpRunner{}, r4)
 }
 
 func TestNewScript(t *testing.T) {
-	runner := New("k6")
+	runner := new("k6", "")
```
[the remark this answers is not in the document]
The-9880: Good catch, I'd prefer it this way too

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -299,6 +306,14 @@ func newConnectionBackoff() *backoff.Backoff {
 	}
 }
 
+func validateBlacklistIp(ip *string, logger zerolog.Logger) {
```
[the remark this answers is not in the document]
The-9880: The wrapper also served to hide the cyclomatic complexity from the linter 🙈 I like the updated form, but I'll maybe absorb the parent `if` into it so I can keep the linter happy for now.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -54,6 +54,7 @@ func run(args []string, stdout io.Writer) error {
 		enablePProf          = flags.Bool("enable-pprof", false, "exposes profiling data via HTTP /debug/pprof/ endpoint")
 		httpListenAddr       = flags.String("listen-address", "localhost:4050", "listen address")
 		k6URI                = flags.String("k6-uri", "k6", "how to run k6 (path or URL)")
+		k6BlacklistedIP      = flags.String("blocked-nets", "10.0.0.0/8", "blacklisted CIDR IP for k6 requests")
```
mem, line 57:
```suggestion
		k6BlacklistedIP      = flags.String("blocked-nets", "10.0.0.0/8", "IP networks to block in CIDR notation, disabled if empty")
```

mem (approved):

PR #635, opened 2024-02-27
Fix: CheckSettings JSON unmarshal after k6 settings rename

ka3de:
Due to the CheckSettings K6 field renaming to Scripted, previous JSON representations of CheckSettings are incorrectly decoded. This commit overloads the UnmarshalJSON method for CheckSettings in order to transparently translate previous representations that might contain data for the K6 field to the new Scripted field.


ka3de:
Discarded.

PR #641, opened 2024-03-04
Switch to using buf to manage protobuf code

mem:


The-9880 (approved):
This seems nice! I guess we're hoping to stop importing the agent in the `-api` and instead use a dependency on `buf.build/grafana/synthetic-monitoring-agent`?

Was there a discussion on this that I should catch up on?

mem:
> This seems nice! I guess we're hoping to stop importing the agent in the `-api` and instead use a dependency on `buf.build/grafana/synthetic-monitoring-agent`?
> 
> Was there a discussion on this that I should catch up on?

For now I just want to get rid of a custom build process and gain access to the backwards compatibility verification tool.

There wasn't a discussion about this.

PR #648, opened 2024-03-06
Use github.com/grafana/loki/pkg/pusher instead of logproto

mem:
For a long time, we have used an internal copy of Loki's logproto package. At some point Loki moved that to a "push" package, and heavily modified "logproto". That push package is a separate Go module, so we don't need to worry about extra dependencies.

Use that "push" package instead.


roobre (approved):

roobre:
Somewhat interesting how loki defines this data types, which then are sent using a prometheus compatible sender 🤔

It really is like prometheus but for log.

PR #650, opened 2024-03-07
Fix telemetry error handling

ka3de:


internal/telemetry/region_pusher.go
```diff
@@ -173,7 +174,7 @@ func (p *RegionPusher) push(m sm.RegionTelemetry) {
 	)
 
 	start := p.metrics.start()
-	defer p.metrics.end(err, start)
+	defer func() { p.metrics.end(err, start) }()
```
roobre, line 177: Ah, the classic. Been there done that 😅

roobre (approved):

PR #653, opened 2024-03-11
Add telemetry sampled executions

ka3de:
Adds a new `SampledExecutions` field for `CheckClassTelemetry` protobuf message which will hold the sampled executions. Executions are sampled based on 60s "slots".
Example:
```
59s -> 1 sampled execution
60s -> 1 sampled execution
61s -> 2 sampled executions
125 -> 3 sampled executions
```


internal/telemetry/region_pusher.go
```diff
@@ -129,8 +130,11 @@ func (p *RegionPusher) AddExecution(e Execution) {
 		tenantTele[e.CheckClass] = clTele
 	}
 
+	duration := e.Duration.Seconds()
+
 	clTele.Executions++
-	clTele.Duration += float32(e.Duration.Seconds())
+	clTele.Duration += float32(duration)
+	clTele.SampledExecutions += int32(math.Ceil(duration / 60))
```
roobre, line 137: I think there is a way to do this without floating point shenanigans:
https://go.dev/play/p/TDHDpWk_zBv

Now, performance is not critical here but floating point errors might do unexpected things, if you think the integer-based solution is as readable as this is I'd go for that.

roobre (approved):
LGTM!
Left a small suggestion about avoiding floating point, but I don't think it's super critical.

internal/telemetry/region_pusher.go
ka3de: Thanks for the suggestion! Yeah, I think you are right, but I believe this does not work for cases on which `duration % 60 == 0`. For example when a check duration is `60s` we would like to count that as `1` "sampled execution", but following this approach it would count it as `2`. Same happens if we use `duration.Truncate(time.Minute)`.
We could do something such as:
```
clTele.SampledExecutions += int32(e.Duration / time.Minute)
if e.Duration%time.Minute != 0 {
    clTele.SampledExecutions++
}
```
WDYT?

mem: adding the modulo operation and the branching there defeats the purpose of doing integer math.

If you want to go that route, you can use:

```
(t + 1*time.Minute - 1) / time.Minute
```

(the `-1` there is not a mistake)

ka3de: Right :expressionless: see [137e6c5](https://github.com/grafana/synthetic-monitoring-agent/pull/653/commits/137e6c5e68ccabc63ace4af3722e4704c998565e).

mem (approved):

internal/telemetry/region_pusher.go
roobre: Damn, math is hard.

PR #665, opened 2024-03-25
Allow returning stale data from tenants cache

ka3de:
In the case where an entry in the cache was expired, if the data could not be refreshed by requesting it to the API, an error was returned. This implied a blocking behavior for various actions, such as:
- [Check executions](https://www.google.com/url?q=https://github.com/grafana/synthetic-monitoring-agent/blob/8599f14b91658c3766ed0243bf23a77d420c073b/internal/limits/tenant.go%23L41&sa=D&source=docs&ust=1711445883630651&usg=AOvVaw0fdvZjbtsvxcfooEP4ID3Z): If limits could not be verified.
- [Publish events](https://www.google.com/url?q=https://github.com/grafana/synthetic-monitoring-agent/blob/8599f14b91658c3766ed0243bf23a77d420c073b/internal/pusher/v2/tenant_pusher.go%23L96&sa=D&source=docs&ust=1711445883630608&usg=AOvVaw1H3i9mDvDrO9Tlka3tzJSh): If tenant remotes could not be obtained.

These actions should continue to run even if the agents lose connection with the API, or the API is behaving abnormally. Therefore, this change favours returning stale tenants data from the cache in case it can not be refreshed by requesting it to the API.

The worst case scenario for each mentioned case after this change is:
- Check execution: The check will be executed with limits that might have changed.
- Publish: The metrics and logs will be pushed using remotes that might have changed.

Both cases can lead to metrics and logs backends to reject our requests. But these changes happen infrequently, so most often the data returned by the cache will still be valid and checks will be able to run even if the agent is disconnected from the API. And even in the case on which these errors happen, is better than blocking preventively.

### DIFF
--- a/internal/tenants/manager_test.go
+++ b/internal/tenants/manager_test.go
@@ -15,12 +15,17 @@ import (
 type testTenantsClient struct {
 	tenants      map[int64]sm.Tenant
 	requestCount map[int64]int
+	err          error
 }
 
 var errTenantNotFound = errors.New("tenant not found")
 
 func (c *testTenantsClient) GetTenant(ctx context.Context, in *sm.TenantInfo, opts ...grpc.CallOption) (*sm.Tenant, error) {
 	c.requestCount[in.Id]++
+
+	if c.err != nil {
+		return nil, c.err
+	}
 
 	tenant, found := c.tenants[in.Id]
 	if !found {
@@ -60,8 +65,8 @@ func TestTenantManagerGetTenant(t *testing.T) {
 	defer cancel()
 
 	tenantCh := make(chan sm.Tenant)
-
-	tm := NewManager(ctx, &tc, tenantCh, 500*time.Millisecond)
+	cacheExpirationTime := 200 * time.Millisecond
+	tm := NewManager(ctx, &tc, tenantCh, cacheExpirationTime)
 
 	t1 := tc.tenants[1]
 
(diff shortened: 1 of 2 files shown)


internal/tenants/manager_test.go
```diff
@@ -142,4 +147,17 @@ func TestTenantManagerGetTenant(t *testing.T) {
 	require.NotNil(t, tenant)
 	require.Equal(t, 0, tc.requestCount[t3.Id])
 	require.Equal(t, t3, *tenant)
+
+	// wait for tenants to expire
+	time.Sleep(cacheExpirationTime)
```
The-9880, line 152: Just a side thought: we could maybe use something like a `TestClock` to test cache expirations in general - instead of calling `time.Sleep` we would do something like `testClock.Advance(200 ms)` and it would validate that these expirations work without having to pause execution.

Then you don't have to consider reducing the timeout from `500ms` -> `200ms` in tests as well 😄

The-9880 (approved):
Looks good to me!

internal/tenants/manager_test.go
ka3de: You are absolutely right that sleeping in a test is not ideal, but mocking time is not so straightforward, as we would have to define an interface for it, have two different implementations, one for std and one for our "test time", and have an instance for it in the cache, so it gets kind of cumbersome. For other cases where it's required, such as when using a ticker, I usually do [this](https://github.com/grafana/synthetic-monitoring-agent/blob/main/internal/telemetry/ticker.go), but for this case sleeping some time seems reasonable :confused:

PR #674, opened 2024-04-08
proto fields for scripted check enablement

The-9880:
First in line for [restricting scripted/browser checks to enabled probes](https://github.com/grafana/synthetic-monitoring/issues/41).

We add a `Capabilities` structure which contains flags for different 'features' we might enable on probes. The Agent can have a method which takes this structure in and validates its setup against this config.

The first field in here is `DisableScriptedChecks` which is set if the probe is disallowed from receiving scripted checks to execute.


mem:
Looking at the other PR (https://github.com/grafana/synthetic-monitoring-agent/pull/675), I'm thinking that whether or not a probe is allowed to run scripts is something that should be set the API level (because the probe is created and available before it connects to the API). The API should tell the probe "you are supposed to run scripts" and the probe can validate that the feature flag is set (or that it's able to run k6, because the binary is available, or it's been provided with a URL for the runner).

So I think the proto message should be extended to include that info, and the API should be extended so that a probe is eligible for running k6 checks or not.

The-9880:
> Looking at the other PR (#675), I'm thinking that whether or not a probe is allowed to run scripts is something that should be set the API level (because the probe is created and available before it connects to the API). The API should tell the probe "you are supposed to run scripts" and the probe can validate that the feature flag is set (or that it's able to run k6, because the binary is available, or it's been provided with a URL for the runner).
> 
> So I think the proto message should be extended to include that info, and the API should be extended so that a probe is eligible for running k6 checks or not.

I think that having the API determine if the probe should run the check or not will add some UI overhead and also some friction in toggling this for a probe.

Here are a few of my thoughts:
* Having the probe opt-in via its config (`ProbeInfo`) upon registration is more flexible - otherwise, I would have to create a 'new probe' on the API and change my token to toggle scripted checks execution.
* It's easy to toggle on/off for our public clusters by updating the config (see [this PR](https://github.com/grafana/sm-proxy/pull/88)).
* We already [update probe info](https://github.com/grafana/synthetic-monitoring-api/blob/main/internal/grpc/checks.go#L220) in the DB when probes (re)connect to the API, so there's not significant new overhead introduced by making this field reflect the probe's runtime config.

pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -133,6 +134,7 @@ message Probe {
   bool   deprecated        = 14 [(gogoproto.jsontag) = "deprecated"];                           // whether the probe has been deprecated
   double created           = 100 [(gogoproto.jsontag) = "created"];                             // time the probe was created
   double modified          = 101 [(gogoproto.jsontag) = "modified"];                            // last time modified
+  bool k6Enabled           = 102 [(gogoproto.jsontag) = "k6Enabled"];                           // probe has scripted checks enabled
```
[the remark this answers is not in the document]
The-9880: `ScriptedCheckEnabled` is sent by the Probe to the API to let the API know that this probe can run scripted checks.

Otherwise, the API should throw an error if the user assigns scripted checks to this probe.

Adding `k6Enabled` to the `Probe` object allows:
1. The FE to see which probes are eligible for scripted checks - in the future, we can maybe filter out ineligible probes when creating scripted checks.
2. Fetching this field from the DB - we call `toSM` on `db.probe`. This is used in [this PR](https://github.com/grafana/synthetic-monitoring-api/pull/815) so we can throw the API error properly.

pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -117,22 +117,27 @@ message RegisterProbeResult {
...
+  string            commit                = 12 [(gogoproto.jsontag) = "commit"];                               // commit hash used to build the probe
+  string            buildstamp            = 13 [(gogoproto.jsontag) = "buildstamp"];                           // timestamp when the probe was built
+  bool              deprecated            = 14 [(gogoproto.jsontag) = "deprecated"];                           // whether the probe has been deprecated
+  double            created               = 100 [(gogoproto.jsontag) = "created"];                             // time the probe was created
+  double            modified              = 101 [(gogoproto.jsontag) = "modified"];                            // last time modified
+  ProbeCapabilities capabilities          = 102 [(gogoproto.jsontag) = "capabilities"];                        // probe feature enablements
+}
+
+message ProbeCapabilities {
+  bool disableScriptedChecks = 1 [(gogoproto.jsontag) = "disableScriptedChecks"]; // probe should not run scripted checks
```
The-9880, line 140: wonder if we should nest this under `Probe`

mem: yeah, that makes sense, thanks!

mem (approved):

pkg/pb/synthetic_monitoring/checks.proto
The-9880: I realize this comment might've been vague - I mean nest this whole proto under `Probe`.

i.e. `synthetic_monitoring.Probe.Capabilities`

The-9880: Nested! Going to merge.

PR #683, opened 2024-04-15
skip interpolating multihttp variables into the metric names

rdubrock:
Having the actual URL here can lead to an active series explosion. If a variable is passed as part of the URL that changes on each run, it means new active series will get generated every time the check executes. To stop this, we should use the un-interpolated value as the metric name.

In pseudo code example: 

```
var random := randomNumber()

url := https://www/example.com/{{ random }}

makeRequest(url) // Here the request will actually go to https://www.example.com/3252
reportMetrics("https://www/example.com/{{ random }}") // metric values will be reported as { url="https://www.example.com/{{random}}" }
```


mem (approved):
I think this makes sense.

PR #684, opened 2024-04-17
Terminate agent if capabilities not supported

The-9880:
Part of the new capabilities framework, probes must support features that are expected of them upon connection or else will terminate with an error verbose enough to diagnose the issue.

Un-drafting this, but should merge after the [API changes](https://github.com/grafana/synthetic-monitoring-api/pull/815).

### DIFF
--- a/internal/checks/checks.go
+++ b/internal/checks/checks.go
@@ -47,7 +47,7 @@ const (
 	errTransportClosing    = TransientError("transport closing")
 	errProbeUnregistered   = TransientError("probe no longer registered")
 	errIncompatibleApi     = Error("API does not support required features")
-	errCapabilityK6Missing = Error("K6 is required for scripted check support")
+	errCapabilityK6Missing = Error("k6 is required for scripted check support")
 )
 
 // Backoffer defines an interface to provide backoff durations.
@@ -297,6 +297,14 @@ func (c *Updater) Run(ctx context.Context) error {
 				Err(err).
 				Str("connection_state", c.api.conn.GetState().String()).
 				Msg("cannot connect, bailing out")
+			return err
+
+		case errors.Is(err, errCapabilityK6Missing):
+			// probe missing k6 to support capability
+			c.logger.Error().
+				Err(err).
+				Str("connection_state", c.api.conn.GetState().String()).
+				Msg("k6 requirement missing, bailing out")
 			return err
 
 		case errors.Is(err, context.Canceled):
--- a/internal/checks/checks.go
+++ b/internal/checks/checks.go
@@ -43,10 +43,11 @@ type TransientError Error
 func (e TransientError) Error() string { return string(e) }
 
 const (
-	errNotAuthorized     = Error("probe not authorized")
-	errTransportClosing  = TransientError("transport closing")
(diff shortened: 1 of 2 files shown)


internal/checks/checks.go
```diff
@@ -43,10 +43,11 @@ type TransientError Error
...
 const (
-	errNotAuthorized     = Error("probe not authorized")
-	errTransportClosing  = TransientError("transport closing")
-	errProbeUnregistered = TransientError("probe no longer registered")
-	errIncompatibleApi   = Error("API does not support required features")
+	errNotAuthorized       = Error("probe not authorized")
+	errTransportClosing    = TransientError("transport closing")
+	errProbeUnregistered   = TransientError("probe no longer registered")
+	errIncompatibleApi     = Error("API does not support required features")
+	errCapabilityK6Missing = Error("K6 is required for scripted check support")
```
mem, line 50: it should be `k6`, lowercase

internal/checks/checks.go
```diff
@@ -465,6 +470,14 @@ func (c *Updater) loop(ctx context.Context) (bool, error) {
 	return connected, errorHandler(err, "getting changes from synthetic-monitoring-api", signalFired)
 }
 
+func (c *Updater) validateProbeCapabilities(capabilities *sm.Probe_Capabilities) error {
+	if !capabilities.DisableScriptedChecks && c.k6Runner == nil {
+		return errCapabilityK6Missing
```
mem, line 475: is this really causing the agent to exit?

I cannot see why. This is an error, that's true. That should cause the Run function to land in the default case of the switch, on line 309, is that right? If so, isn't it sleeping trying to back off?

mem (commented):
I'm probably missing something...

internal/checks/checks.go
The-9880: Good catch, added a case for the new error so we avoid backing off pointlessly

mem: I was trying to suggest some changes, but I was having hard time writing it down. I pushed a new commit (511b371be364e9687331808dda9bdcb4f71ab9b4). Could you please review that?

internal/checks/checks.go
```diff
@@ -38,15 +38,26 @@ type Error string
...
 func (e TransientError) Error() string { return string(e) }
 
+var _ error = TransientError("")
+
+// FatalError is an error that causes the program to terminate.
+type FatalError Error
+
+func (e FatalError) Error() string { return string(e) }
+
+var _ error = FatalError("")
```
The-9880, line 53: @mem I like the changes! I'm just unsure why `var _ error` is being declared for both `FatalError` and `TransientError`?

mem: Blank variable declarations don't actually declare variables. They are compile-time assertions.

```
var _ interfaceType = whatever...
```

means that the compiler must validate that the result of the right-hand-side satisfies the interface `interfaceType`.

In this case, that means that `FatalError` satisfies the `error` interface (implements all its methods).

mem (approved):

internal/checks/checks.go
The-9880: Nice, thank you!

PR #688, opened 2024-04-18
Update capabilities proto

The-9880:
`Disable` field might be unnatural for clients to see/use. We can handle the double-negative of the field in the backend and just send clients a more natural `enable` boolean flag.

This follows from this discussion: https://github.com/grafana/synthetic-monitoring-api/pull/815/files#r1569787810


pkg/pb/synthetic_monitoring/checks.proto
```diff
@@ -123,7 +123,8 @@ message RegisterProbeResult {
 // Probe represents a probe.
 message Probe {
   message Capabilities {
-    bool disableScriptedChecks = 1 [(gogoproto.jsontag) = "disableScriptedChecks"]; // probe should not run scripted checks
+    bool disableScriptedChecks = 1 [(gogoproto.jsontag) = "disableScriptedChecks", deprecated = true]; // probe should not run scripted checks
```
The-9880, line 126: I considered renaming the existing field since it isn't in use yet, but `buf` has a `NO FIELD DELETE` validation which gets in the way... Maybe for the best

PR #689, opened 2024-04-18
Log configuration at start up

mem:
When the program is starting up, log the configuration parameters. This helps with debugging and support.

Be careful not to log the API secret.

This is a no-op change, but it's a little noisy becuase of the introduction of the config struct.

### DIFF
--- a/cmd/synthetic-monitoring-agent/secret.go
+++ b/cmd/synthetic-monitoring-agent/secret.go
@@ -1,0 +1,25 @@
+package main
+
+// Secret represents a string that shouldn't be logged.
+type Secret string
+
+// String returns a redacted version of the secret.
+func (s Secret) String() string {
+	return "<redacted>"
+}
+
+// Set sets the value of the secret.
+func (s *Secret) Set(value string) error {
+	*s = Secret(value)
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


cmd/synthetic-monitoring-agent/secret.go
```diff
@@ -0,0 +1,25 @@
...
+// String returns a redacted version of the secret.
+func (s Secret) String() string {
+	return "<redacted>"
+}
+
+// Set sets the value of the secret.
+func (s *Secret) Set(value string) error {
+	*s = Secret(value)
+	return nil
+}
```
roobre, line 15: Does `Set(string) error` make this implement any interface?

roobre (approved):
Having a `string` implement `Stringer` is a pretty clever way of redacting!

cmd/synthetic-monitoring-agent/secret.go
mem: ah, yeah, it's what flag.Value wants. It _also_ wants String() for that matter.

mem: It's needed for line 83 to work.

PR #691, opened 2024-04-19
Feature: automatically set up GOMEMLIMIT

mem:
Automatically set up GOMEMLIMIT to 90% of the cgroup allowed memory. If
running in an environment where there's no cgroup limit set up, fallback
to 90% of the system's memory. Note that this doesn't mean that the
program will use 90% of the system's memory, but that if it does get to
that point, the garbaga collector will try to be more aggresive.

It is strongly recommended to set a limit via cgroups.


mem:
This needs #689

The-9880 (approved):

PR #694, opened 2024-04-29
k6runner: handle errors reported by http runners

roobre:
Previously, k6 runners used to communicate errors by returning a non-200 status code, and not returning the logs of the execution. mq proxy, in addition to returning non-200 status codes, also includes an error code and description for errors that occur while running the script.

With the introduction of https://github.com/grafana/sm-k6-runner/pull/153, we now report as errors exceptions being thrown on the script, such as trying to access undefined variables, or invocations of `fail()`. In this scenario, users likely expect to see the check reported as a failure, but also to see the logs and metrics produced by it.

This PR changes the error-handling logic, so that the response is always parsed for status code that (should) include it, and the success of the check is calculated by checking if there response contains an error.

### DIFF
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -75,6 +75,11 @@ func NewScript(script []byte, k6runner Runner) (*Script, error) {
 	return &r, nil
 }
 
+var (
+	ErrBuggyRunner = errors.New("runner returned buggy response")
+	ErrFromRunner  = errors.New("runner reported an error")
+)
+
 func (r Script) Run(ctx context.Context, registry *prometheus.Registry, logger logger.Logger, internalLogger zerolog.Logger) (bool, error) {
 	k6runner := r.runner.WithLogger(&internalLogger)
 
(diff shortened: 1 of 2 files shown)


internal/k6runner/k6runner.go
```diff
@@ -111,7 +112,8 @@ func (r Script) Run(ctx context.Context, registry *prometheus.Registry, logger l
 		return false, err
 	}
 
-	return !resultCollector.failure, nil
+	success := result.Error == "" && result.ErrorCode == ""
+	return success, nil
```
roobre, line 116: Not sure if at this point we should build a new error with `result.Error` 🤔

The-9880: It makes sense to do I think

roobre: Delved a bit more on what this error is used for, and it seems like it is just `log.Warn`ed?

https://github.com/grafana/synthetic-monitoring-agent/blob/e3403fb23b923802e4c0552c6a3da180ece97ab5/internal/prober/scripted/scripted.go#L57-L65

If this is the case we might want to return an error when the code is something that is not-on-us.

The-9880 (approved):

internal/k6runner/k6runner.go
roobre: Refined (slang for "made more complicated") the error handling logic in the latest commit, LMK what you think :)

PR #699, opened 2024-04-30
Bump Go to 1.22

mem:
This has no effect on the released images because those are already using Go 1.21, but it's nice to keep stuff in sync.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/grafana/synthetic-monitoring-agent
 
-go 1.21
+go 1.22
 
 require (
 	github.com/go-kit/kit v0.13.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/grafana/synthetic-monitoring-agent
 
-go 1.21
+go 1.22
```
The-9880, line 3: Should we put down `1.22.0` instead?  I have `1.22.2` locally and I find without the full release number, it keeps pestering me about there not being a toolchain for `1.22`

The-9880 (approved):
LGTM, just had a non-blocking question

go.mod
mem: really?

let me look at what the logs say...

mem: the logs for the build do not say anything because I force GOTOOLCHAIN to local.

mem: I'll look at this later.

PR #711, opened 2024-05-24
checks/test: make timer big enough for context cancel to be picked up

roobre:
I cannot reproduce the flaky test locally, so let's try things and see what sticks.

---

I think this was it: The 100 nanosecond timer was probably long enough for context cancellation to be picked up before the timer expired in our beefy machines, but not in the handicapped CI runners.


The-9880 (approved):
👀

Issue #637, opened 2024-02-28
Variables cannot be used in the body of MultiHTTP requests


peterschretlen:
The [docs](https://grafana.com/docs/grafana-cloud/monitor-public-endpoints/checks/multihttp/#variables) state:
> It’s possible to use this anywhere where a string is expected, for example as the value of a header or the value of an assertion

So it is reasonable to expect that you'd be able to use the variable value in the body of a request.  Say for example you have a variable `$token` with value `abc` from the previous request, you should be able to setup a POST request like this:
```
{
      "acceptance_token": "${token}"
}
```

And expect the body to be:
```
{
      "acceptance_token": "abc"
}
```

It appears we just [copy the body as-is](https://github.com/grafana/synthetic-monitoring-agent/blob/0a03a914d4e602e3a2693726a603f112767d472f/internal/prober/multihttp/script.go#L89-L106)

pascal-arevalo:
Are there any updates on this? 😄

PR #713, opened 2024-05-29
Fix: Interpolate variables into MultiHTTP request bodies

The-9880:
Addresses https://github.com/grafana/synthetic-monitoring-agent/issues/637.
Writing `encoding.b64decode(<BODY>)` into the template doesn't let us catch where the user is referencing variables.

We should either:
1. Write the whole string expression (`"literal " +vars['var1']+ " more text " +vars['var2']`) directly into the template, or
2. Add a functional block in the template to take the b64-decoded `$body` and replace any interpolated variables in the template code itself.

For option 1, `buildBody` could likely just return the result of `performVariableExpansion` on the body payload.

Option 2 is probably better practice as the user input won't escape the string; however users who are trying to execute arbitrary code _could_ just leverage scripted checks.

### DIFF
--- a/internal/prober/multihttp/script_test.go
+++ b/internal/prober/multihttp/script_test.go
@@ -320,17 +320,9 @@ func TestBuildBody(t *testing.T) {
 		input    input
 		expected string
 	}{
-		"variable in body is interpolated correctly": {
-			input:    input{body: &sm.HttpRequestBody{Payload: []byte("test ${testVar} works")}},
-			expected: "'test '+vars['testVar']+' works'", // 'test '+vars['testVar']+' works'
-		},
-		"body starting with variable is interpolated correctly": {
-			input:    input{body: &sm.HttpRequestBody{Payload: []byte("${testVar} test")}},
-			expected: "vars['testVar']+' test'", // vars['testVar']+' test'
-		},
 		"not empty": {
 			input:    input{body: &sm.HttpRequestBody{Payload: []byte("test")}},
-			expected: "'test'", // 'test'
+			expected: `encoding.b64decode("dGVzdA", 'rawstd', "b")`,
 		},
 		"nil": {
 			input:    input{body: nil},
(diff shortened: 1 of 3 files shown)


internal/prober/multihttp/script_test.go
```diff
@@ -320,9 +320,17 @@ func TestBuildBody(t *testing.T) {
 		input    input
 		expected string
 	}{
+		"something else": {
```
The-9880, line 323: oops, fix the test case name 🙈

internal/prober/multihttp/script_test.go
```diff
@@ -342,6 +342,42 @@ func TestBuildBody(t *testing.T) {
 	}
 }
 
+func TestInterpolateBodyVariables(t *testing.T) {
```
roobre, line 345: We can get tests to run a bit faster if we t.Parallel() all the things!

```suggestion
func TestInterpolateBodyVariables(t *testing.T) {
	t.Parallel()

```

internal/prober/multihttp/script_test.go
```diff
@@ -342,6 +342,42 @@ func TestBuildBody(t *testing.T) {
...
+			input: input{body: &sm.HttpRequestBody{Payload: []byte("${variable1} is ${variable1} fun ${variable2} ok ${variable3}")}},
+			expected: []string{
+				"body.replace('${variable1}', vars['variable1'])",
+				"body.replace('${variable2}', vars['variable2'])",
+				"body.replace('${variable3}', vars['variable3'])",
+			},
+		},
+	}
+	for name, tc := range testcases {
+		t.Run(name, func(t *testing.T) {
```
roobre, line 374: Same as above
```suggestion
		t.Run(name, func(t *testing.T) {
			t.Parallel()

```

roobre (approved):
LGTM! Just a small nit about t parallelism

internal/prober/multihttp/script_test.go
The-9880: Ah true, thank you!

roobre (approved):

PR #722, opened 2024-06-06
Enable K6 by default in agent deployments

The-9880:
Users upgrading from older probes (will) have k6 scripts disabled by default in the API if they weren't already running k6 checks. There are 2 cases for them:
1. Had `k6` FF in args: no impact, this setup makes it default-enabled.
2. Did not have `k6` FF: this was the default. If they use the Docker image or have k6 on their PATH, then their upgraded agent would be able to set up a `k6Runner`. Any previously-assigned MultiHTTP or scripted checks may start running. They will need to enable in the UI to assign k6 checks.

For new probes, this change aligns the agent's default with the default assumption of the API, so it should reduce issues with setup.

Also updated the error message when k6 is missing - feedback appreciated.


internal/checks/checks.go
```diff
@@ -53,7 +53,7 @@ func (e FatalError) Error() string { return string(e) }
 var _ error = FatalError("")
 
 const (
-	errCapabilityK6Missing = FatalError("k6 is required for scripted check support")
+	errCapabilityK6Missing = FatalError("k6 is required for scripted check support - configure k6 or edit probe capabilities in the SM app")
```
The-9880, line 56:
```suggestion
	errCapabilityK6Missing = FatalError("k6 is required for scripted check support - configure k6 or disable the probe's k6 capability in the SM app")
```

Thoughts on this rewording? I can't find a good succinct way to describe this.

roobre: What comes to my mind is
```suggestion
	errCapabilityK6Missing = FatalError("synthetic monitoring backend requested k6 to be enabled on this probe, but it is disabled by CLI arg")
```

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -133,6 +133,12 @@ func run(args []string, stdout io.Writer) error {
 		}
 	}
 
+	if !config.DisableK6 {
+		if err := features.Set(feature.K6); err != nil {
+			return fmt.Errorf("cannot set k6 feature: %w", err)
+		}
+	}
```
roobre, line 140: It's not fully clear to me how the new flag (`--disable-k6`) and the old one (`--features=k6`) interact.

I think it makes sense having a dedicated K6 flag for disabling k6, but that makes me wonder if we should get rid of, or deprecate the feature flag.

The-9880: This change does get rid of the `--features` flag on L105-106. I could "formally" deprecate it by detecting it and printing out a message.

---

**How the new vs old flags interact**

The old flag `--features` is captured by `flags.Var`, which passes the cmd argument to the `Set` method of the container object. `features.Set` assumes a csv string and splits it to save each element.

Example: `features.Set("traceroute,k6,adhoc")` -> `map[string]struct{}{"traceroute": struct{}, "k6": struct{}, "adhoc": struct{}}`.

With this change, if `--disable-k6` is `false` or unset, we explicitly call `features.Set("k6")` to mark k6 as enabled by default.

roobre: Ah, you're definitely right, I did not see the removal of the `--features` CLI flag 🙈 
Looking good :)

roobre (approved):

PR #723, opened 2024-06-07
examples/kubernetes: update k8s example

roobre:
Supersedes #570 

This PR intends to solve the problem of our k8s example not working out of the box.

It takes a different approach compared to https://github.com/grafana/synthetic-monitoring-agent/pull/570, where instead of making the agent listen on all interfaces by default, it comments the default liveness/readiness probes, which achieves the same goal.

Additionally, it makes the following minor changes:
- Adds some extra comments to the secrets, with references to the docs
- Reorders some fields to make them more visually pleasant
- Moves the backend URL from the deployment to the secret (so user hopefully only needs to edit one file instead of two).
- Launches the sm probe directly instead of doing so from a shell, which I believe is more idiomatic


lujameva:
Just tested this using the latest docker image docker.io/grafana/synthetic-monitoring-agent@sha256:4d54daa546087b69df192bb2d1d9c1a525da71af48b04a170f8a7a7398b6a0ce and the pod is still not ready because the readiness probe at localhost:4050/ready always returns 503.

So I commented the readiness probe. The liveness probe is working after this change though. Thanks for that!

examples/kubernetes/deployment.yaml
```diff
@@ -21,32 +21,39 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  key: api-server
+                  name: sm-agent-1
           image: grafana/synthetic-monitoring-agent:latest
           name: agent
           ports:
             - containerPort: 4050
               name: http-metrics
-          readinessProbe:
```
xcheng75, line 42: why we are commenting out readiness probe?

ka3de (approved):

examples/kubernetes/deployment.yaml
ka3de: By default the agent is not listening on all interfaces, which is required for liveness and readiness probes. See the new added comments that explain it in this PR.

PR #726, opened 2024-06-10
Fix: deprecate --features and warn user

The-9880:
Allow the `--features` flag to be present when running the agent, but keep it defunct and warn users that it's deprecated.


cmd/synthetic-monitoring-agent/main.go
```diff
@@ -103,6 +105,7 @@ func run(args []string, stdout io.Writer) error {
 	flags.BoolVar(&config.AutoMemLimit, "enable-auto-memlimit", config.AutoMemLimit, "automatically set GOMEMLIMIT")
 	flags.BoolVar(&config.DisableK6, "disable-k6", config.DisableK6, "disables running k6 checks on this probe")
 	flags.Float64Var(&config.MemLimitRatio, "memlimit-ratio", config.MemLimitRatio, "fraction of available memory to use")
+	flags.StringVar(&config.FeatureString, "features", featuresPlaceholder, "deprecated: optional feature flags")
```
The-9880, line 108: Parsing it as a string now, makes setting a default value to detect a little easier.

mem (dismissed):
Please restore the `--features` flag as it existed previously, just make it output a warning if traceroute or k6 are specified.

mem (approved):
Thank you!

PR #727, opened 2024-06-11
Create `k6runner.Script` and make probers aware of it

roobre:
Before this PR, `MultiHTTP` and `Scripted` `Prober`s created k6 runners using only the raw script data (`[]byte`) as input to the runner. This have worked nicely, as until now runners only required two things to work: The raw script data, and a timeout that was encoded in the context for `Runner.Run()`.

We have seen however a couple scenarios that make me think this is not enough for the near future:
- We want to decouple the context from `Runner.Run()`, a.k.a. the request timeout, from the check timeout that is communicated to the runners. This is needed to properly implement what #715 workarounded.
- We want to make k6 runners aware of some check metadata, for troubleshooting. https://github.com/grafana/sm-k6-runner/pull/191 worked around this using a hash, but it's not ideal.
- We want to be able to implement different queues for browser and non-browser scripts (https://github.com/grafana/sm-k6-runner/issues/174). We tentatively want to do this by extracting script features (https://github.com/grafana/sm-k6-archiver/pull/22), which will eventually need to be sent to the runners.

For these three reasons, I think it is beneficial to create a richer type so that `Prober`s can feed all these details to the k6 runner, which this PR does.

Please note that this PR is not feature complete, namely how timeouts are handled does not fully makes sense. This is because I tried to keep the patchset as small as possible for reviewability. How timeouts are handled is improved in #724.

### DIFF
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -22,9 +22,24 @@ import (
 	"github.com/spf13/afero"
 )
 
+// Script is a k6 script that a runner is able to run, with some added instructions for that runner to act on.
+type Script struct {
+	Script   []byte   `json:"script"`
+	Settings Settings `json:"settings"`
+	// TODO: Add Metadata and Features.
+}
+
+type Settings struct {
+	// Timeout for k6 run, in milliseconds. This value is a configuration value for remote runners, which will instruct
+	// them to return an error if the operation takes longer than this time to complete. Clients should expect that
+	// requests to remote runners may take longer than this value due to network and other latencies, and thus clients
+	// should wait additional time before aborting outgoing requests.
+	Timeout int64 `json:"timeout"`
+}
+
 type Runner interface {
 	WithLogger(logger *zerolog.Logger) Runner
-	Run(ctx context.Context, script []byte) (*RunResponse, error)
+	Run(ctx context.Context, script Script) (*RunResponse, error)
 }
 
 type RunnerOpts struct {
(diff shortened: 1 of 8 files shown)


internal/k6runner/k6runner.go
```diff
@@ -22,9 +22,24 @@ import (
...
+	// them to return an error if the operation takes longer than this time to complete. Clients should expect that
+	// requests to remote runners may take longer than this value due to network and other latencies, and thus clients
+	// must not cancel outgoing requests after this time passes. Outgoing requests should be cancel on RetryOn instead.
+	Timeout int64 `json:"timeout"`
+}
+
 type Runner interface {
 	WithLogger(logger *zerolog.Logger) Runner
-	Run(ctx context.Context, script []byte) (*RunResponse, error)
+	Run(ctx context.Context, script Script) (*RunResponse, error)
```
roobre, line 42: This is the core change of this PR: Replace the `[]byte` argument in `Runner.Run` with the new `Script` type. All other changes are essentially adapting the rest of the codebase to this change.

internal/k6runner/k6runner.go
```diff
@@ -22,9 +22,24 @@ import (
 	"github.com/spf13/afero"
 )
 
+// Script is a k6 script that a runner is able to run, with some added instructions for that runner to act on.
+type Script struct {
+	Script   []byte   `json:"script"`
+	Settings Settings `json:"settings"`
+	// TODO: Add Metadata and Features.
```
roobre, line 29: Here is where, in the future, we will add metadata (e.g. check ID, tenant ID) for the runners to log later, and script features (e.g. browser enabled, k6 version).

mem (approved):

PR #737, opened 2024-06-19
Add the basic stuff to support a browser check type

mem:
- Add the settings to the proto definitions and generate the corresponding code.
- Add the fake data for the scraper test.
- Generate the corresponding accounting data.
- Scaffold the scraper.


mem:
This is part of https://github.com/grafana/synthetic-monitoring/issues/72

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -430,6 +437,9 @@ func (c AdHocCheck) Type() CheckType {
 	case c.Settings.Grpc != nil:
 		return CheckTypeGrpc
 
+	case c.Settings.Browser != nil:
+		return CheckTypeScripted
+
```
roobre, line 442: Should this be 

```suggestion
	case c.Settings.Browser != nil:
		return CheckTypeBrowser

```

instead?

roobre (approved):
We might need a dedicated class to bill browser differently (I'm not sure how that works exactly), but for the time being I think this is the way to go.

PR #740, opened 2024-06-26
bump k6 version using renovate

roobre:
Doing this by hand is tiring. Renovate is configured to not bump anything else, so we keep working with dependabot as we are used to.

This yields PRs like https://github.com/roobre/grafana-synthetic-monitoring-agent/pull/3

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -1,0 +1,19 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:best-practices",
+    ":semanticCommitsDisabled",
+  ],
+  "enabledManagers": ["custom.regex"], // We use renovate for things that dependabot cannot handle only.
+  "customManagers": [
+    {
+      "customType": "regex",
+      "datasourceTemplate": "github-releases",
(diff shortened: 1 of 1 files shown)


mem:
> Not sure if this should be bumped in steplock with
> 
> https://github.com/grafana/synthetic-monitoring-agent/blob/f62a17384c89105979304e867ce672e1e45372b4/xk6/sm/go.mod#L9
> 
> If it should, we definitely haven't been doing it 😅, but it can be done.

That's the output extension.

I guess that should be handled by dependabot... which might get weird.

mem (commented):
Thank you.

Is this going to work? I think renovate needs to be enabled for the repository, and there's [a process](https://github.com/grafana/security-internal/blob/main/how-to-configure-renovate.md?plain=1#L28) for that.

roobre:
@mem I was counting on using the SaaS renovate, I think it is allowed/enabled in the grafana org (or at least I have used it [before](https://github.com/grafana/xk6-disruptor/blob/main/.github/renovate.json5)).

roobre:
Closing in favor of https://github.com/grafana/synthetic-monitoring-agent/pull/745

PR #741, opened 2024-06-26
k6runner/test: ensure logs are sent to loki when runner reports user errors

roobre:
Doesn't actually fix, but related to https://github.com/grafana/synthetic-monitoring-app/issues/835.

When running a script on a remote runner, roughly two kind of errors can occur:
1. An error occurs reaching the remote runner, meaning that the script could not be scheduled. kan example of this can be network errors reaching our infra.
2. The script was executed, and it returned an error. In this case the runner returns the k6 error log as a part of the response body.

This PR adds tests asserting that those errors in the response body (2) are sent to Loki regardless of the reported `errorCode`.

### DIFF
--- a/internal/k6runner/k6runner_test.go
+++ b/internal/k6runner/k6runner_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -177,7 +178,13 @@ func TestScriptHTTPRun(t *testing.T) {
 
 	var (
 		testMetrics = testhelper.MustReadFile(t, "testdata/test.out")
-		testLogs    = testhelper.MustReadFile(t, "testdata/test.log")
+		// testLogs is a file containing a bunch of log lines. All lines but one have level=debug, which are discarded
+		// by the loki submitter implementation.
+		testLogs = testhelper.MustReadFile(t, "testdata/test.log")
+		// nonkdebugLogLine is the only line on testLogs that does not have level=debug, therefore the only one that is
+		// actually submitted. We use it in the test table below as a sentinel to assert whether logs have been
+		// submitted or not.
+		nonDebugLogLine = `time="2023-06-01T13:40:26-06:00" level="test" msg="Non-debug message, for testing!"` + "\n"
 	)
 
 	for _, tc := range []struct {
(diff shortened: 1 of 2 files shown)


internal/k6runner/k6runner_test.go
```diff
@@ -178,6 +179,11 @@ func TestScriptHTTPRun(t *testing.T) {
 	var (
 		testMetrics = testhelper.MustReadFile(t, "testdata/test.out")
 		testLogs    = testhelper.MustReadFile(t, "testdata/test.log")
+		// nonDebugLogLine is a log line with level != debug present in testLogs.
+		// Non-debug lines are the ones that are actually submitted to the logger, thus the ones we expect to find on it
+		// during testing.
```
ka3de, line 184: nit: Maybe is because I lack some context with the runner, but I find this explanation a little bit confusing. It mentions `debug` level, but it is not clear from here the reasoning for that, unless you go search the implementation and see that we are explicitly discarding any log line with `debug` level and no `source` label.

internal/k6runner/k6runner_test.go
```diff
@@ -404,13 +417,33 @@ func TestK6LogsToLogger(t *testing.T) {
...
 	}
-	return nil
+
+	if len(keyvals)%2 != 0 {
+		return errors.New("not the same number of keys and vals")
+	}
+
+	if l.buf == nil {
+		return nil
+	}
```
ka3de, line 437: nit: Since we are effectively supporting two different behaviors for the `testLogger` depending on whether a buffer is set for it or not, should we add a comment for that as the `testLogger` description? E.g.: If no buffer is set it only verifies inputs, if a buffer is set it writes to the buffer.

ka3de (approved):
The modifications to the test look good! Just made a couple of small nit observations on comments in the test implementation.

internal/k6runner/k6runner_test.go
roobre: Thinking about it I might as well make two different implementation for this 🤔

internal/k6runner/k6runner_test.go
roobre: You're most likely right, it was not very obvious why I was looking specifically for this one. I've reworded this a bit, LMK what you think!

internal/k6runner/k6runner_test.go
roobre: Done in the last fixup!

ka3de (approved):

PR #743, opened 2024-06-26
Correct the listen address to scrape metrics to port 4050

The-9880:
@alexwulf noticed agent installations are using port `4031` to serve `/metrics`.

Our default in the docs ([A](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/set-up/set-up-private-probes/#production-deployments), [B](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/set-up/set-up-private-probes/#troubleshooting-private-probes)) and the code ([ref](https://github.com/grafana/synthetic-monitoring-agent/blob/main/cmd/synthetic-monitoring-agent/main.go#L78)) is `4050`. We should align the config file to this.


roobre (approved):
LGTM. I guess this is technically a breaking change, but I suspect not that many people is connecting to the agent.

PR #745, opened 2024-06-27
Bump k6 version using renovate

mem:


.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,26 @@
...
+        id: generate-token
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a #v2.1.0
+        with:
+          revoke: true
+          app_id: ${{ secrets.RENOVATEGRAFANABOT_ID }}
+          private_key: ${{ secrets.RENOVATEGRAFANABOT_PEM }}
+      - name: Self-hosted Renovate
+        uses: renovatebot/github-action@v39.0.5 #v.39.0.5
+        with:
+          renovate-version: 37.5.3@sha256:613fe86ea2730c062a42bf097a02563e35cb0d8df4fef08d2c4e50b8cd2ccdbc
```
roobre, line 24: > If omitted the action will use the latest Docker tag.

Eck.

I can hack a renovate config real quick so it can update itself, on a subsequent PR.

roobre (approved):

.github/workflows/renovate.yaml
mem: I was following the documentation mentioned in the commit. I did wonder why they are pinning renovate to what seems to be an older version.

In my experience, people writing Github actions play loose with the versioning of the Github action.

As for the renovate version itself, I would expect renovate to know better, so I assume there's some incompatibility between 37 and the current 40.

roobre: I'd bet on out of date docs. I've been using renovate daily for two years and they've never broken anything so far. Heck, the thing is so smart that when they deprecate things, it makes a PR migrating the config for you.

PR #750, opened 2024-06-28
Fix .github/renovate.json

mem:
Documentation says I should use these options, renovate says I shouldn't. I guess renovate knows better.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -6,15 +6,10 @@
   ],
 
   "branchPrefix": "grafanarenovatebot/",
-  "username": "GrafanaRenovateBot",
   "gitAuthor": "GrafanaRenovateBot <renovategrafana@grafana.com>",
   "platformCommit": true,
   "dependencyDashboard": false,
-  "platform": "github",
   "forkProcessing": "enabled",
-  "repositories": [
-    "grafana/synthetic-monitoring-agent"
-  ],
 
   "enabledManagers": ["custom.regex"],
   "customManagers": [


mem:
renovate doesn't work without (some) of this options...

ka3de:
There is a conflict due to `forkProcessing` being disabled in #751. I will approve this once that's fixed.

roobre:
> Documentation says I should use these options, renovate says I shouldn't. I guess renovate knows better.
> renovate doesn't work without (some) of this options...

Self-hosted renovate typically has two configuration files, [one for renovate itself](https://docs.renovatebot.com/self-hosted-configuration/), which configures things like authentication, and [another one for each repo](https://docs.renovatebot.com/configuration-options/) that renovate manages.

With this setup of running renovate inside CI, which is not the usual way of running it, we're blurring the line a bit because we're running one renovate instance per repo. `username`, `platform`, and `repositories` are configs for renovate itself and not for the repo, so the bot is probably reading those in repo context and getting a bit confused.

I think the proper way to fix this is to split the configuration in two as renovate expects it. `username`, `platform`, and `repositories` should be on a different file referenced here: https://github.com/grafana/synthetic-monitoring-agent/blob/d798772b1fb616d7aead242325458035f057b53e/.github/workflows/renovate.yaml#L26

And that file should be different from the file in [default paths](https://docs.renovatebot.com/configuration-options/) for the repo config, which contains the rest of the config.

roobre:
I've split the config files as renovate wants them here: https://github.com/grafana/synthetic-monitoring-agent/pull/767/files

mem:
Merged #767 that actually fixes this.

Thanks, @roobre

PR #767, opened 2024-07-04
Split and fix renovate config

roobre:
When selfhosted, renovate requires a separate config file ([ref](https://docs.renovatebot.com/self-hosted-configuration/)). This PR splits out self-hosted options to a different file, and points renovate itself to it. Additionally:
- Migrates repo config to json5 (as it allows comments)
- Simplifies the regex looking for grafana-build-tools container image
- Removes `gitAuthor`, so renovate can rebase PRs created by itself with `platformCommit` (https://github.com/renovatebot/renovate/discussions/29106)

### DIFF
--- a/.github/renovate-selfhosted.json
+++ b/.github/renovate-selfhosted.json
@@ -1,0 +1,7 @@
+{
+  "username": "GrafanaRenovateBot",
+  "platform": "github",
+  "repositories": [
+    "grafana/synthetic-monitoring-agent"
+  ]
+}
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -37,7 +37,8 @@
     {
       // Update grafana-build-tools tags across the repository.
       "customType": "regex",
-      "datasourceTemplate": "github-tags",
+      "depNameTemplate": "ghcr.io/grafana/grafana-build-tools",
+      "datasourceTemplate": "docker",
       "versioningTemplate": "semver",
       "fileMatch": [
         "^\\.github/workflows/.*\\.ya?ml$",
@@ -46,7 +47,7 @@
         ".*\\.mk$"
       ],
       "matchStrings": [
-        "\\bghcr\\.io/(?<depName>grafana/grafana-build-tools):(?<currentValue>\\S+)\\b"
+        "ghcr.io/grafana/grafana-build-tools:(?<currentValue>\\S+)"
       ]
     }
   ]
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -5,10 +5,13 @@
     ":semanticCommitsDisabled"
   ],
 
+
   "branchPrefix": "grafanarenovatebot/",
-  "gitAuthor": "GrafanaRenovateBot <renovategrafana@grafana.com>",
   // Used when renovate runs as a github app.
   // https://docs.renovatebot.com/configuration-options/#platformcommit
+  // Setting platformCommit to `true`, as required by Grafana policy, seems to make renovate think all PRs are modified,
+  // as the dynamic author configured by github does not match the author set in `gitAuthor`. It is recommended to
+  // leave it unset: https://github.com/renovatebot/renovate/discussions/29106.
   "platformCommit": true,
   "dependencyDashboard": false,
   "forkProcessing": "disabled",
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -6,19 +6,15 @@
   ],
 
   "branchPrefix": "grafanarenovatebot/",
-  "username": "GrafanaRenovateBot",
   "gitAuthor": "GrafanaRenovateBot <renovategrafana@grafana.com>",
+  // Used when renovate runs as a github app.
+  // https://docs.renovatebot.com/configuration-options/#platformcommit
   "platformCommit": true,
   "dependencyDashboard": false,
-  "platform": "github",
   "forkProcessing": "disabled",
 
-  "repositories": [
-    "grafana/synthetic-monitoring-agent"
-  ],
-
+  // TODO: We should be able to use the defaults if we turn dependabot off.
   "enabledManagers": ["custom.regex", "gomod"],
-
   "gomod": {
     "enabled": true
   },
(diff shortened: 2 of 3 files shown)


mem:
> When selfhosted, renovate requires a separate config file ([ref](https://docs.renovatebot.com/self-hosted-configuration/)). This PR splits out self-hosted options to a different file, and points renovate itself to it. Additionally:
> 
> * Migrates repo config to json5 (as it allows comments)

I avoided using json5 because pretty much nothing supports it. There are good reasons for that. If you want comments in json, can I suggest that you use jsonnet? Every json document is a valid jsonnet input.

And I was disappointing to find that renovate explicitly declined supporting YAML as the configuration format. Not that I like YAML, but I admit that it's better than JSON at this kind of thing.

> * Removes `gitAuthor`, so renovate can rebase PRs created by itself with `platformCommit` ([Renovate marks nearly all PRs as "edited" with `platformCommit: true` renovatebot/renovate#29106](https://github.com/renovatebot/renovate/discussions/29106))

I copied the configuration from the docs, thinking that was going to provide a simpler path forward. I was so mistaken...

I noticed in the renovate docs that there's some kind of staging / testing functionality. Do you have any experience with that? I would love to avoid the PR / merge / diagnose / PR / ... cycle if possible.

.github/renovate.json5
```diff
@@ -48,7 +50,7 @@
         ".*\\.mk$"
       ],
       "matchStrings": [
-        "\\bghcr\\.io/(?<depName>grafana/grafana-build-tools):(?<currentValue>\\S+)\\b"
+        "ghcr.io/grafana/grafana-build-tools:(?<currentValue>\\S+)"
```
mem, line 53: I do not like to play loose with regular expressions.

What's the problem with correctly anchoring the regular expression? And what's the problem with correctly escaping things that need escaping?

I almost threw up when I realized that renovate does global matches, because that does make writing regular expressions much more complicated, e.g. you cannot reasonably anchor at the beginning or end of a line.

.github/renovate-selfhosted.json
mem, line 1: Can you point me to the documentation that explains what's going on here? I can see that you are pointing the action at this file. How is the other one used then?

My guess: you are saying that this file is the configuration file for the GitHub App, and the other file is the configuration file for renovate (i.e. renovate doesn't care about this one). Is that it?

.github/renovate.json5
```diff
@@ -38,8 +38,10 @@
       ]
     },
     {
+      // Update grafana-build-tools tags across the repository.
       "customType": "regex",
-      "datasourceTemplate": "github-tags",
+      "depNameTemplate": "ghcr.io/grafana/grafana-build-tools",
+      "datasourceTemplate": "docker",
```
mem, line 44: It's OK for me either way, but why is this better?

.github/renovate.json5
mem: Incidentally, the reason why they do global matches is because they encourage you do to what you are removing here: use named captures to provide configuration values, so I don't quite understand why you are removing that instead of just changing the named capture so that it starts before `ghcr`.

mem (commented):
It's OK, I guess.

I would like to understand why for the changes. I can figure out the what myself just by reading thru the changes.

.github/renovate.json5
roobre: > What's the problem with correctly anchoring the regular expression?

This is not a strong opinion.

I've been using renovate and its terrible (and yet awesome) regexes for several years, and I've gone through efforts of making regexes very specific and robust so they don't catch things they shouldn't by mistake. Then I stopped doing it. Since I do the simplest, dumbest, possible regex I've noticed two things:
- Dumb regexes have never caught up things they shouldn't. Not a single time.
- I can actually read the dumb regex three days after I write it.

For those two reasons I keep the shortest, simplest possible one that I cannot anticipate breaking.

> use named captures to provide configuration values

Also not a strong opinion. I've found prematurely generalizing managers also obscures the reasoning for the rule being added, and also adds cognitive complexity (just like the regex).

.github/renovate-selfhosted.json
roobre: I added a link in the PR, but probably not in the most readable way. Sorry about that.

Yep, this is the "self hosted configuration", as renovate refers to it. It is documented here: https://docs.renovatebot.com/self-hosted-configuration/

Note the following paragraph:

> Do not put the self-hosted config options listed on this page in your "repository config" file (renovate.json for example), because Renovate will ignore those config options, and may also create a config error issue.

The repository configuration is described, in turn, here: https://docs.renovatebot.com/configuration-options/. That one is not referenced anywhere because it is one of the (many, arguably too many) default paths renovate looks for after cloning the repo.

roobre: This is a bit more obvious when renovate does not live in the same repo it manages (e.g. when you run the bot itself on k8s), as the self-hosted configuration lives next to the bot, and not in the repository.

roobre:
> I avoided using json5 because pretty much nothing supports it. There are good reasons for that. If you want comments in json, can I suggest that you use jsonnet? Every json document is a valid jsonnet input.

That's fair enough, I've only ever used json5 for renovate only, but I'm curious about what _else_ do you want to support it 🤔

We can go go with plain old json but comments are really, really useful for these kind of rules :(

.github/renovate.json5
roobre: I don't think it is better, it should work the same. It is perhaps a bit more obvious. When I first saw this I wondered: Is there a reason to not do the obvious thing and look at the OCI registry for OCI image tags?

.github/renovate-selfhosted.json
mem: Argh...

Thanks for the explanation.

The other day when I was working on this I started to hate renovate because the documentation is really bad (having a lot of documentation is not the same thing as having good documentation). Now I hate it even more. I did see that line you refer to, I did not (obviously) pick up on the subtlety it's trying to express.

I guess I have to write a PR for the repo where I got the example from...

.github/renovate.json5
mem: The reason is that I was trying to test the configuration before actually committing anything to the repo, and with docker it was failing in even less useful ways than with github-tags.

I *do* think docker is the right thing to do here, because in this case the tag might be published a long time _before_ the docker image is available.

roobre: That's interesting, I'll follow up on that to see why that might happen.

PR #768, opened 2024-07-04
ci/renovate: allow renovate to update itself

roobre:


roobre:
Back to draft as it will receive changes as a result of review comments in #784

d0ugal (approved):

PR #782, opened 2024-07-08
Default to `sm-k6` binary instead of `k6`

roobre:
As described in: https://github.com/grafana/support-escalations/issues/11346#issuecomment-2213578649

For non-docker installations, we ship a `sm-k6` binary with the required sm extension that we must use. However, this binary is never looked up, and instead we default to a `k6` binary in the system which may not have the required extension, resulting in obscure failures with exit code 255.

This PR changes the default value to `sm-k6`, and adjusts the Dockerfile so the `k6` binary is copied inside with this name.

https://github.com/grafana/synthetic-monitoring-agent/compare/look-sm-k6-first is an alternative to this, but I'm leaning towards this impl for simplicity.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ARG TARGETARCH
 ARG HOST_DIST=$TARGETOS-$TARGETARCH
 
 COPY dist/${HOST_DIST}/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
-COPY dist/${HOST_DIST}/k6 /usr/local/bin/k6
+COPY dist/${HOST_DIST}/k6 /usr/local/bin/sm-k6
 COPY scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ ARG TARGETARCH
 ARG HOST_DIST=$TARGETOS-$TARGETARCH
 
 COPY dist/${HOST_DIST}/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
-COPY dist/${HOST_DIST}/k6 /usr/local/bin/k6
+COPY dist/${HOST_DIST}/k6 /usr/local/bin/sm-k6
```
mem, line 16: We are breaking something with this. I just need to figure out what.

We are not breaking the container, because if we change the default at the same time, I cannot imagine why someone using a container would be providing that value explicitly. Someone using the container in k8s on the other hand, might be passing that value explicitly.

We are not passing a k6 value in the default .service file, so this is why package users are getting errors. We install the required binary as "sm-k6" but the agent expects this to be called "k6". So this change is actually fixing something.

My expectation is that some Dockerfile is going to break, but that's going to manifest as a failure to build the docker image.

mem (approved):

Dockerfile
roobre: Ah, this is what this will break:
https://github.com/grafana/sm-k6-runner/blob/8c9a0e723cf3c2d3fda16c3d697b0454f9b203d9/Dockerfile#L24

We'll fix it when we get there tho.

PR #784, opened 2024-07-08
renovate: add workflow to validate config

nadiamoe:
Since we don't use neither the dependency dashboard, nor the SaaS renovate, it may become hard to know if we introduce modifications to the renovate config that do not make sense.

This PR adds a simple check that runs the `renovate-config-validator` tool to ensure the config is valid.

### DIFF
--- a/.github/workflows/renovate-validate.yaml
+++ b/.github/workflows/renovate-validate.yaml
@@ -1,6 +1,8 @@
 name: Validate renovate config
 on:
-  pull_request: {}
+  pull_request:
+    paths:
+      - '**renovate*.*'
 jobs:
   validate:
     runs-on: ubuntu-latest


.github/workflows/renovate-validate.yaml
```diff
@@ -0,0 +1,12 @@
+name: Validate renovate config
+on:
+  pull_request: {}
```
mem, line 3: we should filter this by path, so that it doesn't run on every PR

.github/workflows/renovate-validate.yaml
```diff
@@ -0,0 +1,12 @@
...
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
+      - run: |-
+          docker run -i --rm -v $(realpath .):/repo -w /repo --entrypoint renovate-config-validator ghcr.io/renovatebot/renovate --strict
```
mem, line 12: oh, run, we are validating an old renovate configuration file with a new renovate version.

Doesn't this work with the github action?

roobre: It doesn't seem possible sadly :(

The `docker-cmd-file` is automatically mounted and `basename`d:
https://github.com/renovatebot/github-action/blob/044641a5b884facaf733f679cc57474321fa039d/src/renovate.ts#L42-L49

I can see your concern of these getting out of sync though. I can add a regex to #768 so renovate updates both things (action and this image) it that looks good to you.

.github/workflows/renovate-validate.yaml
mem: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

.github/workflows/renovate-validate.yaml
mem: sure!

.github/workflows/renovate-validate.yaml
roobre: TIL!

.github/workflows/renovate-validate.yaml
roobre: @mem Any other thoughts? I can make the change we were discussing here in #768 if that's fine by you 😊

mem (approved):

PR #787, opened 2024-07-09
Add browser limits and class

mem:
This is adding the "BROWSER" check class and the corresponding "MaxBrowserChecks" limit so that we can make the constants and fields available to the API.

*Not* using the new class in intetional. A follow up change can actually wire this up in the appropriate locations, once we have the support on the API side.


The-9880 (approved):

ka3de:
@mem I think related to this PR, there is a TODO comment made on a previous PR that we should address. [This one](https://github.com/grafana/synthetic-monitoring-agent/blob/28d3edbbc0a44ca2315d08a9dd696def1c651f33/pkg/pb/synthetic_monitoring/checks_extra.go#L270). Browser should be a separate `CheckClass`, this also has an impact on the telemetry, which needs this to count browser check executions separately from scripted.

PR #788, opened 2024-07-09
Report browser check's class as browser

mem:
For the purpose of having some clarity in the data that is made available to users, report the class of the browser check as "browser", not "scripted". This allows us to segment the data when reporting executions.

### DIFF
--- a/internal/telemetry/region_pusher_test.go
+++ b/internal/telemetry/region_pusher_test.go
@@ -22,127 +22,171 @@ const (
 	regionID = 1
 )
 
-var (
-	ee = []Execution{
-		{
-			LocalTenantID: 1,
-			CheckClass:    sm.CheckClass_PROTOCOL,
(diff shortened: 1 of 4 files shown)


internal/telemetry/region_pusher_test.go
```diff
@@ -22,142 +22,194 @@ const (
...
-		},
-	}
+type testData struct {
+	executions []Execution
+	message    sm.RegionTelemetry
+}
 
-	mm = []sm.RegionTelemetry{
-		// 0: this is the expected message for ee[0] -> ee[3] executions
+func getTestDataset(idx int) testData {
```
mem, line 30: I ended up doing this because at some point it felt like I was fighting a water balloon: I would squeeze here and it would blow up over there.

ka3de (approved):

roobre (approved):

PR #789, opened 2024-07-10
Fix browser check class

ka3de:
Addresses previous TODO comments left in #737 about the correct Check Class for browser checks. After the introduction of a specific browser check class in #787 we can address these comments and make browser checks a class on its own, that is necessary in order to distinguish telemetry from scripted checks.


ka3de:
Included in #788.

PR #801, opened 2024-07-22
Info message for -features suggested deprecated syntax.

perjahn:
The "--feature" syntax is not allowed any longer.


roobre (approved):
Good catch, thank you!

PR #809, opened 2024-07-30
feat: Validate browser capability

ka3de:
k6 functionality is required in the agent by default, and is NOT required only in case both, scripted and browser checks, have been disabled for the probe from the API.

Updates: https://github.com/grafana/synthetic-monitoring/issues/116

### DIFF
--- a/internal/checks/checks_test.go
+++ b/internal/checks/checks_test.go
@@ -269,6 +269,8 @@ func TestHandleCheckOp(t *testing.T) {
 }
 
 func TestCheckHandlerProbeValidation(t *testing.T) {
+	t.Parallel()
+
 	testcases := map[string]struct {
 		opts          UpdaterOptions
 		probe         sm.Probe
@@ -417,6 +419,8 @@ func TestCheckHandlerProbeValidation(t *testing.T) {
 
 	for testName, tc := range testcases {
 		t.Run(testName, func(t *testing.T) {
+			t.Parallel()
+
 			u, err := NewUpdater(tc.opts)
 			require.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


internal/checks/checks_test.go
```diff
@@ -345,17 +415,19 @@ func TestCheckHandlerProbeValidation(t *testing.T) {
 		},
 	}
 
-	for _, tc := range testcases {
-		u, err := NewUpdater(tc.opts)
-		require.NoError(t, err)
+	for testName, tc := range testcases {
+		t.Run(testName, func(t *testing.T) {
```
roobre, line 419: If we're feeling fancy we might as well add `t.Parallel()` here (with the usual copy guard)

```suggestion
	for testName, tc := range testcases {
		tc := tc
			
		t.Run(testName, func(t *testing.T) {
		t.Parallel()
		
```

roobre (approved):
Looking nice, just one small nit

internal/checks/checks_test.go
ka3de: Absolutely, see [bcacc34](https://github.com/grafana/synthetic-monitoring-agent/pull/809/commits/bcacc344a8ab931546875192b7e7bcc8465652f8).
I think we should iteratively modify every test and subest to run in parallel, or at some point dedicate some time for a big PR.

roobre: @ka3de Do mind the `tc := tc` line, otherwise the test will break badly and cryptically as goroutines start to share a variable that changes across iterations.

ka3de: Yeah, I usually forget that they did the trick for us and write the `tc := tc` anyway, but I think it's time to not do it :smile:

PR #829, opened 2024-08-19
Add with-browser Docker image

The-9880:
Issue: https://github.com/grafana/synthetic-monitoring/issues/71
Adds a 3rd stage to the Dockerfile to build the sm-agent image from an Alpine base, installing `chromium-switshader`.

The Drone CI pipeline is also updated to publish new image versions (tagged `*-with-browser`) corresponding to the new build target:
* `latest-with-browser`
* `<semver-long>-with-browser`

We should probably prevent this from publishing until we have some logic in the agent that can leverage it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,14 +22,17 @@ ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
 # third stage with alpine base for better access to chromium
 FROM alpine:3.18 as with-browser
 
+RUN apk --no-cache add tini
 RUN apk --no-cache add chromium-swiftshader
-
-ENV SM_CHROME_BIN=/usr/bin/chromium-browser
-ENV SM_CHROME_PATH=/usr/lib/chromium/
+RUN adduser -D -u 12345 -g 12345 sm
 
 COPY --from=release /usr/local/bin/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
 COPY --from=release /usr/local/bin/sm-k6 /usr/local/bin/sm-k6
 COPY --from=release /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=release /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
-ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
+USER sm
+ENV SM_CHROME_BIN=/usr/bin/chromium-browser
+ENV SM_CHROME_PATH=/usr/lib/chromium/
+
+ENTRYPOINT ["tini", "--", "/usr/local/bin/synthetic-monitoring-agent"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get -y install ca-certificates
 
 ARG TARGETPLATFORM
 
-FROM --platform=$TARGETPLATFORM debian:stable-slim
+FROM --platform=$TARGETPLATFORM debian:stable-slim as release
 ARG TARGETOS
 ARG TARGETARCH
 ARG HOST_DIST=$TARGETOS-$TARGETARCH
@@ -16,5 +16,20 @@ COPY dist/${HOST_DIST}/synthetic-monitoring-agent /usr/local/bin/synthetic-monit
 COPY dist/${HOST_DIST}/k6 /usr/local/bin/sm-k6
 COPY scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
+
+# third stage with alpine base for better access to chromium
+FROM alpine:3.18 as with-browser
+
+RUN apk --no-cache add chromium-swiftshader
+
+ENV SM_CHROME_BIN=/usr/bin/chromium-browser
+ENV SM_CHROME_PATH=/usr/lib/chromium/
+
+COPY --from=release /usr/local/bin/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
+COPY --from=release /usr/local/bin/sm-k6 /usr/local/bin/sm-k6
+COPY --from=release /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
+COPY --from=release /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,15 +24,12 @@ FROM alpine:3.18 as with-browser
 
 RUN apk --no-cache add tini
 RUN apk --no-cache add chromium-swiftshader
-RUN adduser -D -u 12345 -g 12345 sm
 
 COPY --from=release /usr/local/bin/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
 COPY --from=release /usr/local/bin/sm-k6 /usr/local/bin/sm-k6
 COPY --from=release /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=release /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
-USER sm
-ENV SM_CHROME_BIN=/usr/bin/chromium-browser
-ENV SM_CHROME_PATH=/usr/lib/chromium/
+ENV K6_BROWSER_ARGS=no-sandbox,disable-dev-shm-usage
 
 ENTRYPOINT ["tini", "--", "/usr/local/bin/synthetic-monitoring-agent"]
--- a/scripts/configs/drone/main.jsonnet
+++ b/scripts/configs/drone/main.jsonnet
@@ -1,4 +1,4 @@
-local go_tools_image = 'ghcr.io/grafana/grafana-build-tools:v0.23.0';
+local go_tools_image = 'ghcr.io/grafana/grafana-build-tools:v0.22.0';
 
 local step(name, commands, image=go_tools_image) = {
   name: name,
--- a/scripts/configs/drone/main.jsonnet
+++ b/scripts/configs/drone/main.jsonnet
@@ -75,7 +75,7 @@ local vault_secret(name, vault_path, key) = {
   },
 };
 
-local docker_step(tag, os, arch, version='') =
+local docker_step(tag, os, arch, version='', with_browser=false) =
   // We can't use 'make docker' without making this repo priveleged in drone
   // so we will use the native docker plugin instead for security.
   local platform = std.join('/', [ os, arch, if std.length(version) > 0 then version ]);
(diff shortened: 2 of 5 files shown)


scripts/configs/drone/main.jsonnet
```diff
@@ -87,25 +87,38 @@ local docker_step(tag, os, arch, version='') =
...
+      target: if with_browser then 'with-browser' else 'release',
       build_args: [
         'TARGETPLATFORM=' + platform,
         'TARGETOS=' + os,
         'TARGETARCH=' + arch,
       ] + if std.length(version) > 0 then [
         'TARGETVARIANT=' + version,
+      ] else []
+       + if with_browser then [
+        'WITH_BROWSER=true',
```
roobre, line 99: What does this build arg do? I'm not finding it in the dockerfile 🤔

scripts/configs/drone/main.jsonnet
```diff
@@ -87,25 +87,38 @@ local docker_step(tag, os, arch, version='') =
...
     },
   };
 
-local docker_build(os, arch, version='') =
-  docker_step('docker build', os, arch, version)
+local docker_build(os, arch, version='', with_browser=false) =
+  local tag = if with_browser then
+      'docker build (with browser)'
+        else
+      'docker build';
```
roobre, line 108: Not sure I get what's going on here. Why are we setting `tag` to a string with spaces and parentheses? If this isn't the image tag I'd take this chance to add a comment here.

scripts/configs/drone/main.jsonnet
```diff
@@ -174,6 +191,24 @@ local docker_publish(repo, auth, tag, os, arch, version='') =
     ])
     + releaseOnly,
 
+    step(
+      'docker publish (with browser) tags',
+      [
+        '{ echo latest-with-browser,$(eval ./scripts/version)-with-browser ; } > .tags',  // use with-browser tags for docker plugin
```
roobre, line 197: I'd suggest to use a shorter `-browser` suffix instead of `-with-browser`, just for the sake of brevity. I think the short version does not leave much room for ambiguity. WDYT?

Dockerfile
```diff
@@ -18,3 +18,18 @@ COPY scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
+
+# third stage with alpine base for better access to chromium
+FROM alpine:3.18 as with-browser
```
roobre, line 23: I think we're mixing two substantially different linux flavors here (alpine and debian) and I'm afraid of unforeseen consequences. Most prominently, the libc implementation is different, which may cause a binary to work in one but not the other. Since the binary is built outside of docker with many many layers of Makefiles, I don't know if it is being built with `CGO_ENABLED=0` (hope/think it is), otherwise it will break on alpine for sure. 

We're also copying the ca-certificates from debian into alpine, which should be okay... but looks strange.

If you have manually tested this works, I wouldn't block the PR for this, but I'd strongly suggest to make a follow-up PR migrating the base image to alpine. We've been looking for an excuse to do this to lower our number of CVEs, so this is an excuse as good as any other.

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get -y install ca-certificates
 
 ARG TARGETPLATFORM
 
-FROM --platform=$TARGETPLATFORM debian:stable-slim
+FROM --platform=$TARGETPLATFORM debian:stable-slim as release
```
roobre, line 10: This changes what is the "default" result of running `docker build` on this repo. I think it might be good to document these two targets and its differences in the `README`.

Dockerfile
```diff
@@ -18,3 +18,18 @@ COPY scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
...
 
 ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
+
+# third stage with alpine base for better access to chromium
+FROM alpine:3.18 as with-browser
+
+RUN apk --no-cache add chromium-swiftshader
+
+ENV SM_CHROME_BIN=/usr/bin/chromium-browser
+ENV SM_CHROME_PATH=/usr/lib/chromium/
```
roobre, line 28: What are these two for?

Dockerfile
```diff
@@ -18,3 +18,18 @@ COPY scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
...
+
+COPY --from=release /usr/local/bin/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
+COPY --from=release /usr/local/bin/sm-k6 /usr/local/bin/sm-k6
+COPY --from=release /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
+COPY --from=release /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
```
roobre, line 35: With browser added in, we'll have the agent start the k6 process, which will itself start a chromium process, which will spawn a gazillion processes itself. I think this calls for adding `tini` to make sure all those children are properly terminated.

Dockerfile
```diff
@@ -18,3 +18,18 @@ COPY scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
+
+# third stage with alpine base for better access to chromium
+FROM alpine:3.18 as with-browser
+
+RUN apk --no-cache add chromium-swiftshader
```
roobre, line 25: At this point we will be running everything as root. I think this can be left for a follow-up as well, but running a browser as root makes me a bit scared so we might want to avoid that.

roobre (commented):
Left a few small comments, and a few bigger ones that we can leave for follow-up PRs if this one has been tested working.

scripts/configs/drone/main.jsonnet
The-9880: Thank you for catching this -- I started with an ARG-based solution first and then migrated to a built target one, so this needs to be removed

scripts/configs/drone/main.jsonnet
The-9880: Yes this isn't the image tag, I was following the existing use of the `tag` param in this jsonnet file. I will rename this in the new usages to avoid confusion

scripts/configs/drone/main.jsonnet
The-9880: I agree! I noticed other Docker images like that for `k6` were using `-with-browser`, but I don't mind shortening it for us

Dockerfile
The-9880: Right. We have some `helper` make scripts on `docker build ...` which aren't using in any automations AFAICT. The result of those would build the new final stage instead of the old one so that is worth documenting as well but shouldn't impact any of our current pipelines I think.

Will update `README`!

Dockerfile
The-9880: I wondered if providing an env var might be useful instead of relying on PATH to have `chromium`/assuming the binary name, for when sm-agent leverages it.

Dockerfile
The-9880: 👍 I can add a user and operate the remainder of the target from there

Dockerfile
The-9880: I believe it is being built with `CGO_ENABLED=0` (@mem to confirm if you have a second).

I built and ran this Dockerfile locally and it seems to run the agent binary, though I didn't actually allocate a workload to it so maybe I'll do that and report back.

In either case, using Debian or Alpine, I've noticed that attempting to run headless chromium would result in an error I'm still trying to debug: `Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory`. I used `chromiuim`/`chromium-shell` on Debian in that case and got the same issue as `chromium-swiftshader` on Alpine.

I'm still working on understanding that, wondering if you have any insight as you got this running for crocochrome :eyes:

scripts/configs/drone/main.jsonnet
The-9880: Well, normally I would want it to be consistent with other releases but because we build a longer tag name for our images, including the commit hash, I think I like keeping it shorter for this case

Dockerfile
The-9880: If we land on agreeing that this is fine, I would for sure migrate the base to alpine as well.

Dockerfile
The-9880: Will we add `xk6-browser` to the k6 setup for this? We may want to add `K6_BROWSER_HEADLESS` and `K6_BROWSER_ARGS` for that

Dockerfile
roobre: I think that error message is benign, I also get that on crocochrome is my memory serves. Probably just one of the million things chrome is trying to run at startup.

Dockerfile
roobre: `xk6-browser` is now part of `k6`, so it's already there :D

I believe it should already be looking for the correct path, but we definitely want to tweak `K6_BROWSER_ARGS`. For instance, `--disable-dev-shm-usage` is required for some webpages to not crash on a container.

Dockerfile
roobre: README changes look good, thanks!

Dockerfile
roobre: This might fail in some corner case configurations (e.g. if both user namespaces and setuid bits are disabled) but I think it should work good enough for now!

roobre: Re-thinking this (sorry)... This might get us into problems where running the agent as non-root might cause some check types to fail. e.g. traceroute/ping checks require either elevated privileges or net_raw capabilities to be added to the binary.

It might be better for now to run everything as root, and open an issue to narrow-down the privileges without breaking agent checks.

roobre (approved):
If this passes some happy path tests, it looks good to me as a first iteration to unblock early adopters and internal tests.

I'd suggest to open a few follow-up issues to act on before we make this more widely available:
- Migrate base image to alpine to consolidate in one distro
- Pin and auto update chromium versions (Copy from https://github.com/grafana/crocochrome/pull/5 when it is merged)
- Run both agent and chromium as non-root, granting required capabilities (i.e. `net_raw`) to the agent binary.

scripts/configs/drone/main.jsonnet
roobre: Hmm... that's a good point, consistency is good... I don't see a clear winner but I'm still leaning towards `-browser` for short. Maybe we can ask the rest of the team to see if anyone has a strong opinion?

roobre:
Forgot to mention, making sure that `--disable-dev-shm-usage` is passed to chromium is probably something we should figure out in this PR.

Dockerfile
The-9880: I assigned some workloads to this and it seems to run fine on my linux VM -- I will resolve this thread, with the takeaway that we want to migrate the original base to alpine as well!

Dockerfile
The-9880: Done!

Dockerfile
The-9880: The K6 Dockerfile seems to do the following:
```
# no-sandbox chrome arg is required to run chrome browser in
# alpine and avoids the usage of SYS_ADMIN Docker capability
ENV K6_BROWSER_ARGS=no-sandbox
```

Should we be using no-sandbox as well? I'll add it tentatively to this PR, will re-request your review again before I merge anything.

scripts/configs/drone/main.jsonnet
```diff
@@ -174,6 +188,24 @@ local docker_publish(repo, auth, tag, os, arch, version='') =
     ])
     + releaseOnly,
 
+    step(
+      'docker publish (with browser) tags',
+      [
+        '{ echo latest-browser,$(eval ./scripts/version)-browser ; } > .tags',  // use with-browser tags for docker plugin
```
roobre, line 194: As a small nit, is there a reason for this command using eval and curly braces?
On my PC, this seems to work:
```console
echo "latest-browser,$(./scripts/version)-browser"
```

roobre (approved):
Looking good!

scripts/configs/drone/main.jsonnet
The-9880: That should work - updating!

The-9880:
Merging without re-approval as the only change was to sort out the `.drone.yml` conflict with main.

PR #839, opened 2024-08-27
renovate: fix grafana-build-tools dependency regex

roobre:
`\S+` is too wide of a match. For a line like this:

```
local go_tools_image = 'ghcr.io/grafana/grafana-build-tools:v0.15.0';
```

It would match `v0.15.0';`, which is not a valid version. `[\w-+.]` matches all characters allowed in semver.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -59,7 +59,7 @@
         ".*\\.mk$"
       ],
       "matchStrings": [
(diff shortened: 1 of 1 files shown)


ka3de (approved):

roobre:
Merged from CLI as github was a bit drunk
image

Issue #841, opened 2024-08-27
Migrate pipelines to Github Actions


nadiamoe:
There are roughly three pipelines that need to be migrated:

```[tasklist]
### Tasks
- [ ] "Push/pr" pipeline, a.k.a. build, lint, test, etc. #842
- [ ] Release pipeline for container images (dockerhub, ghcr)
- [ ] Release pipeline for linux packages
- [ ] Technically also part of the release pipeline, but better done on a separate PR, triggering argo workflows
```

PR #842, opened 2024-08-27
ci: migrate push/pr pipeline to GHA

roobre:
First step for https://github.com/grafana/synthetic-monitoring-agent/issues/841


roobre:
There is one (1) test failing in GHA but neither in Drone nor locally. This can be either trivial or a rabbit hole, hoping for the first one.

.github/workflows/push-pr.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - id: version
+        run: |-
+          echo "golangci=$(make golangci-lint-version)" >> $GITHUB_OUTPUT
+      # Use the golangci-lint action, which provides Github-specific features such as diff annotations.
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: ${{ steps.version.outputs.golangci }}
+          args: --timeout=5m # 1m is not enough, experimentally.
+          only-new-issues: true
```
mem, line 77: I do have a problem with this (ignoring the obvious bug): this doesn't match what `make lint` does. That's my biggest issue with CI/CD system: what the remote system does does not match what the developer sees locally, and that ends up in that nasty cycle of "fix it" / "push" / "try again" / "push" / ...

The (maybe not) obvious bug is that golangci-lint's configuration for this repository is located in `scripts/configs/golangci.yml`. That probably should be moved to `.golangci.yaml` because that has a couple of advantages.

I'm open to reading arguments in favor of using a GHA, but those have to be weighted against developer's sanity.

roobre: > The (maybe not) obvious bug is that golangci-lint's configuration for this repository is located in scripts/configs/golangci.yml. That probably should be moved to .golangci.yaml because that has a couple of advantages.

That's a good point. I'll take this chance to move it to the default location so other things can also pick it up.

> I'm open to reading arguments in favor of using a GHA, but those have to be weighted against developer's sanity.

I couldn't agree more! I'm all in for matching CI and local behavior (that's why among other things I push for self-contained Docker builds), but my definition of developer sanity also includes not having to go back and forth from `line:col` entries in the logs to the source code. Diff annotations are very convenient, and that's why I prefer using GHA-native things when available.

I'll add the `make lint` step to the pipeline too, so we don't lose linters. We'd be running golangci twice, but I think that's worth it.

.github/workflows/push-pr.yaml
```diff
@@ -0,0 +1,81 @@
...
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: Test
+        run: |
+          make build-xk6-linux-amd64 # Tests require a valid k6 binary in dist/k6
```
mem, line 59: If this is needed, it means there's a missing dependency in the makefile.

I vaguely remember adding that...

golangci.yml
mem, line 1: it's `.golangci.yaml` by default:

```
$ golangci-lint config path
.golangci.yaml
```

roobre: Cursed default to hidden files... Thanks for catching that!

.github/workflows/push-pr.yaml
roobre: I think it is there on another repo (perhaps the runner)? But yep, one test failed cryptically without this.

mem (changes requested):
Can we please hold this one?

The almost the same but slightly different approach we have in the repos is causing more than just a little bit of grief.

roobre:
@mem I think unifying across repos might be good (with caveats here and there), but could we do the unification work later on? While we get that figured out (which no doubt will take time), drone is causing trouble, requiring logins, breaking with grafana-build-tools images, etc.

CI pipelines are not a breaking change and can always be changed uniform later, and the work here is already done.

PR #844, opened 2024-08-27
 ci: migrate publish container workflow to GHA

nadiamoe:
Rebases https://github.com/grafana/synthetic-monitoring-agent/pull/842


mem:
This is done, isn't it?

PR #852, opened 2024-08-28
multihttp/script: fix conditional rendering of `redirects` param

roobre:
Prior to this pr, the `redirects` param to k6's `http/request` function was always rendered. This PR fixes the template so the param is only emitted when no headers are specified, in which case the number of redirects to follow will be the default value of 10.

If headers are specified, the `redirect` param is rendered and thus no redirects will be performed.

Related: https://github.com/grafana/synthetic-monitoring-app/issues/919

### DIFF
--- a/internal/prober/multihttp/script.tmpl
+++ b/internal/prober/multihttp/script.tmpl
@@ -88,7 +88,7 @@ export default function() {
 	{{ range $queries }}{{ . }};
 	{{ end -}}
 	{{- $method := .Request.Method.String }}
-	
+
 	body = {{ buildBody .Request.Body }};
 	{{- $bodyVars := interpolateBodyVars "body" .Request.Body }}
 	{{ range $bodyVars }}{{ . }};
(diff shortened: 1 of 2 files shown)


internal/prober/multihttp/script.tmpl
```diff
@@ -101,8 +101,10 @@ export default function() {
 		  name: '{{ $idx }}', // TODO(mem): give the user some control over this?
 		  __raw_url__: '{{ .Request.Url }}',
 		},
-		redirects: 0{{ if gt (len $headers) 0 }},
```
ka3de, line 104: Wasn't this setting redirects always to 0? The _if_ just affects the _headers_ param definition, doesn't it? Or am I getting it wrong? What I mean is that I don't see a reason not to remove redirects completely, which means that we will use k6 defaults (10), and later on, once this can be configured by the user, defined based on that new field.

roobre: You're getting it right here, I somehow misread the template here and thought it was printing an additional zero. Let's figure out what the original intention was, and change the PR when we know that.

internal/prober/multihttp/script.tmpl
```diff
@@ -101,8 +101,10 @@ export default function() {
...
-		headers: {{ $headers }}{{ end }}
+		{{- if gt ($headers | len) 0 }}
```
mem, line 104: Uhm...

This is changing what this is doing.

The template is outputting:
```
redirects: 0
```

If `$headers` contains something, it does this:

```
redirects: 0,
headers: xxx
```

The new version is not outputting anything, except in the case where $headers contains something, in which case it outputs:

```
redirects: 0,
headers: xxx
```

The handling of the redirects needs fixing, but not like this.

PR #856, opened 2024-08-29
Retire dependabot and fully replace it with Renovate

roobre:
After enabling renovate on this repo, we've had Renovate and Dependabot "coexisting", i.e. both creating PRs for the same thing.

This PR disables Dependabot and moves remaining bits of config it had to Renovate.


ka3de (approved):
I've wanted to discuss this for some time, but yes, since renovate seems to be our chosen option, this was needed. Thanks! I'll replicate this in sm-alerts also, and we can extend it to other repos that still have both coexisting.

PR #862, opened 2024-08-30
Output unadorned version tag for releases

mem:
There's stuff out there with pretty weird ideas as to what constitutes a "correct" version number. We have been trying to solve this problem of "incorrect versioning" for over 30 years with no success. You have to wonder why that is.

For the specific case of releases, in addition to using vX.Y.Z-0-sha, also output a vX.Y.Z tag. That allows stuff to be happy about seeing something that matches its expectations in terms of version numbers. We still need to output something else in the n != 0 case, because we need to be able to differentiate between intermediate builds. It's incorrect in semver terms, because those are prereleases that sort _before_ vX.Y.Z, not after it. To do it correctly, we would have to transform vX.Y.Z to vX.Y.(Z+1) and append the prerelease bit to that. Since that would become confusing really fast (in the face of existing practice), let's not do that. As long as nothing tries to "upgrade" vX.Y.Z-n-sha to vX.Y.Z, we shoud be OK.

### DIFF
--- a/scripts/generate-tags
+++ b/scripts/generate-tags
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+set -e
+set -u
+
+# If an optional suffix is provided, append it to all the tags as `-suffix`.
+# Note that this breaks the semver format, because appending -foobar means it's
+# a pre-release version, not a "flavor" as some people claim. The more
+# appropriate way to express this would be to use `+suffix`, but we'll defer
+# that until we have fixed the different places where this might be used.
+suffix=${1:+-}${1:-}
+
+rootdir=$(git rev-parse --show-toplevel)
+version=$("${rootdir}/scripts/version")
+
+version_base=$(echo "${version}" | cut -d- -f1)
+version_seq=$(echo "${version}" | cut -d- -f2)
+
+printf "%s," "latest${suffix}"
+printf "%s" "${version}${suffix}"
+if [ "${version_seq}" = "0" ]; then
+  # Strip the additional -0-* suffix to generate a additional tag that looks like vX.Y.Z
+  printf ",%s" "${version_base}${suffix}"
+fi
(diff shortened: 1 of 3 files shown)


scripts/generate-tags
```diff
@@ -0,0 +1,24 @@
...
+# a pre-release version, not a "flavor" as some people claim. The more
+# appropriate way to express this would be to use `+suffix`, but we'll defer
+# that until we have fixed the different places where this might be used.
+suffix=${1:+-}${1:-}
+
+rootdir=$(git rev-parse --show-toplevel)
+version=$("${rootdir}/scripts/version")
+
+version_base=$(echo "${version}" | cut -d- -f1)
+version_seq=$(echo "${version}" | cut -d- -f2)
```
roobre, line 17: By doing this, we are dropping the short sha out of the output of `/scripts/version`, is that okay?

```
10:01:04 ~/Devel/synthetic-monitoring-agent $> git describe --tags --always --long
v0.25.2-28-g3e038e9

10:01:47 ~/Devel/synthetic-monitoring-agent $> git describe --tags --always --long | cut -d- -f1
v0.25.2

10:01:08 ~/Devel/synthetic-monitoring-agent $> git describe --tags --always --long | cut -d- -f2
28
```

roobre (commented):
Approach looks good to me. Just want to double-check we want to drop the hash from the output of the version script.

scripts/generate-tags
mem: I'm not sure I follow...

For a non-tagged commit, this outputs:

```
latest,v0.25.2-34-g3fb991a7
```

For a tagged commit:

```
latest,v0.25.2-0-g5a039921,v0.25.2
```

roobre: My bad, I think I misread this and thought we were using `version_base` later on instead of the full `version`.

The-9880 (approved):

Issue #845, opened 2024-08-27
Migrate base image to Alpine


The-9880:
We recently merged an update to start releasing a version of the Agent image with chromium for browser checks - this is an Alpine-based image.

We should migrate the base image to Alpine as well to consolidate on one distro.

PR #875, opened 2024-09-08
Migrate base image from Debian to Alpine

The-9880:
Issue: https://github.com/grafana/synthetic-monitoring-agent/issues/845

We've expanded the Dockerfile into a multi-stage build to include chromium for a browser-enabled agent on Alpine. This PR migrates the base release to Alpine as well.

Tested locally on Colima using a linux/amd64 VM.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 # First stage obtains the list of certificates.
-FROM --platform=$BUILDPLATFORM debian:stable-slim@sha256:64bc71feaa7ec2ac758a6a3a37c0f0d6ebccf0a45e3f5af1f1d3b5d4cb316b29 AS build
-RUN apt-get update && apt-get -y install ca-certificates
+FROM --platform=$BUILDPLATFORM alpine:3.20.3 AS build
+RUN apk --no-cache add ca-certificates-bundle
 
 # Second stage copies the binaries, configuration and also the
 # certificates from the first stage.
 
 ARG TARGETPLATFORM
 
-FROM --platform=$TARGETPLATFORM alpine:3.20.3 as release
+FROM --platform=$TARGETPLATFORM alpine:3.20.3 AS release
 ARG TARGETOS
 ARG TARGETARCH
 ARG HOST_DIST=$TARGETOS-$TARGETARCH
@@ -21,7 +21,7 @@ ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
 
 # Third stage copies the setup from the base agent and
 # additionally installs Chromium to support browser checks.
-FROM alpine:3.20.3 as with-browser
+FROM --platform=$TARGETPLATFORM alpine:3.20.3 AS with-browser
 
 RUN apk --no-cache add tini
 RUN apk --no-cache add chromium-swiftshader
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 # First stage copies the binaries, configuration and installs the
 # certificates for the base agent.
-
 ARG TARGETPLATFORM
 
 FROM --platform=$TARGETPLATFORM alpine:3.20.3 as release
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,13 @@
-# First stage copies the binaries, configuration and installs the
-# certificates for the base agent.
+# First stage obtains the list of certificates.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,13 +1,13 @@
 # First stage obtains the list of certificates.
-FROM --platform=$BUILDPLATFORM debian:stable-slim@sha256:64bc71feaa7ec2ac758a6a3a37c0f0d6ebccf0a45e3f5af1f1d3b5d4cb316b29 AS build
-RUN apt-get update && apt-get -y install ca-certificates
+FROM --platform=$BUILDPLATFORM alpine:3.20 AS build
+RUN apk --no-cache add ca-certificates
```
roobre, line 3: I'd suggest to get rid of this stage and the subsequent `COPY --from=build /etc/ssl/certs/ca-certificates.crt ...` and just install ca-certificates on the final layer. I _think_ `apk --no-cache` should leave little leftovers, if at all, and will lower the overall complexity.

Dockerfile
```diff
@@ -1,13 +1,13 @@
...
 
 # Second stage copies the binaries, configuration and also the
 # certificates from the first stage.
 
 ARG TARGETPLATFORM
 
-FROM --platform=$TARGETPLATFORM debian:stable-slim@sha256:64bc71feaa7ec2ac758a6a3a37c0f0d6ebccf0a45e3f5af1f1d3b5d4cb316b29 as release
+FROM --platform=$TARGETPLATFORM alpine:3.20 as release
```
roobre, line 10: It might be good to pin the alpine patch too:

```suggestion
FROM --platform=$TARGETPLATFORM alpine:3.20.3 as release
```

Renovate takes care of updating this, so the stricter the pin, the more often we will see PRs bumping it.

roobre (commented):
Looking good! Just a few nits.

Dockerfile
The-9880: I'll update the version on the chromium build as well then

Dockerfile
The-9880: I think it might be worth leaving the `build` target in for now.

The build for linux/arm64/v8 [fails](https://drone.grafana.net/grafana/synthetic-monitoring-agent/2291/1/8) to run `apk --no-cache add ca-certificates` when building the release.

The-9880:
@roobre Innocent bump 👀 let me know what you think!

Dockerfile
roobre: That sounds a bit unsettling, but might be due to buildx shenanigans. I think we can revisit that when we fully migrate the build container pipeline to GHA.

roobre (approved):
Looking good, all hail alpine! 🎉

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get -y install ca-certificates
 
 ARG TARGETPLATFORM
 
-FROM --platform=$TARGETPLATFORM debian:stable-slim@sha256:64bc71feaa7ec2ac758a6a3a37c0f0d6ebccf0a45e3f5af1f1d3b5d4cb316b29 as release
+FROM --platform=$TARGETPLATFORM alpine:3.20.3 as release
```
mem, line 10:
```suggestion
FROM --platform=$TARGETPLATFORM alpine:3.20.3 AS release
```

Dockerfile
```diff
@@ -19,8 +19,9 @@ COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 
 ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
 
-# third stage with alpine base for better access to chromium
-FROM alpine:3.20@sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d as with-browser
+# Third stage copies the setup from the base agent and
+# additionally installs Chromium to support browser checks.
+FROM alpine:3.20.3 as with-browser
```
mem, line 24:
```suggestion
FROM --platform=$TARGETPLATFORM alpine:3.20.3 AS with-browser
```

Dockerfile
mem: You should change lines 2-3 to:

```
FROM --platform=$BUILDPLATFORM alpine:3.20.3 AS release
RUN apk --no-cache add ca-certificates-bundle
```

Dockerfile
The-9880: From mem's [suggestion](https://github.com/grafana/synthetic-monitoring-agent/pull/875#discussion_r1759574118), it looks like it works if we use `ca-certificates-bundle` instead

PR #886, opened 2024-09-12
Move browser Docker build into separate Dockerfile

The-9880:
Edit: realized the issue and revamped the PR.

Turns out the `target` for the multi-stage Dockerfile was not being set when publishing it - we were only setting it when doing the dry-run build. This explains why both images on Dockerhub included chromium.

This updates the `target` when publishing the image. Additionally publishes the browser image to GCR so we can confirm this works ASAP. When confirmed, I will remove that addition.


The-9880:
Hold on, I think I see something 🤦‍♂️

scripts/configs/drone/main.jsonnet
```diff
@@ -197,11 +197,20 @@ local docker_publish_with_browser(repo, auth, tag, os, arch) =
     )
     + dependsOn([ 'docker publish (release)' ]),
 
+    // TODO(the-9880): remove after release fix confirmed
+    docker_publish_with_browser(gcrio_repo, grcio_auth, 'gcr.io', 'linux', 'amd64') + devAndRelease,
```
roobre, line 201: Just a reminder so we don't forget about the TODO

roobre (approved):

Issue #846, opened 2024-08-27
Pin + auto-update Chromium version


The-9880:
Configure the chromium package in the browser release so that its version is pinned and auto-updated.

PR #890, opened 2024-09-15
renovate: update pinned alpine packages

The-9880:
Closes https://github.com/grafana/synthetic-monitoring-agent/issues/846.

Duplicates what @roobre's done to add a regex manager and a custom datasource, so renovate can update pinned alpine packages.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -61,6 +61,34 @@
       "matchStrings": [
         "ghcr.io/grafana/grafana-build-tools:(?<currentValue>[\\w-+.]+)"
       ]
-    }
-  ]
+    },
+    {
+      // Update pinned alpine packages in Dockerfile.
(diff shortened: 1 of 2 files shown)


roobre:
After some discussion, @mem and I more or less converged to the regex that is now on `main` (https://github.com/grafana/crocochrome/commit/b577bd75efa9269c277c212d626789e495250041). Would you be able to sync this one as well, so they are the same?

The-9880:
> After some discussion, @mem and I more or less converged to the regex that is now on `main` ([grafana/crocochrome@b577bd7](https://github.com/grafana/crocochrome/commit/b577bd75efa9269c277c212d626789e495250041)). Would you be able to sync this one as well, so they are the same?

Updated!

roobre (approved):

PR #898, opened 2024-09-19
Fix: allow probers to provide a duration value

mem:
Some probers need to override the duration that is reported, mostly because we use probe_duration_seconds as the "latency" of the probe, which isn't exactly correct. `probe_duration_seconds` was meant to be used as the duration of the probe: how long did it take for this probe to run. Capture that value in the logs for debugging purposes, and use `probe_duration_seconds` as if it meant the time it takes the target to produce a response, including *some* setup, but not e.g. the overhead introduced by retries or the time it takes for external programs to spin up.

This is OK to change since these are all internal packages.

### DIFF
--- a/internal/prober/icmp/icmp_impl.go
+++ b/internal/prober/icmp/icmp_impl.go
@@ -12,7 +12,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-func probeICMP(ctx context.Context, target string, module Module, registry *prometheus.Registry, logger log.Logger) (success bool) {
+func probeICMP(ctx context.Context, target string, module Module, registry *prometheus.Registry, logger log.Logger) (success bool, duration float64) {
 	var (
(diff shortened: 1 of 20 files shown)


mem:
This is the second time I have needed this functionality. The first time I didn't want the big noisy change, so I hacked around it.

For the second time around, I think it's better to do it right.

This will benefit all k6-based checks, too. Maybe traceroute, too?

internal/prober/icmp/icmp_impl.go
```diff
@@ -90,7 +91,9 @@ func probeICMP(ctx context.Context, target string, module Module, registry *prom
 
 	pinger.OnSetup = func() {
 		if !setupDone {
-			durationGaugeVec.WithLabelValues("setup").Add(time.Since(setupStart).Seconds())
+			setupDuration := time.Since(setupStart).Seconds()
```
mem, line 94: This is the actual change I need. :-P

PR #907, opened 2024-09-27
adhoc: clamp ad-hoc check timeout to at most 20s

roobre:
Ad-hoc checks are run pseudo-synchronously from the UI, which does not wait indefinitely for a check to run. This PR limits the check timeout to at most 20s, which we consider already more than what an user will wait for data to show up.


peterschretlen:
In the app we add a 30 second buffer beyond the configured timeout, to allow for long running checks, and give time for the logs to ingest. 

https://github.com/grafana/synthetic-monitoring-app/blob/d1b11aeccaf158b4fc25dd54e8a6845e37b125de/src/components/CheckTestResultsModal.tsx#L97

I realize waiting longer than 20s for results seems long, but if we do this there is no way to validate a check running longer than 20s without actually saving it?   I think we should respect the timeout requested by the UI.

roobre:
No need to do this if the timeout in the UI is not hardcoded.

PR #908, opened 2024-09-27
feat: adhoc: add some grace time to k6 check timeouts

roobre:
k6-backed checks need some grace time between the check timeout, which is enforced by the runners, and the request timeout, which is the time for which the agent will wait to hear back from the runners. If those numbers are the same, the agent is almost guaranteed to give up on the runner before the latter is able to communicate that a timeout occurred, yielding an unhelpful error message (as the agent does not know what happened).

### DIFF
--- a/internal/adhoc/adhoc.go
+++ b/internal/adhoc/adhoc.go
@@ -392,13 +392,23 @@ func (h *Handler) defaultRunnerFactory(ctx context.Context, req *sm.AdHocRequest
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


ka3de (approved):

ka3de:
Should the commit message be named such as: `feat(adhoc): add grace time to k6-backend checks`? :innocent:

roobre:
I changed the PR title and squashed thinking that would change the commit message, but it didn't 😓

PR #923, opened 2024-10-02
k6runner/http: add tracing support

roobre:
Very wip, here be dragons.


cmd/synthetic-monitoring-agent/main.go
```diff
@@ -176,6 +171,16 @@ func run(args []string, stdout io.Writer) error {
 		zerolog.SetGlobalLevel(zerolog.ErrorLevel)
 	}
 
+	baseCtx, err := tracing.Context(context.Background())
+	if err != nil {
+		zl.Err(err).Msg("could not configure tracer provider, disabling tracing functionality")
+	}
```
mem, line 177: This is what I mean elsewhere: I read this code and I go like "disabling? you aren't disabling anything! I can see you are using the context as if it was OK".

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -176,6 +171,16 @@ func run(args []string, stdout io.Writer) error {
...
+
+	baseCtx, cancel := context.WithCancel(baseCtx)
+	defer cancel()
```
mem, line 180: this is possibly nitpicking, but I would have expected this to come first.

internal/prober/browser/browser.go
```diff
@@ -60,12 +62,18 @@ func (p Prober) Name() string {
 }
 
 func (p Prober) Probe(ctx context.Context, target string, registry *prometheus.Registry, logger logger.Logger) (bool, float64) {
+	ctx, probeSpan := tracing.TracerProvider(ctx).Tracer("").Start(ctx, "browser check")
```
mem, line 65: I almost want to say the span should start one level up.

The fact that the code is repeated across a bunch of different probers makes me think that's probably correct.

`Tracer("")` is one of those "ugh!" otel things...

and `Tracer(...)` is the other one, where Java slipped into Go. What that thing is doing is going to a global (!!!) registry of tracers and pulling one out. I have always found it so confusing that I almost want to say that's what should be in the context instead of the tracer provider.

But that's the reason why I think the span should start one level up, because it would be useful to have the check ID, tenant ID, region ID, probe name, and possibly a couple of bits of info in the trace, and that should be added to the span.

It might make sense for the tracer to have the name of the probe as an attribute, but I never remember which bit of info ends up where. I think the cluster and instance are getting added by Alloy?

mem: Also, the prober knows its name, you should use that instead of hard coding `"browser check"`

internal/k6runner/http.go
```diff
@@ -106,14 +109,18 @@ func (r HttpRunner) Run(ctx context.Context, script Script) (*RunResponse, error
 		waitRemaining := max(0, wait-time.Since(start))
 		r.logger.Warn().Err(err).Dur("after", waitRemaining).Msg("retrying retryable error")
 
+		// Discard context, backoffSpan needs no children.
```
mem, line 112: Sorry, you lost me. If it has no children, and it's not adding any data, why is it interesting?

internal/k6runner/http.go
```diff
@@ -165,7 +172,21 @@ func (r HttpRunner) request(ctx context.Context, script Script) (*RunResponse, e
...
+			http.DefaultTransport,
+			otelhttp.WithTracerProvider(trace.SpanFromContext(ctx).TracerProvider()),
+			// Span names do not include method and path by default to avoid cardinality explosion with paths containing
+			// IDs. As this is not the case with this endpoint, we use a custom formatter that includes both.
+			otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
+				return fmt.Sprintf("%s %s", r.Method, r.URL.Path)
+			}),
+			otelhttp.WithPropagators(propagation.TraceContext{}), // Send TraceIDs in outgoing requests.
+		),
+	}
```
mem, line 187: I distrust otel a lot...

Isn't this doing a lot of extra work over and over again?

Wouldn't it be better to create a client once (e.g. `sync.OnceValue`) 

Also, this line:

```
otelhttp.WithTracerProvider(trace.SpanFromContext(ctx).TracerProvider()),
```

isn't the tracer provider stored in the context? You don't need to extract the span to extract its provider, you can ask for the provider directly from the context. I guess there are situations where you have multiple tracer provider, but unless I'm missing something this isn't one of them.

internal/k6runner/k6runner.go
```diff
@@ -92,15 +94,27 @@ var (
 func (r Processor) Run(ctx context.Context, registry *prometheus.Registry, logger logger.Logger, internalLogger zerolog.Logger) (bool, error) {
 	k6runner := r.runner.WithLogger(&internalLogger)
 
-	// TODO: This error message is okay to be Debug for local k6 execution, but should be Error for remote runners.
-	result, err := k6runner.Run(ctx, r.script)
+	k6Ctx, k6Span := trace.SpanFromContext(ctx).TracerProvider().Tracer("").Start(ctx, "k6 processor")
```
mem, line 97: This is also, again, the reason why I think you should store the tracer and not the tracerprovider in the context. This would become:

```
tracing.TracerFromContext(ctx).Start(ctx, "k6 processor")
```

Because the context contains a span, this creates a subspan with the current span as parent.

This aslo avoids the weird `Tracer("")` thing.

mem (changes requested):
Thank you.

Sorry about the comments all over the place. I really dislike the otel API, it's messy, complicated and it has very low discoverability. A lot of the comments come from that place.

cmd/synthetic-monitoring-agent/main.go
roobre: I am using the context as if it is okay because it _is_ okay: It either has a working tracerprovider, or a noop tracerprovider, but there will be one. The log message is just telling a white lie because "using noop tracerprovider" is probably more confusing to users.

internal/k6runner/k6runner.go
roobre: I kid you not when I say I had exactly that code written, it was called `tracing.Tracer(ctx)` and it basically did `return tracing.TracerProvider(ctx).Tracer("")`.

I decied to remove it because as horrible as they may be, otel idioms and incantations are universal. The `""` thing is awful and the line is long, but it is the same thing everybody uses.
Very much like those nasty `panicIfNil(error)` functions you see in demo code. It is verbose, but it is the verbose everybody (that has used the otel sdk) will understand.

internal/k6runner/http.go
roobre: I think it might answer the question of _what_ is happening between requests. Not a hard opinion there, if you think it's obvious I can drop it.

internal/k6runner/http.go
roobre: > Isn't this doing a lot of extra work over and over again?

Theoretically, yes. We should be able to store this client somewhere, but figuring out where was more work than just creating it anew.

roobre: > isn't the tracer provider stored in the context? You don't need to extract the span to extract its provider, you can ask for the provider directly from the context. I guess there are situations where you have multiple tracer provider, but unless I'm missing something this isn't one of them.

Also possible. My justification is as follows: Storing the span in the context and retrieving the tracerprovider from it is "the otel way". I will do it like that everywhere I can. Storing the tracerprovider in the context directly is a hack I made up to avoid adding the tracerprovider to the constructor of every scraper and prober in the code. I will do it only where necessary, so if at some poing we want to drop the hack (the `tracing` package), we will need to change the least amount of code possible.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,66 @@
...
+// Context creates a context with a TracerProvider added to it which the TracerProvider and Tracer functions can
+// retrieve. It attempts to create a TracerProvider which exports traces over OTLP+HTTP to the URL defined in the
+// environment. If that's not possible, it returns the supplied context unnmodified and an error.
+// If both OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_EXPORTER_OTLP_TRACES_ENDPOINT are unset, instead of returning a broken
+// exporter (otel default behavior), this function returns none.
+func Context(ctx context.Context) (context.Context, error) {
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") == "" && os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT") == "" {
+		// Otel is not configured, do not enable tracing.
+		return ctx, nil
+	}
```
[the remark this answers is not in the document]
roobre: Unfortunately checking for env var emptyness is hard to avoid. The otel library has a semi-complex logic of preferring one or the other that I'd like to preserve intact, as such is **the otel way**. Using env vars to configure tracing is also the otel way.

Unfortunately for us, if they are unset the otel libraries will not return an error, but a broken tracer.

Again I'd rather not try to fix fundamental issues with the otel libraries, but stick to what everybody is doing and expecting.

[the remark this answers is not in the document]
roobre: I've moved the env-checking code and fallback-to-noop to main, LMK if this looks better.

cmd/synthetic-monitoring-agent/main.go
roobre: Makes sense, should be fixed now :)

internal/k6runner/http.go
mem: > Theoretically, yes. We should be able to store this client somewhere, but figuring out where was more work than just creating it anew.

Note there's a subtle change here: when using DefaultClient, the client is not getting created over and over again. Clients _are_ reusable. If I remember things correctly, reusing clients allows the underlying code the reuse *connections*.

> My justification is as follows: Storing the span in the context and retrieving the tracerprovider from it is "the otel way" 

I know. I'm not debating that. I'm debating that the otel way doesn't buy you anything in this case. In otel it's theoretically possible to have multiple tracer providers (I challenge you to find something in the documentation that explains _why_ you might want to do that) but within the actual use case we have here, I cannot see that happening.

> To avoid adding the tracerprovider to the constructor of every scraper and prober in the code.

Hmm...

The current implementation is confusing.

- The tracerprovider is in the context, but you don't want to use it.
- Looking at the line of code I flagged, I'm not sure whether that context has or hasn't a span.

The tracerprovider should be added to the scraper and passed down from there to the point where the root span should be created, which is probably the runProber function in scraper.go. There a context is created, which is where the span should be added.

If you do that, the only way to get a tracer provider is from the span, and that resolves the issue of why it's done one way or the other.

internal/k6runner/k6runner.go
mem: > otel idioms and incantations are universal.

Please remind me to talk about this elsewhere.

mem: > Same argument as above: Storing the span in the context and retrieving the tracerprovider from it is "the otel way". I will do it like that everywhere I can.

I'll accept that, but then please remove the tracer provider from the context. It makes things hard to reason about.

mem (changes requested):

internal/prober/browser/browser.go
mem: > I'm fine moving it one up, what I'm not sure is if we do want traces for non-k6 checks. The possibility is there, but do we actually have a use for them?

I would rather have the consistency.

If traces for protocol checks aren't useful, we can configure a sampling policy to drop those.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -176,6 +173,24 @@ func run(args []string, stdout io.Writer) error {
 		zerolog.SetGlobalLevel(zerolog.ErrorLevel)
 	}
 
+	baseCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var tracerProvider trace.TracerProvider = noop.NewTracerProvider()
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") != "" || os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT") != "" {
```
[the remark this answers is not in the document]
roobre: > This is still a problem.

I don't see the problem. What you suggest is what I would do if this SDK was used in vacuum, but it is not. The entire ecosystem is build around the expectation that you set those environment variables and it works. It's not just the trace exporter, it's the sampling algorithm, parameters, all of it is configured automatically from environment variables. We would be going out of our way to change how one thing is configured, and by doing that we make that thing inconsistent with the rest of the ecosystem, and also with the rest of the things that can be configured in this very project. I really don't think that's a good idea.

> you to realize how problematic the functional options pattern is

I do, I don't like it either. Big fan of opts structs myself.

internal/tracing/tracing.go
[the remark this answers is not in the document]
roobre: Continuing discussion elsewhere.

internal/k6runner/http.go
roobre: To be fair, I haven't added backoff spans on other places so I might as well drop this one and avoid the confusion entirely. We can add it later if it becomes a problem.

internal/k6runner/http.go
roobre: > The current implementation is confusing.

I will grant you that, it's not the clearest. I was hoping the comment would help, but it's true that it creates some ambiguity. Let me try to drop the context thingie and pass the tracerprovider everywhere. It will be a noisy PR.

internal/k6runner/k6runner.go
roobre: > I'll accept that, but then please remove the tracer provider from the context. It makes things hard to reason about.

Giving that a try, let's continue the discussion on the thread above.

cmd/synthetic-monitoring-agent/main.go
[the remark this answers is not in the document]
roobre: I've reworked the stuff. At this point it didn't make sense to keep with the `fixup!`s, so I've rebased the entire thing. Changes roughly are:
- Objects now receive the `TracerProvider` normally, not through the context. I can hardly find words for how painful it was to add this on everywhere.
- Root spans are now initiated by the scraper and the adhoc handler. Probers no longer know what a trace is. This means we lose the ability to record probe errors in a span, as Probers hide that behind the `success` boolean. We can create a child span from the Prober just with that info, but that would be duplicating it on every prober. Your call.

internal/k6runner/http.go
roobre: I've reworked the stuff. At this point it didn't make sense to keep with the `fixup!`s, so I've rebased the entire thing. Changes roughly are:
- Objects now receive the `TracerProvider` normally, not through the context. I can hardly find words for how painful it was to add this on everywhere.
- Root spans are now initiated by the scraper and the adhoc handler. Probers no longer know what a trace is. This means we lose the ability to record probe errors in a span, as Probers hide that behind the `success` boolean. We can create a child span from the Prober just with that info, but that would be duplicating it on every prober. Your call.

internal/scraper/scraper.go
```diff
@@ -524,6 +532,18 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		timeout = time.Duration(s.check.Timeout) * time.Millisecond
 	}
 
+	ctx, checkSpan := s.tracerProvider.Tracer("").Start(ctx, "check")
+	defer checkSpan.End()
+	checkSpan.SetAttributes(
+		attribute.String("type", s.prober.Name()),
+		attribute.String("region", s.probe.Region),
```
roobre, line 539: This seems to be the probe region (as expected), not the tenant region. Not sure where to get the latter from.

cmd/synthetic-monitoring-agent/main.go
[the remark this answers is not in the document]
mem: > The entire ecosystem is build around the expectation that you set those environment variables and it works.

The ecosystem is built around an ill-defined notion of "ease of use". I have had to debug issues with Otel-using libraries that obtain configuration from the environment "because it's there".

I do not want stuff happening behind our backs, and more importantly I do not want stuff happening behind our user's backs. I don't want users accidentally pushing traces. If they do want to push traces, they can explicitly ask for it.

I have added a commit on top of this PR. If you are not OK with that, remove it.

PR #928, opened 2024-10-07
k6runner: add check metadata and type to remote runner requests

roobre:
Similar approach to https://github.com/grafana/synthetic-monitoring-agent/pull/904, but I chose to make a distinction between pure metadata, that is, data that makes it only to the telemetry and does not take part in any decision making, and actual information that is used for decision making. The latter is strongly typed.

### DIFF
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/go-logfmt/logfmt"
 	"github.com/grafana/synthetic-monitoring-agent/internal/prober/logger"
+	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
@@ -20,17 +21,50 @@ import (
 
 // Script is a k6 script that a runner is able to run, with some added instructions for that runner to act on.
 type Script struct {
-	Script   []byte   `json:"script"`
+	// Script is the blob of bytes that is to be run.
+	Script []byte `json:"script"`
+	// Settings is a common representation of the fields common to all implementation-specific check settings that the
+	// runners are interested about.
 	Settings Settings `json:"settings"`
-	// TODO: Add Metadata and Features.
+	// CheckInfo holds information about the SM check that triggered this script.
+	CheckInfo CheckInfo `json:"check"`
+	// TODO: Add features.
 }
 
+// Settings is a common representation of the fields common to all implementation-specific check settings that the
+// runners are interested about.
 type Settings struct {
 	// Timeout for k6 run, in milliseconds. This value is a configuration value for remote runners, which will instruct
 	// them to return an error if the operation takes longer than this time to complete. Clients should expect that
 	// requests to remote runners may take longer than this value due to network and other latencies, and thus clients
 	// should wait additional time before aborting outgoing requests.
 	Timeout int64 `json:"timeout"`
+}
+
+// CheckInfo holds information about the SM check that triggered this script.
+type CheckInfo struct {
+	// Type is the string representation of the check type this script belongs to (browser, scripted, multihttp, etc.)
+	Type string `json:"type"`
+	// Metadata is a collection of key/value pairs containing information about this check, such as check and tenant ID.
+	// It is loosely typed on purpose: Metadata should only be used for informational properties that will make its way
+	// into telemetry, and not for making decision on it.
+	Metadata map[string]any `json:"metadata"`
+}
+
+// CheckInfoFromSM returns a CheckInfo from the information of the given SM check.
+func CheckInfoFromSM(smc sm.Check) CheckInfo {
+	ci := CheckInfo{
+		Metadata: map[string]any{},
+	}
+
+	ci.Type = smc.Type().String()
+	ci.Metadata["id"] = smc.Id
+	ci.Metadata["tenantID"] = smc.TenantId
+	ci.Metadata["regionID"] = smc.RegionId
+	ci.Metadata["created"] = smc.Created
+	ci.Metadata["modified"] = smc.Modified
+
+	return ci
 }
 
 // ErrNoTimeout is returned by [Runner] implementations if the supplied script has a timeout of zero.
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -9,8 +9,8 @@ import (
 	"strings"
 
 	"github.com/go-logfmt/logfmt"
+	smmmodel "github.com/grafana/synthetic-monitoring-agent/internal/model"
 	"github.com/grafana/synthetic-monitoring-agent/internal/prober/logger"
-	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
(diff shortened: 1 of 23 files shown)


internal/k6runner/k6runner.go
```diff
@@ -33,6 +41,31 @@ type Settings struct {
...
 
+// CheckInfo holds information about the SM check that triggered this script.
+type CheckInfo struct {
+	// Type is the string representation of the check type this script belongs to (browser, scripted, multihttp, etc.)
+	Type string `json:"type"`
+	// Metadata is a collection of key/value pairs containing information about this check, such as check and tenant ID.
+	// It is loosely typed on purpose: Metadata should only be used for informational properties that will make its way
+	// into telemetry, and not for making decision on it.
+	Metadata map[string]any `json:"metadata"`
+}
```
roobre, line 52: This is where I'd like to have more reviewer attention, as these changes are hard to revert.

I chose to separate actionable information (`Type`), which will be used downstream to make decisions, from non-actionable data that will be basically used to enrich logging. I think this makes the relevant information "strongly typed", if that makes sense. Looking forward to your thoughts on this.

internal/k6runner/k6runner.go
```diff
@@ -33,6 +41,31 @@ type Settings struct {
...
+func (c CheckInfo) FromSM(smc sm.Check) CheckInfo {
+	c.Type = smc.Type().String()
+
+	if c.Metadata == nil {
+		c.Metadata = map[string]any{}
+	}
+	c.Metadata["id"] = smc.Id
+	c.Metadata["tenantID"] = smc.TenantId
+	c.Metadata["created"] = smc.Created
+	c.Metadata["modified"] = smc.Modified
```
mem, line 64: I don't love the merge nature of this, and I do have a problem with the signature of the method.

```
anotherCheckInfoWithRandomGarbageInIt := checkInfoWithRandomGarbageInIt.FromSM(check)
```

`anotherCheckInfoWithRandomGarbageInIt` has the same map as `checkInfoWithRandomGarbageInIt`. That's not cool. Please note: https://go.dev/play/p/B9HNY11N8c4

This code in the test is weird:

```
ci := CheckInfo{}.FromSM(check)
```

I would prefer a simple constructor approach:

```
func NewCheckInfo(check sm.Check) CheckInfo

ci := NewCheckInfo(check)
```

If you need the merging semantics, I would need to see an example to understand why.

mem: Mostly out of curiosity, why do you want created and modified in there?

If anything, I would add the version...

mem: And if you are including the tenant ID you *must* include the region ID.

mem (changes requested):

internal/k6runner/k6runner.go
roobre: > That's kind of the reason for using map[string]string instead of map[string]any.

I'm fine-ish with `map[string]string` for metadata. I leaned towards `any` because that will show integers as integers in the runner logs, instead of strings. If at some point we want to correlate logs, I think it would be nice that what is logged as an int in one place is also logged as an int on another.

> Unless you introduce actual coupling between the agent and the runner, which is not going to happen, this is all in internal/, the "strong type" argument is not very strong. I'm not saying it's wrong, just that it's not giving you what you say it gives you.

Practically it's giving nothing, but I think the separation conveys meaning to readers. They will notice that particular field is special and think twice before adding or removing it.

internal/k6runner/k6runner.go
```diff
@@ -20,11 +21,18 @@ import (
...
 type Script struct {
-	Script   []byte   `json:"script"`
+	// Script is the blob of bytes that is to be run.
+	Script []byte `json:"script"`
+	// Settings is a common representation of the fields common to all implementation-specific check settings that the
+	// runners are interested about.
 	Settings Settings `json:"settings"`
-	// TODO: Add Metadata and Features.
+	// Check holds information about the SM check that triggered this script.
+	Check CheckInfo `json:"check"`
```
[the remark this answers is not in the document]
roobre: Makes sense.

internal/k6runner/k6runner.go
```diff
@@ -33,6 +41,31 @@ type Settings struct {
...
+	Type string `json:"type"`
+	// Metadata is a collection of key/value pairs containing information about this check, such as check and tenant ID.
+	// It is loosely typed on purpose: Metadata should only be used for informational properties that will make its way
+	// into telemetry, and not for making decision on it.
+	Metadata map[string]any `json:"metadata"`
+}
+
+// FromSM populates the given Check from the information of the SM check. Existing fields are overwritten.
+func (c CheckInfo) FromSM(smc sm.Check) CheckInfo {
+	c.Type = smc.Type().String()
```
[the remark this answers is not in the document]
roobre: Same as discussed here: https://github.com/grafana/synthetic-monitoring-agent/pull/928#discussion_r1793329085
Type safety is not there, but I think the difference makes the developer aware this field is special and used for more things other than logging. Let's keep the discussion there.

internal/k6runner/k6runner.go
roobre: > And if you are including the tenant ID you must include the region ID.
That's a good catch, I definitely missed that

> Mostly out of curiosity, why do you want created and modified in there?
No particular reason other than "seemed vaguely useful for debugging". Might be useful if we want to plot the age distribution from logs? Again, no strong reason.

roobre: Changed the weird merge thingie to a constructor.

internal/k6runner/k6runner.go
mem: OK, I'm good with that, I just wanted to make sure the expectations are correct.

Do note that JSON numbers are super weird, because the JS in JSON stands for JavaScript, and JavaScript has no integers, only floats.

JSON mixed with Go is even weirder, because, when dealing with `any`, Go says "if JSON numbers are floats then I will decode them into float64 because that is what makes sense". You can override that using [UseNumber](https://pkg.go.dev/encoding/json#Decoder.UseNumber) (which requires using a `json.Decoder` not just `json.Unmarshal`) plus asking nicely for the specific conversion you want.

Do note this has an implication when marshaling elements from `map[string]any` into JSON via zerolog: the JSON decoder placed a float64 in the map's value, so `logger.Any(...)` sees a float, and it will log it out as a float. If you go thru the `json.Number` route, I think that should do the right thing (because `json.Number`'s underlying type is `string` 😆).

internal/k6runner/k6runner.go
mem: > That's a good catch, I definitely missed that. ... Where is it tho? I cannot find it in the check.

Argh. You have the wrong type of check because you are in the Prober, which receives an sm.Check, and you want a model.Check.

in scraper.go, in `collectData` you still have access to it because the scraper holds the correct type of check.

This calls `getProbeMetrics`, which no longer has access to that information.

This calls `runProber`, which is what calls `prober.Probe`, which is what ends up calling

internal/k6runner/k6runner.go
```diff
@@ -33,6 +41,31 @@ type Settings struct {
...
+	// Type is the string representation of the check type this script belongs to (browser, scripted, multihttp, etc.)
+	Type string `json:"type"`
+	// Metadata is a collection of key/value pairs containing information about this check, such as check and tenant ID.
+	// It is loosely typed on purpose: Metadata should only be used for informational properties that will make its way
+	// into telemetry, and not for making decision on it.
+	Metadata map[string]any `json:"metadata"`
+}
+
+// CheckInfoFromSM returns a CheckInfo from the information of the given SM check.
+func CheckInfoFromSM(smc sm.Check) CheckInfo {
```
mem, line 55: This is what needs a `model.Check` instead of a `sm.Check`.

mem (approved):
This is good, after adding the region_id

internal/k6runner/k6runner.go
roobre: That's another of those 100 line changes across internal tests and tens of implementations, unfortunately. It will take some time.

roobre:
@mem Painstakingly added `model.Check` everywhere, just want to double-check if this is what you had in mind.

mem (approved):
Thank you!

PR #966, opened 2024-10-30
feat: fetch precompiled xk6 extension from `grafana/xk6-sm`

nadiamoe:
This PR removes the `sm` extension code from this repo and adjust Makefiles and Dockerfiles to pull k6 from https://github.com/grafana/xk6-sm.

A bullet-point summary
- `$(DISTDIR)/$(OS)-$(ARCH)/sm-k6` targets have been added to `curl` k6 from grafana/xk6-sm releases
  - `sm-k6-native` is a convenient alias for the host's OS/ARCH, while `sm-k6` downloads sm-k6 for all known architectures
- No dependencies have been added to `test` or `package` targets. They can be, but as the package target does not depend on build, I followed that pattern.
  - Instead, `make sm-k6` is manually invoked in the pipeline.
- ~I've bothered to make goreleaser work with this, just to have drone not fail while we get to remove it. For that, the `build` target needs to depend on `sm-k6-native`. If we decide to not keep dependencies, we can remove this when we ditch goreleaser (details on why this was needed are on a comment).~

Part of https://github.com/grafana/synthetic-monitoring/issues/160


roobre:
Drone CI fails as it is going to be decommissioned soon, and agreed with @mem that it is not worth maintaining.

roobre:
@mem Rebased!

mem (approved):

roobre:
⚠️ There is a bug in `xk6-sm` that breaks browser metrics. We should change this PR to point to the last known good release of `xk6-sm` before merging.

PR #979, opened 2024-11-06
Fix: simplify TestTenantPusher

mem:
CI is failing consistently with this test. The best I can tell, it might be because some data is shared between tests, and the test might be time-sensitive.


ka3de (approved):
Thanks.

PR #1052, opened 2024-11-18
Simplify Dockerfile, remove redundant `Dockerfile.build`.

roobre:
This PR simplifies the Dockerfile so the `browser` flavor is based on the regular one, speeding up build and reducing duplication, as well as making the non-browser image the default and updating the README accordingly.

On the CI side, it changes the pipeline to use that Dockerfile (instead of Dockerfile.build).


mem (changes requested):
This cannot be merged like this, see comment.

roobre:
@mem I'm just toying with this for the moment, changing things and see what breaks to understand better the pipeline. I have no intention to merge it as-is (that's why it's a draft)

roobre:
This will go away soon as we remove drone.

PR #1056, opened 2024-11-20
chore: Dockerfile: revert split installation for amd64/arm64

roobre:
Reverts most of #1053. My understanding is that what happened is a very rare event, and I don't think the complexity of the workaround is worth having.


roobre:
Will be replaced by https://github.com/grafana/chromium-swiftshader-alpine/

PR #1061, opened 2024-11-22
Base browser image on grafana/chromium-swiftshader-alpine

roobre:
The new base image is built at https://github.com/grafana/chromium-swiftshader-alpine, and should allow us to both get chromium updates as PRs, while at the same time being stable so an outdated version does not break builds in the foreseeable future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,7 @@ RUN apk --no-cache add ca-certificates-bundle
 # Second stage copies the binaries, configuration and also the
 # certificates from the first stage.
 
-ARG TARGETPLATFORM
-
-FROM --platform=$TARGETPLATFORM alpine:3.20.3 AS release
+FROM alpine:3.20.3 AS release
 ARG TARGETOS
 ARG TARGETARCH
 ARG HOST_DIST=$TARGETOS-$TARGETARCH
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,18 +19,9 @@ ENTRYPOINT ["/usr/local/bin/synthetic-monitoring-agent"]
...
 
-# Renovate updates the pinned packages below.
-# The --repository arg is required for renovate to know which alpine repo it should look for updates in.
-# To keep the renovate regex simple, only keep one package installation per line.
-# Furthermore, we split this into two lines to allow for the arm64 and amd64 versions of chromium to be different, as 
-# they have drifted in the past.
-RUN apk --no-cache add --repository community tini=0.19.0-r3
-RUN [[ "$TARGETARCH" != "amd64" ]] || apk --no-cache add --repository community --arch x86_64 chromium-swiftshader=131.0.6778.69-r0
-RUN [[ "$TARGETARCH" != "arm64" ]] || apk --no-cache add --repository community --arch aarch64 chromium-swiftshader=131.0.6778.69-r0
+RUN apk --no-cache add --repository community tini
```
roobre, line 24: I think `tini` is fairly stable and doesn't have much activity, so it's probably not worth setting renovate to update packages just for it. I can bring it back if someone things it's still useful.

mem (approved):

PR #1063, opened 2024-11-25
feat: k6runner: improve error handling for k6 output

### LINKED ISSUE #1060: Backport cloud k6 runner improvements to k6 local runner
The cloud runner has a great deal of custom code to deal with the particulars of how `k6` reports errors and exit codes. This code should be ported to the local runner used in private probes to get consistent behavior.

nadiamoe:
Closes #1060

This pull requests brings a number of improvements to how the local k6 runner, mostly used in private probe context, handles errors from k6. Most prominently:
- For errors that we can attribute to the script contents (user errors), it no longer returns `nil, err` (understood as "k6 could not be run"), but rather a response with a textual representation of the error, and the k6 output. This still causes the test to be marked as failed, but the error is reported to Loki and thus made visible.
- Logs are now parsed and searched for specific strings that, in SM context, are considered fatal errors.
- Limits are imposed on how many bytes will be read for log and metrics output. These are last-resort measures to prevent the whole payload to be rejected by mimir or loki if it is too big.

Based on https://github.com/grafana/synthetic-monitoring-agent/pull/1061 as I want CI/CD to be able to run.

### DIFF
--- a/internal/k6runner/local.go
+++ b/internal/k6runner/local.go
@@ -1,9 +1,12 @@
 package k6runner
 
 import (
+	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"os/exec"
 	"time"
 
@@ -25,6 +28,8 @@ func (r Local) WithLogger(logger *zerolog.Logger) Runner {
 }
 
 func (r Local) Run(ctx context.Context, script Script) (*RunResponse, error) {
+	logger := r.logger.With().Object("checkInfo", &script.CheckInfo).Logger()
+
 	afs := afero.Afero{Fs: r.fs}
 
 	checkTimeout := time.Duration(script.Settings.Timeout) * time.Millisecond
@@ -39,7 +44,7 @@ func (r Local) Run(ctx context.Context, script Script) (*RunResponse, error) {
 
 	defer func() {
 		if err := r.fs.RemoveAll(workdir); err != nil {
-			r.logger.Error().Err(err).Str("severity", "critical").Msg("cannot remove temporary directory")
+			logger.Error().Err(err).Str("severity", "critical").Msg("cannot remove temporary directory")
 		}
 	}()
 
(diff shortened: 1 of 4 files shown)


devtony15:
hello @roobre, thank you for your work! Do you have an estimate for when this might be merged?

roobre:
Hey @devtony15, I expect to finish testing this week, so hopefully it will be merged a few days later. As per Grafana policy we have stopped making releases due to the holiday season, so my expectation is for this to be released early January.

devtony15:
@mem hey, could you review it please?)

internal/k6runner/local.go
```diff
@@ -115,36 +120,68 @@ func (r Local) Run(ctx context.Context, script Script) (*RunResponse, error) {
...
+			Msg("cannot run k6")
 
-	r.logger.Debug().Str("stdout", stdout.String()).Str("stderr", stderr.String()).Dur("duration", duration).Msg("ran k6 script")
-	r.logger.Info().Dur("duration", duration).Msg("ran k6 script")
+		dumpK6OutputStream(r.logger, zerolog.ErrorLevel, &stdout, "stream", "stdout")
+		dumpK6OutputStream(r.logger, zerolog.ErrorLevel, &stderr, "stream", "stderr")
 
-	var result RunResponse
+		logs, _ := afs.ReadFile(logsFn)
+		dumpK6OutputStream(r.logger, zerolog.InfoLevel, bytes.NewReader(logs), "stream", "logs")
```
ka3de, line 142: _nit:_ I understand these dumps are just for our own observability on the k6 output when the test is errored by a non-user / non-k6 error, right? Would it make sense to also log the possible error on reading the logs file and avoid the extra call to `dumpK6OutputStream` in that case?

I guess that is rare and would have low impact, but not sure if we could miss it at some point. It does feel a little bit cumbersome so up to you. Could be something like:
```suggestion
		if logs, err := afs.ReadFile(logsFn); err == nil {
			dumpK6OutputStream(r.logger, zerolog.InfoLevel, bytes.NewReader(logs), "stream", "logs")
		} else {
			logger.Error().Err(err).Msg("error reading k6 logs output")
		}
```

ka3de (approved):
LGTM. Just added a couple of _nit_ comments.

ka3de (approved):

PR #1074, opened 2024-11-27
Fix: trigger argo release workflow from GHA

mem:
We have the drone pipeline running in parallel with the GHA one, with the exception of the release step. In order to unblock this, disable triggering the argo workflow in Drone and enable it in GHA. I'm 100% sure this is going to fail without changes on the workflow side, but making those changes first will very likely break the drone pipeline. Start it here, with dev releases, so that we can figure out what else needs changing but keep the prod path enabled in Drone, in case we need to make a release before this is done.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -274,19 +274,18 @@ jobs:
         with:
           go-version-file: ./go.mod
 
-    # Commented out until we know it's working.
-    # - name: Trigger argo workflow
-    #   uses: grafana/shared-workflows/actions/trigger-argo-workflow@7d18a46aafb8b875ed76a0bc98852d74b91e7f91 # v1.0.0
-    #   with:
-    #     namespace: synthetic-monitoring-cd
-    #     workflow_template: deploy-${{ needs.preflight.outputs.repo_name }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_publish.yaml
```diff
@@ -274,19 +274,18 @@ jobs:
...
-    #       commit_author=${{ github.actor }}
-    #       commit_link=${{ github.server_url }}/${{ github.repository }}/commit/${{ github.sha }}
+      - name: Trigger argo workflow
+        uses: grafana/shared-workflows/actions/trigger-argo-workflow@7d18a46aafb8b875ed76a0bc98852d74b91e7f91 # v1.0.0
+        with:
+          namespace: synthetic-monitoring-cd
+          workflow_template: deploy-${{ needs.preflight.outputs.repo_name }}
+          extra_args: "--generate-name deploy-${{ needs.preflight.outputs.repo_name }}-"
+          parameters: |
+            mode=${{ (inputs.mode == 'prod' && 'release') || (inputs.mode == 'dev' && 'dev') }}
```
roobre, line 284: I'm not sure if this trick is genius or nasty but I'm definitely adding it to my book.

roobre (approved):
Assuming the argo workflow is now accepting the new parameters, everything looks good!

PR #1092, opened 2024-12-10
Fix: increase SHA-1 short version length

mem:
It's been observed that two different runs of git produce two different values of the short SHA-1 value (which is weird, because this is supposed to be based on the number of objects in the repository). Fix this at 12, because that is longer than what git automatically decides to use.


ckbedwell (approved):
LGTM

PR #1115, opened 2024-12-16
prober/ping: replace deprecated go-kit imports and go-ping dependency

roobre:
Closes https://github.com/grafana/synthetic-monitoring-agent/pull/1090


mem:
I think at least half of this is no longer required, is that right?

nadiamoe:
Probably so, closing for now.

PR #1166, opened 2025-01-28
fix: tag images with the bare version.

d0ugal:
At one point we tagged images with the bare version. For example [v0.29.10](https://hub.docker.com/layers/grafana/synthetic-monitoring-agent/v0.29.10/images/sha256-ddad54c24c4a22cb18169d3e8a52e5a38bf8ac94d21c7e0276f87c5c25fc3372). In the move from drone to github actions this has changed. The latest release has the version [v0.32.0-0-gc17d85830691](https://hub.docker.com/layers/grafana/synthetic-monitoring-agent/v0.32.0-0-gc17d85830691/images/sha256-b55df3b4394dd44a9ae7f28163fb475c922a7f94ae2833cc03641294dde24c8c).

Tools like renovate ignore these tags and thus my images have not been updated since v0.29.10. This PR adds the bare version back, which should make it easier for users to keep up with the latest.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -38,6 +38,7 @@ jobs:
       image: ghcr.io/grafana/grafana-build-tools:v0.37.0@sha256:64c6ff07fce6f93ba3732d4617b7ea3a370d7e87fd4e4cb92fd4faa25364cd50
     outputs:
       version: ${{ steps.version.outputs.value }}
+      version_bare: ${{ steps.version.outputs.bare_value }}
 
     steps:
       - name: checkout
@@ -73,6 +74,7 @@ jobs:
         run: |
           make version
           echo "value=$(cat dist/version)" >> "$GITHUB_OUTPUT"
+          echo "bare_value=$(cat dist/version | cut -d- -f1)" >> "$GITHUB_OUTPUT"
 
       - name: build
         run: make build-native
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_publish.yaml
```diff
@@ -73,6 +74,7 @@ jobs:
         run: |
           make version
           echo "value=$(cat dist/version)" >> "$GITHUB_OUTPUT"
+          echo "bare_value=$(cat dist/version | cut -d- -f1)" >> "$GITHUB_OUTPUT"
```
nadiamoe, line 77: I think this might be a problem if we ever do a pre-release, that is, a tag that itself contains `-`. A way to sidestep this would be to modify the `version` script, e.g. adding handling for a `--short` argument so it doesn't pass `--long` to `git` if `--short` is passed.

That being said, us creating tags with dashes sounds a bit unlikely so unless that bothers you, we can move forward.

nadiamoe (approved):

.github/workflows/build_and_publish.yaml
d0ugal: I went backwards and forwards on how to get the version without the extras. Eventually I settled on this after seeing it used here; https://github.com/grafana/synthetic-monitoring-agent/blob/main/scripts/release#L33

So I think we will have issues in other places if versions include a hyphen. However, that is probably a good reason to capture this logic (like you said) and put it in a in a script and reuse it.

PR #1179, opened 2025-02-04
Pass secret store from API

d0ugal:
This change uses the new protobuf fields from https://github.com/grafana/synthetic-monitoring-agent/pull/1192 and passes them down to the k6 executable.

This change requires quite a bit of plumbing to get the details in the right place.

It also includes a small refactor of local.go to make it easier to test the construction of the CLI args.


d0ugal:
This change depends on https://github.com/grafana/gsm-api-go-client/pull/1/files

mem (commented):
I don't see any show stoppers, but I'm still going thru this.

mem (approved):

PR #1234, opened 2025-03-06
feat: Add the sm-k6-gsm binary for secrets

d0ugal:
~~This depends on https://github.com/grafana/xk6-sm/pull/75/ and therefore the release: https://github.com/grafana/xk6-sm/pull/74~~

To support secrets in the short term we need to add a second k6 binary which tracks the upstream secrets work. This means it is an unreleased k6 version.

This binary will only be used when secret source configuration is provided from the API. This will start to happen with https://github.com/grafana/synthetic-monitoring-agent/pull/1179

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ ARG HOST_DIST=$TARGETOS-$TARGETARCH
 RUN adduser -D -u 12345 -g 12345 sm
 
 ADD --chown=sm:sm --chmod=0500 https://github.com/grafana/xk6-sm/releases/download/v0.0.3-pre/sm-k6-${TARGETOS}-${TARGETARCH} /usr/local/bin/sm-k6
+ADD --chown=sm:sm --chmod=0500 https://github.com/grafana/xk6-sm/releases/download/v0.4.0/sm-k6-${TARGETOS}-${TARGETARCH}-gsm /usr/local/bin/sm-k6-gsm
 COPY --chown=sm:sm --chmod=0500 --from=setcapper /usr/local/bin/synthetic-monitoring-agent /usr/local/bin/synthetic-monitoring-agent
 COPY --chown=sm:sm scripts/pre-stop.sh /usr/local/lib/synthetic-monitoring-agent/pre-stop.sh
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -24,6 +24,7 @@ ARG HOST_DIST=$TARGETOS-$TARGETARCH
 RUN adduser -D -u 12345 -g 12345 sm
 
 ADD --chown=sm:sm --chmod=0500 https://github.com/grafana/xk6-sm/releases/download/v0.0.3-pre/sm-k6-${TARGETOS}-${TARGETARCH} /usr/local/bin/sm-k6
+ADD --chown=sm:sm --chmod=0500 https://github.com/grafana/xk6-sm/releases/download/v0.4.0/sm-k6-${TARGETOS}-${TARGETARCH}-gsm /usr/local/bin/sm-k6-gsm
```
d0ugal, line 27: Upgrading this past the above binary is intentional. These will be out of sync for a while until the default binary is updated and then we can drop the second one.

scripts/make/sm-k6.mk
```diff
@@ -11,11 +11,17 @@ $(DISTDIR)/$(1)-$(2)/sm-k6:
...
 	chmod +x "$$@"
 
-sm-k6: $(DISTDIR)/$(1)-$(2)/sm-k6
+$(DISTDIR)/$(1)-$(2)/sm-k6-gsm:
+	mkdir -p "$(DISTDIR)/$(1)-$(2)"
+	# Renovate updates the following line. Keep its syntax as it is.
+	curl -sSL https://github.com/grafana/xk6-sm/releases/download/v0.4.0/sm-k6-$(1)-$(2)-gsm -o "$$@"
+	chmod +x "$$@"
+
+sm-k6: $(DISTDIR)/$(1)-$(2)/sm-k6 $(DISTDIR)/$(1)-$(2)/sm-k6-gsm
```
nadiamoe, line 20: In practice, this will only be used by the pipeline that packages the agent for Debian/Fedora (and derivatives). Container images rely on the `ADD` instruction (which you added).

However, since you haven't added that file to nfpm: https://github.com/grafana/synthetic-monitoring-agent/blob/ad3ffbbecf44565f74aefedc5dd3a91b6d138ad2/scripts/package/nfpm.yaml.template#L23

This means this is effectively ignored. If we do not intend to ship this in linux packages, I'd suggest getting rid of the make-related changes. If we do want to ship this binary in linux packages, then we should edit the nfpm config.

d0ugal: Good spot, thanks. Adding this to images only is probably enough for now.

d0ugal: Although, that could become problematic so its probably better to add it everywhere 🤔

d0ugal: Updated, hopefully I understood correctly

Dockerfile
nadiamoe: Just to make sure this is intended, renovate will update *both* URLs:
https://github.com/grafana/synthetic-monitoring-agent/blob/ad3ffbbecf44565f74aefedc5dd3a91b6d138ad2/.github/renovate.json5#L32

nadiamoe (approved):
LGTM!

PR #1267, opened 2025-03-25
fix: Pass Tenant ID to adhoc checks

mem:
It seems ad hoc checks have never used the tenant ID. The secrets handling code needs this information in order to be able to retrieve secrets.


The-9880 (approved):

d0ugal (approved):
Good spot!

PR #1274, opened 2025-04-01
fix: Remove second k6 binary

d0ugal:
Now that we have a more recent k6 version in xk6-sm[1] we can consolidate

[1]: https://github.com/grafana/xk6-sm/pull/106

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,7 @@ ARG HOST_DIST=$TARGETOS-$TARGETARCH
 
 RUN adduser -D -u 12345 -g 12345 sm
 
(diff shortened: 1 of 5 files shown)


nadiamoe (approved):

Dockerfile
```diff
@@ -24,7 +24,6 @@ ARG HOST_DIST=$TARGETOS-$TARGETARCH
 RUN adduser -D -u 12345 -g 12345 sm
 
 ADD --chown=sm:sm --chmod=0500 https://github.com/grafana/xk6-sm/releases/download/v0.4.0/sm-k6-${TARGETOS}-${TARGETARCH} /usr/local/bin/sm-k6
```
nadiamoe, line 26: The `xk6-sm` version that ships k6 v0.58.0 is not yet released, right? It might make sense to release that and update the version together with this PR.

PR #1281, opened 2025-04-04
fix: Change how the secret store is handled

mem:
In the current version of the code, the secret store is populated when the prober is created. The problem with that is that the prober is created *once* in the lifetime of the check, but the secret store needs to be updated more or less on a continuous basis (because it contains the token used to access the secrets).

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -43,6 +43,19 @@ import (
 const (
 	exitFail             = 1
 	defTelemetryTimeSpan = 5 // min
+
+	// The expiration time for access tokens is 10 minutes. Until we get
+	// expiration information from the API, set this to a low number.
+	// Normally it should be 15 minutes.
+	//
+	// This specific number is accounting for ~ 3 minutes of total
+	// execution time for scripts, plus 30 seconds of buffer for retrieving
+	// secrets, sending script to runner, etc. We expect secrets to be
+	// loaded near the start of the script, but given it's a program we
+	// cannot enforce that. k6 would have enforced that by making secrets
+	// part of the options block, but we do not want to go anywhere near
+	// that.
+	secretsTimeout = 450 * time.Second // 7.5 minutes
 )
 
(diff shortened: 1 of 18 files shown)


internal/tenants/manager.go
```diff
@@ -15,10 +16,15 @@ type Manager struct {
 	timeout       time.Duration
 	tenantsMutex  sync.Mutex
 	tenants       map[int64]*tenantInfo
+	logger        zerolog.Logger
 }
 
 var _ pusher.TenantProvider = &Manager{}
 
+const (
+	secretsTimeout = 8 * time.Minute
```
d0ugal, line 25: Why 8 minutes?

mem: > Why 8 minutes? I think it should be 7 if its 10 minus [maxScriptedTimeout](https://github.com/grafana/synthetic-monitoring-agent/blob/main/pkg/pb/synthetic_monitoring/checks_extra.go?rgh-link-date=2025-04-04T13%3A01%3A14Z#L151).

ah, good point. I remember you mentioned this, but I forgot why you said it should be less.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -43,6 +43,19 @@ import (
 const (
 	exitFail             = 1
 	defTelemetryTimeSpan = 5 // min
+
+	// The expiration time for access tokens is 10 minutes. Until we get
+	// expiration information from the API, set this to a low number.
+	// Normally it should be 15 minutes.
+	//
+	// This specific number is accounting for ~ 3 minutes of total
+	// execution time for scripts, plus 30 seconds of buffer for retrieving
```
d0ugal, line 52: Wouldn't that be 6.5 minutes?

d0ugal (approved):

cmd/synthetic-monitoring-agent/main.go
mem: argh, I meant ~ 2 minutes, which is I think what we talked about.

PR #1282, opened 2025-04-04
fix: Bust cache on expiry minus timeout

d0ugal:
This is https://github.com/grafana/synthetic-monitoring-agent/pull/1279 but it also subtracts the max timeout, so that we don't keep trying to use a token beyond its expiry.

### DIFF
--- a/internal/tenants/manager.go
+++ b/internal/tenants/manager.go
@@ -64,18 +64,36 @@ func (tm *Manager) run(ctx context.Context) {
 // calculateValidUntil determines the expiration time for a tenant based on the timeout
 // and the secret store expiration date (if set), returning the earlier of the two.
 func (tm *Manager) calculateValidUntil(tenant *sm.Tenant) time.Time {
-	validUntil := time.Now().Add(tm.timeout)
+	now := time.Now()
 
 	if tenant.SecretStore != nil && tenant.SecretStore.Expiry > 0 {
 		seconds, nanonseconds := math.Modf(tenant.SecretStore.Expiry)
+		// Subtract MaxScriptedTimeout to ensure the token is valid for the maximum running time.
 		expirationTime := time.Unix(int64(seconds), int64(nanonseconds*1e9))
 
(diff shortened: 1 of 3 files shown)


internal/tenants/manager.go
```diff
@@ -57,6 +60,20 @@ func (tm *Manager) run(ctx context.Context) {
 	}
 }
 
+// calculateValidUntil determines the expiration time for a tenant based on the timeout
+// and the secret store expiration date (if set), returning the earlier of the two.
+func (tm *Manager) calculateValidUntil(tenant *sm.Tenant) time.Time {
+	validUntil := time.Now().Add(tm.timeout)
+	if tenant.SecretStore != nil && tenant.SecretStore.Expiry > 0 {
+		// Subtract MaxScriptedTimeout to ensure the token is valid for the maximum running time
+		expirationTime := time.Unix(0, int64(tenant.SecretStore.Expiry*1e9)).Add(-sm.MaxScriptedTimeout)
```
d0ugal, line 69: @mem If we have (for example) 2 minutes left on the token and the cache is invalidated because we subtract the MaxScriptedTimeout. Will the API generate a new token or return the existing token because as far as it is concerned it is still valid?

mem: Right now it's generating a new token. It should probably do something smarter than that, because it's going to get 20+ requests from different agents, and it's going to generate a new token for each of them (that would be OK). We should probably try to avoid doing that.

d0ugal: Thanks. I think this change makes sense then. Otherwise, iiuc, If you have a check that takes 1 minute and runs every minute (and uses secrets) it will fail 10% of the time.

mem (approved):

internal/tenants/manager.go
mem: See https://github.com/grafana/synthetic-monitoring-api/pull/1311/files#diff-3ee267a18c51d1eb270867b59d9b9eb318d16f328ce9908ae869335bce91e017R176

Issue #1362, opened 2025-07-02
Binary tarballs are missing from latest releases


nadiamoe:
Latest releases only include `.deb` and `.rpm` packages, but there are no tarballs (or zip files) containing the bare binaries.

As a workaround, binaries can be extracted from deb/rpm packages and will work fine as they are statically compiled.

ChrisDuncan-personal:
> Hey [@ChrisDuncan-personal](https://github.com/ChrisDuncan-personal), we're working on it. Hopefully in a couple weeks we'll have linux-amd64 and linux-arm64 tarballs again, for future releases too.
> 
> We're tentatively thinking for going to binaries inside a `.tar.gz` file instead of a `.zip` file, will that work for you?

I tried to use Ansible's unarchive module on the .tar.gz file, and it uncompresses the folder for me, but the contents are not the same as what I would get from the "Linuxx86_64.zip" files. Not having the "Linuxx86_64.zip", will force me me have to create additional steps in ansible for my installation.

nadiamoe:
@ChrisDuncan-personal Please take a look a the latest release, which should include a `tar.gz` file with the required binaries in place. The layout of the files inside is equivalent to what the zip files had.

PR #1363, opened 2025-07-02
fix: build a .tar.gz file with bare binaries and attach it to releases

nadiamoe:
Works locally, but probably needs iteration if/when it fails for releases.

Fixes https://github.com/grafana/synthetic-monitoring-agent/issues/1362

### DIFF
--- a/scripts/make/950_release.mk
+++ b/scripts/make/950_release.mk
@@ -42,10 +42,10 @@ package-rpm-$(1)-$(2) : $(DISTDIR)/$(1)-$(2)/nfpm.yaml $(DISTDIR)/changelog.yaml
 		--packager rpm \
 		--target $(DISTDIR)/$(1)-$(2)/
 
-package-tgz-$(1)-$(2) : $(DISTDIR)/$(1)-$(2)/sm-k6 $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent $(ROOTDIR)/CHANGELOG.md $(ROOTDIR)/README.md
+package-tgz-$(1)-$(2) : $(DISTDIR)/$(1)-$(2)/sm-k6 $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent $(ROOTDIR)/CHANGELOG.md $(ROOTDIR)/README.md $(ROOTDIR)/LICENSE
 	# Create a tarball including the binaries, changelog, and readme. --transform is used to place all files at the root
 	# of the tarball. The built-in make variable dollar-caret is escaped with two dollars so it survives package_template.
-	$(S) tar --transform 's|.*/||' -zcf $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent.tar.gz $$^
+	$(S) tar --transform 's|.*/||' -zcf $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent-$(firstword $(subst -, ,$(BUILD_VERSION)))-$(1)-$(2).tar.gz $$^
 
 package-$(1)-$(2) : package-deb-$(1)-$(2) package-rpm-$(1)-$(2) package-tgz-$(1)-$(2)
 	@true


scripts/make/950_release.mk
```diff
@@ -42,7 +42,12 @@ package-rpm-$(1)-$(2) : $(DISTDIR)/$(1)-$(2)/nfpm.yaml $(DISTDIR)/changelog.yaml
 		--packager rpm \
 		--target $(DISTDIR)/$(1)-$(2)/
 
-package-$(1)-$(2) : package-deb-$(1)-$(2) package-rpm-$(1)-$(2)
+package-tgz-$(1)-$(2) : $(DISTDIR)/$(1)-$(2)/sm-k6 $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent $(ROOTDIR)/CHANGELOG.md $(ROOTDIR)/README.md
+	# Create a tarball including the binaries, changelog, and readme. --transform is used to place all files at the root
+	# of the tarball. The built-in make variable dollar-caret is escaped with two dollars so it survives package_template.
+	$(S) tar --transform 's|.*/||' -zcf $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent.tar.gz $$^
```
nadiamoe: I'm pretty sure `--transform` is gnu tar specific, but that seemed better than stage files to a temporary directory.

mem: > place all files at the root of the tarball

are you sure that is what goreleaser was doing?

> The built-in make variable dollar-caret is escaped with two dollars so it survives package_template.

🤷🏽 

I get that this might be surprising to you, but the comment is akin to:

```
// ^ is the XOR operator in Go
zero := 1^1
```

(the only reason why I'm pointing this out is because the comment made me look harder at this, thinking that something might be wrong about it; see lines 22, 31 and possibly others)

More importantly:

> $(S) tar --transform 's|.*/||' -zcf $(DISTDIR)/$(1)-$(2)/synthetic-monitoring-agent.tar.gz $$^

I'm fairly sure nfpm is creating files with versions embedded in them, and goreleaser creates archives with versions embedded in their filenames, too. IOW, it should be synthetic-monitoring-agent-0.1.2.tar.gz instead of synthetic-monitoring-agent.tar.gz. `$(BUILD_VERSION)` contains the version being built. The bit I don't recall is whether goreleaser uses the `v` prefix or not. I want to guess not.

nadiamoe: > are you sure that is what goreleaser was doing?

I am not treating this as a "replicate exactly the previous behavior using a different implementation", but rather as a "provide equivalent functionality with the tools we use now".

I get pissed myself when tarballs come with a `dist` or similarly name inconsequential folder that is obviously a byproduct of the specifics of the build system used and not useful at all to me as a user, so I try to avoid doing that to others.

I also think the vast majority of linux operators (this being a linux-only release) prefer to work with tarballs than with zip files [^1].

FWIW, this was the structure for the latest release where the zip file was included:

```console
$ curl -sSL https://github.com/grafana/synthetic-monitoring-agent/releases/download/v0.26.0/synthetic-monitoring-agent_0.26.0_linux_amd64.zip | bsdtar -t
CHANGELOG.md
LICENSE
README.md
sm-k6/k6
synthetic-monitoring-agent
```

I see the change from `sm-k6/k6` as a fix as we now look for a binary called `sm-k6`, not `k6`. As for the license and the version in the filename, I can definitely fix those :)

[^1]:  As an anecdote, for the example above I had to look how on earth do you the `tar -t` equivalent for zip files, only to find that `unzip` doesn't really do that and you need `bsdtar`.

nadiamoe: After the latest commit, the tarball looks like and contains this:

```console
$ tar -tzf dist/linux-amd64/synthetic-monitoring-agent-v0.38.6-linux-amd64.tar.gz
sm-k6
synthetic-monitoring-agent
CHANGELOG.md
README.md
LICENSE
```

PR #1366, opened 2025-07-02
feat: Implement usage report client

Pokom:
Implement usage report client that can send reports to https://stats.grafana.net/. See https://github.com/grafana/usage-stats/ and https://github.com/grafana/usage-stats-handler/ for backend details about the stats endpoint.

Updates `checks.go` to submit a report whenever a probe is registered or updated. The following metadata is sent over
- Version
- Arch
- OS
- Features
- Public bool
- TenantID

These are the fields that implemented as columns in usage stats(https://github.com/grafana/usage-stats/pull/55). Any other fields that we add to the Report struct will be included as raw json and stored in a `report` field in BigQuery.

Added a flag `disable-usage-reports` which is initially set to true. This provides a buffer so that we can announce that Grafana Labs will start to collect usage statistics for Agent's _and_ ensure documentation is clearly updated to reflect that usage reports will be gathered.

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -77,15 +77,14 @@ func run(args []string, stdout io.Writer) error {
 			DisableK6            bool
 			DisableUsageReports  bool
 		}{
-			GrpcApiServerAddr:   "localhost:4031",
-			HttpListenAddr:      "localhost:4050",
-			K6URI:               "sm-k6",
-			K6BlacklistedIP:     "10.0.0.0/8",
-			SelectedPublisher:   pusherV2.Name,
-			TelemetryTimeSpan:   defTelemetryTimeSpan,
-			AutoMemLimit:        true,
-			MemLimitRatio:       0.9,
-			DisableUsageReports: true,
+			GrpcApiServerAddr: "localhost:4031",
+			HttpListenAddr:    "localhost:4050",
+			K6URI:             "sm-k6",
+			K6BlacklistedIP:   "10.0.0.0/8",
+			SelectedPublisher: pusherV2.Name,
+			TelemetryTimeSpan: defTelemetryTimeSpan,
+			AutoMemLimit:      true,
+			MemLimitRatio:     0.9,
 		}
 	)
 
@@ -106,7 +105,7 @@ func run(args []string, stdout io.Writer) error {
 	flags.IntVar(&config.TelemetryTimeSpan, "telemetry-time-span", config.TelemetryTimeSpan, "time span between telemetry push executions per tenant")
 	flags.BoolVar(&config.AutoMemLimit, "enable-auto-memlimit", config.AutoMemLimit, "automatically set GOMEMLIMIT")
 	flags.BoolVar(&config.DisableK6, "disable-k6", config.DisableK6, "disables running k6 checks on this probe")
-	flags.BoolVar(&config.DisableUsageReports, "disable-usage-reports", config.DisableUsageReports, "disables sending Usage Reports to Grafana")
+	flags.BoolVar(&config.DisableUsageReports, "disable-usage-reports", config.DisableUsageReports, "Disable anonymous usage reports")
 	flags.Float64Var(&config.MemLimitRatio, "memlimit-ratio", config.MemLimitRatio, "fraction of available memory to use")
 	flags.Var(&features, "features", "optional feature flags")
 
@@ -213,17 +212,10 @@ func run(args []string, stdout io.Writer) error {
 		zl.Info().Msg("disabling k6 checks")
 	}
 
-	var usageReporter *usage.UsageReporter
-
+	usageReporter := usage.NewNoOPReporter()
 	if !config.DisableUsageReports {
-		zl.Info().Str("usage endpoint", usage.DefaultUsageStatsEndpoint).Msg("Starting usage stats reporting")
-		endpoint := usage.DefaultUsageStatsEndpoint
-		if config.DevMode {
-			// TODO: Make this point to staging
-			endpoint = "http://localhost:4200"
-		}
-		usageReporter = usage.NewUsageReporter(endpoint, features)
-		zl.Info().Msg("Usage stats reporting")
+		usageReporter = usage.NewHTTPReporter(usage.DefaultUsageStatsEndpoint, features)
+		zl.Info().Msg("enabled collecting anonymous usage reports.")
 	}
 
 	promRegisterer := prometheus.NewRegistry()
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -36,6 +36,7 @@ import (
 	"github.com/grafana/synthetic-monitoring-agent/internal/secrets"
 	"github.com/grafana/synthetic-monitoring-agent/internal/telemetry"
 	"github.com/grafana/synthetic-monitoring-agent/internal/tenants"
+	"github.com/grafana/synthetic-monitoring-agent/internal/usage"
 	"github.com/grafana/synthetic-monitoring-agent/internal/version"
 	"github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
 )
@@ -74,15 +75,17 @@ func run(args []string, stdout io.Writer) error {
 			AutoMemLimit         bool
 			MemLimitRatio        float64
 			DisableK6            bool
+			DisableUsageReports  bool
 		}{
-			GrpcApiServerAddr: "localhost:4031",
-			HttpListenAddr:    "localhost:4050",
-			K6URI:             "sm-k6",
-			K6BlacklistedIP:   "10.0.0.0/8",
-			SelectedPublisher: pusherV2.Name,
-			TelemetryTimeSpan: defTelemetryTimeSpan,
-			AutoMemLimit:      true,
-			MemLimitRatio:     0.9,
+			GrpcApiServerAddr:   "localhost:4031",
+			HttpListenAddr:      "localhost:4050",
+			K6URI:               "sm-k6",
+			K6BlacklistedIP:     "10.0.0.0/8",
+			SelectedPublisher:   pusherV2.Name,
+			TelemetryTimeSpan:   defTelemetryTimeSpan,
+			AutoMemLimit:        true,
+			MemLimitRatio:       0.9,
+			DisableUsageReports: true,
 		}
 	)
 
@@ -103,6 +106,7 @@ func run(args []string, stdout io.Writer) error {
 	flags.IntVar(&config.TelemetryTimeSpan, "telemetry-time-span", config.TelemetryTimeSpan, "time span between telemetry push executions per tenant")
 	flags.BoolVar(&config.AutoMemLimit, "enable-auto-memlimit", config.AutoMemLimit, "automatically set GOMEMLIMIT")
 	flags.BoolVar(&config.DisableK6, "disable-k6", config.DisableK6, "disables running k6 checks on this probe")
+	flags.BoolVar(&config.DisableUsageReports, "disable-usage-reports", config.DisableUsageReports, "disables sending Usage Reports to Grafana")
 	flags.Float64Var(&config.MemLimitRatio, "memlimit-ratio", config.MemLimitRatio, "fraction of available memory to use")
 	flags.Var(&features, "features", "optional feature flags")
 
@@ -207,6 +211,19 @@ func run(args []string, stdout io.Writer) error {
 		zl.Info().Str("k6URI", config.K6URI).Msg("enabling k6 checks")
 	} else {
 		zl.Info().Msg("disabling k6 checks")
+	}
+
+	var usageReporter *usage.UsageReporter
+
+	if !config.DisableUsageReports {
+		zl.Info().Str("usage endpoint", usage.DefaultUsageStatsEndpoint).Msg("Starting usage stats reporting")
+		endpoint := usage.DefaultUsageStatsEndpoint
+		if config.DevMode {
+			// TODO: Make this point to staging
+			endpoint = "http://localhost:4200"
+		}
+		usageReporter = usage.NewUsageReporter(endpoint, features)
+		zl.Info().Msg("Usage stats reporting")
 	}
 
 	promRegisterer := prometheus.NewRegistry()
@@ -319,6 +336,7 @@ func run(args []string, stdout io.Writer) error {
 		TenantLimits:   limits,
 		TenantSecrets:  secrets,
 		Telemeter:      telemetry,
+		UsageReporter:  usageReporter,
 	})
 	if err != nil {
 		return fmt.Errorf("Cannot create checks updater: %w", err)
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -76,7 +76,6 @@ func run(args []string, stdout io.Writer) error {
 			MemLimitRatio        float64
 			DisableK6            bool
 			DisableUsageReports  bool
-			UsageStatsEndpoint   string
 		}{
 			GrpcApiServerAddr: "localhost:4031",
 			HttpListenAddr:    "localhost:4050",
@@ -107,7 +106,6 @@ func run(args []string, stdout io.Writer) error {
 	flags.BoolVar(&config.AutoMemLimit, "enable-auto-memlimit", config.AutoMemLimit, "automatically set GOMEMLIMIT")
 	flags.BoolVar(&config.DisableK6, "disable-k6", config.DisableK6, "disables running k6 checks on this probe")
 	flags.BoolVar(&config.DisableUsageReports, "disable-usage-reports", config.DisableUsageReports, "Disable anonymous usage reports")
-	flags.StringVar(&config.UsageStatsEndpoint, "usage-stats-endpoint", usage.DefaultUsageStatsEndpoint, "usage stats endpoint to send reports to")
 	flags.Float64Var(&config.MemLimitRatio, "memlimit-ratio", config.MemLimitRatio, "fraction of available memory to use")
 	flags.Var(&features, "features", "optional feature flags")
 
@@ -214,10 +212,13 @@ func run(args []string, stdout io.Writer) error {
 		zl.Info().Msg("disabling k6 checks")
 	}
 
-	usageReporter := usage.NewNoOPReporter()
-	if !config.DisableUsageReports {
-		usageReporter = usage.NewHTTPReporter(config.UsageStatsEndpoint, features)
-		zl.Info().Msg("enabled collecting anonymous usage reports.")
+	var usageReporter usage.Reporter
+	switch {
+	case config.DisableUsageReports:
+		usageReporter = usage.NewNoOPReporter()
+	default:
+		usageReporter = usage.NewHTTPReporter(usage.ProdStatsEndpoint)
+		zl.Info().Msg("enabled collecting anonymous usage reports. You can disable collection by setting -disable-usage-reports.")
 	}
 
 	promRegisterer := prometheus.NewRegistry()
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -76,6 +76,7 @@ func run(args []string, stdout io.Writer) error {
 			MemLimitRatio        float64
 			DisableK6            bool
 			DisableUsageReports  bool
+			UsageStatsEndpoint   string
 		}{
 			GrpcApiServerAddr: "localhost:4031",
 			HttpListenAddr:    "localhost:4050",
@@ -106,6 +107,7 @@ func run(args []string, stdout io.Writer) error {
 	flags.BoolVar(&config.AutoMemLimit, "enable-auto-memlimit", config.AutoMemLimit, "automatically set GOMEMLIMIT")
 	flags.BoolVar(&config.DisableK6, "disable-k6", config.DisableK6, "disables running k6 checks on this probe")
 	flags.BoolVar(&config.DisableUsageReports, "disable-usage-reports", config.DisableUsageReports, "Disable anonymous usage reports")
+	flags.StringVar(&config.UsageStatsEndpoint, "usage-stats-endpoint", usage.DefaultUsageStatsEndpoint, "usage stats endpoint to send reports to")
 	flags.Float64Var(&config.MemLimitRatio, "memlimit-ratio", config.MemLimitRatio, "fraction of available memory to use")
 	flags.Var(&features, "features", "optional feature flags")
 
@@ -214,7 +216,7 @@ func run(args []string, stdout io.Writer) error {
 
 	usageReporter := usage.NewNoOPReporter()
 	if !config.DisableUsageReports {
-		usageReporter = usage.NewHTTPReporter(usage.DefaultUsageStatsEndpoint, features)
+		usageReporter = usage.NewHTTPReporter(config.UsageStatsEndpoint, features)
 		zl.Info().Msg("enabled collecting anonymous usage reports.")
 	}
 
--- a/internal/checks/checks.go
+++ b/internal/checks/checks.go
@@ -388,7 +388,7 @@ func (c *Updater) loop(ctx context.Context) (bool, error) {
 
 	c.probe = &result.Probe
 
-	err = c.usageReporter.ReportProbe(ctx, result.Probe)
+	err = c.usageReporter.ReportProbe(ctx, result.Probe, c.features)
 	if err != nil {
 		c.logger.Warn().Err(err).Msg("reporting usage failed")
 	}
--- a/internal/checks/checks.go
+++ b/internal/checks/checks.go
@@ -388,7 +388,6 @@ func (c *Updater) loop(ctx context.Context) (bool, error) {
 
 	c.probe = &result.Probe
 
-	// usageReporter will be nil if the usage report configuration is disabled
 	err = c.usageReporter.ReportProbe(ctx, result.Probe)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


internal/usage/usage.go
```diff
@@ -0,0 +1,96 @@
...
+
+func NewUsageReporter(endpoint string, features feature.Collection) *UsageReporter {
+	if endpoint == "" {
+		endpoint = DefaultUsageStatsEndpoint
+	}
+	return &UsageReporter{
+		endpoint: endpoint,
+		features: features,
+	}
+}
```
Pokom: I'm not sure if passing in features when instantiating is appropriate given that the probe has _some_ concept of what features are enabled. The problem I found is that the string representation of `sm.Probe.Capabilities.String()` doesn't get rendered in a human readable format. This could be problematic if the probe's capabilities are altered after being registered.

internal/usage/usage.go
```diff
@@ -0,0 +1,96 @@
...
+// ReportProbe creates a Report from the probe and then sends the report to the stats api endpoint via the report method.
+func (r *UsageReporter) ReportProbe(ctx context.Context, probe sm.Probe) error {
+	report := &Report{
+		Report:       probe.String(),
+		CreatedAt:    time.Now().Format(time.RFC3339),
+		OS:           runtime.GOOS,
+		Arch:         runtime.GOARCH,
+		Version:      probe.Version,
+		UsageStatsId: uuid.NewString(),
+		Features:     r.features.String(),
```
Pokom: We could remove the need to have UsageReporter be aware of features by using `probe.Capabilities.String()`, but I need to look deeper and figure out how to get the output to be a bit more human readable.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -74,15 +75,17 @@ func run(args []string, stdout io.Writer) error {
...
-			MemLimitRatio:     0.9,
+			GrpcApiServerAddr:   "localhost:4031",
+			HttpListenAddr:      "localhost:4050",
+			K6URI:               "sm-k6",
+			K6BlacklistedIP:     "10.0.0.0/8",
+			SelectedPublisher:   pusherV2.Name,
+			TelemetryTimeSpan:   defTelemetryTimeSpan,
+			AutoMemLimit:        true,
+			MemLimitRatio:       0.9,
+			DisableUsageReports: true,
```
Pokom: Mostly formatting changes here. I've added `DisableUsageReports` since bool's zero value is false, and this ensures we don't start collecting usage reports before updating documentation.

ka3de: So do we expect to merge this disabled initially and then enable it by default when the docs are in place? Could we write the docs and sync its release along with the release of this feature with the flag enabled by default?

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -209,6 +213,19 @@ func run(args []string, stdout io.Writer) error {
...
 	}
 
+	var usageReporter *usage.UsageReporter
+
+	if !config.DisableUsageReports {
+		zl.Info().Str("usage endpoint", usage.DefaultUsageStatsEndpoint).Msg("Starting usage stats reporting")
+		endpoint := usage.DefaultUsageStatsEndpoint
+		if config.DevMode {
+			// TODO: Make this point to staging
+			endpoint = "http://localhost:4200"
```
ka3de: Does this mean there's a staging usage stats endpoint? I'm not sure whether we should make a distinction here, just to understand it better, what was your reasoning for this?

internal/checks/checks.go
```diff
@@ -382,6 +388,14 @@ func (c *Updater) loop(ctx context.Context) (bool, error) {
 
 	c.probe = &result.Probe
 
+	// usageReporter will be nil if the usage report configuration is disabled
+	if c.usageReporter != nil {
+		err = c.usageReporter.ReportProbe(context.Background(), result.Probe)
```
ka3de: Should we pass the `ctx` that is passed to the `loop()` method instead?

internal/checks/checks.go
```diff
@@ -382,6 +388,14 @@ func (c *Updater) loop(ctx context.Context) (bool, error) {
 
 	c.probe = &result.Probe
 
+	// usageReporter will be nil if the usage report configuration is disabled
+	if c.usageReporter != nil {
```
ka3de: As an alternative, not necessarily a suggestion, we could define an interface for `usage.UsageReporter` and provide two different implementations, one that pushes to Grafana usage stats, and one that is _noop_, which can be instantiated when the feature is disabled.

internal/usage/usage.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/google/uuid"
+
+	"github.com/grafana/synthetic-monitoring-agent/internal/feature"
+	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
+)
+
+// Report represents a specific usage event that will be sent to https://stats.grafana.com to be processed and stored.
+// Each attribute represents a column in a BigQuery table that can be easily searched.
+// Adding new attributes to Report will not automatically update the table, and instead needs to be handled in https://github.com/grafana/usage-stats
+type Report struct {
```
ka3de: This type seems to be only used within this package, should it be unexported?

internal/usage/usage.go
```diff
@@ -0,0 +1,96 @@
...
+	Report       string `json:"report"`
+	Version      string `json:"version"`
+	Public       bool   `json:"public"`
+	Features     string `json:"features"`
+	UsageStatsId string `json:"usageStatsId"`
+	TenantID     int64  `json:"tenantId"`
+}
+
+// UsageReporter represents
+type UsageReporter struct {
```
ka3de: Since the package name is `usage`, should this be named `Reporter` so we avoid the repetition on `usage.UsageReporter`?

internal/usage/usage.go
ka3de: Notice that the features set in the API, the ones that are returned when the probe registers (capabilities), will be the ones that dictate which checks can be sent to this probe. And the features that we are passing in this constructor are the ones that govern the agent behavior. So IIRC there must be a match between both configurations in order to run certain type of checks.

I think we have to review this better to understand which is the data that we are interested in.

internal/usage/usage.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// submitReport is responsible for sending a report to the stats endpoint via an http POST request. The primary concern is that
+// the http server responds with http.StatusOK. Otherwise, there are no other expected responses.
+func (r *UsageReporter) submitReport(ctx context.Context, report *Report) error {
+	jsonData, err := json.Marshal(&report)
+	if err != nil {
+		return fmt.Errorf("failed to marshal data: %w", err)
+	}
+	client := &http.Client{Timeout: 10 * time.Second}
```
ka3de: Should we instantiate an HTTP client at the `UsageReporter` level? Instead of creating it on each report submit (even if we expect this to be called just once)?

internal/usage/usage_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			features := feature.NewCollection()
+			for _, f := range tt.features {
+				_ = features.Set(f)
+			}
+			r := NewUsageReporter(tt.endpoint, features)
+			if err := r.ReportProbe(tt.ctx, sm.Probe{}); (err != nil) != tt.wantErr {
```
ka3de: nit: We could just pass `t.Context()` and remove the `ctx` field from the test cases.

internal/usage/usage_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/grafana/synthetic-monitoring-agent/internal/feature"
+	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
+)
+
+func TestUsageReporter_Report(t *testing.T) {
+	// Create a test http server that intercepts requests to https://stats.grafana.net
```
ka3de, line 17: nit:
```suggestion
        t.Parallel()
        
	// Create a test http server that intercepts requests to https://stats.grafana.net
```

ka3de (commented):
Thanks for working on this @Pokom ! This is gonna be very useful :raised_hands: 
Made a first quick pass, I will try to review this better tomorrow and we can catch up on the features discussion and the specific data to send.

cmd/synthetic-monitoring-agent/main.go
Pokom: I think an argument could be made to use staging for local development. Though I think for local development we should automatically disable reporting so we're not storing garbage usage reports. I think setting a different endpoint for `devMode` adds complexity and is probably best to either
1. enable reports and send them to production
2. disable report

internal/checks/checks.go
Pokom: I was considering what that option would look like, and personally think it's cleaner. Let me take a pass at implementing an interface and having a noop implementation 👍🏻

internal/usage/usage.go
Pokom: Likely not, good catch!

internal/usage/usage_test.go
Pokom: TIL, thanks!

internal/usage/usage.go
Pokom: I'm okay with that, I don't have a strong opinion either way. http.Clients are goroutine safe, so we should be okay.

internal/checks/checks.go
Pokom: Let me try it locally and just validate. I can't remember if I had a distinct reason why I didn't, but as long as it's working I don't see the need to create a new ctx.

cmd/synthetic-monitoring-agent/main.go
Pokom: My inclination is to remove the ability to set the endpoint until we have a good reason to do so. The usage reporter is either on and sending to stats.grafana.com, or it's off.

cmd/synthetic-monitoring-agent/main.go
Pokom: Given our recent conversation, let's coordinate writing the docs and sync the release of the docs with the feature.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -74,6 +75,7 @@ func run(args []string, stdout io.Writer) error {
 			AutoMemLimit         bool
 			MemLimitRatio        float64
 			DisableK6            bool
+			DisableUsageReports  bool
```
Pokom, line 78: This now defaults to false, which means the agent will collect usage reports by default.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -103,6 +106,8 @@ func run(args []string, stdout io.Writer) error {
 	flags.IntVar(&config.TelemetryTimeSpan, "telemetry-time-span", config.TelemetryTimeSpan, "time span between telemetry push executions per tenant")
 	flags.BoolVar(&config.AutoMemLimit, "enable-auto-memlimit", config.AutoMemLimit, "automatically set GOMEMLIMIT")
 	flags.BoolVar(&config.DisableK6, "disable-k6", config.DisableK6, "disables running k6 checks on this probe")
+	flags.BoolVar(&config.DisableUsageReports, "disable-usage-reports", config.DisableUsageReports, "Disable anonymous usage reports")
+	flags.StringVar(&config.UsageStatsEndpoint, "usage-stats-endpoint", usage.DefaultUsageStatsEndpoint, "usage stats endpoint to send reports to")
```
Pokom: I added this because I needed _some_ mechanism to override this for local development. I don't know if there's ever a time where someone out in the wild would override this, so maybe we can expose this as an environment variable that's hidden from the user?

internal/usage/usage.go
```diff
@@ -0,0 +1,114 @@
...
+
+const (
+	// UsageReportApplication aligns with the usage-stats service endpoint defined
+	// in github.com/grafana/usage-stats for synthetic monitoring agents
+	UsageStatsApplication = "synthetic-monitoring-agent-usage-report"
+	// Base Endpoint for usage stats
+	DefaultUsageStatsEndpoint = "https://stats.grafana.com"
+)
+
+func NewHTTPReporter(endpoint string, features feature.Collection) Reporter {
```
ka3de: nit: In Go is idiomatic to _[return types and accept interfaces](https://duncanleung.com/go-idiom-accept-interfaces-return-types/)_.
So should we return `*HTTPReporter` here? We can fix `main` by declaring `var usageReporter usage.Reporter` and initializing it based on a _switch_ on the enabled/disabled flag.

internal/usage/usage.go
```diff
@@ -0,0 +1,114 @@
...
+	if endpoint == "" {
+		endpoint = DefaultUsageStatsEndpoint
+	}
+	return &HTTPReporter{
+		endpoint: endpoint,
+		features: features,
```
ka3de: In relation to this [comment](https://github.com/grafana/synthetic-monitoring-agent/pull/1366#discussion_r2180859033).
Yes, I think passing the _features_ to the reporter constructor is "not ideal" as it's not a field of the reporter, it's a field that represents certain characteristics of the probe.

Based on that, should we remove this from the reporter itself and accept it as part of the `ReportProbe()` method?
Notice that, if I'm not wrong, the _features_ are also a field of the `Updater`, therefore we can pass them from `loop()` [here](https://github.com/grafana/synthetic-monitoring-agent/blob/1a2dbdc358cbde6fe7b6660de1bc1f583a13f205/internal/checks/checks.go#L391) as:
```
	err = c.usageReporter.ReportProbe(ctx, result.Probe, c.features)
```

internal/usage/usage_test.go
```diff
@@ -0,0 +1,47 @@
...
+			endpoint: ts.URL,
+			features: []string{},
+		},
+		"Send over a full list of features": {
+			endpoint: ts.URL,
+			features: []string{feature.K6, feature.Traceroute, feature.ExperimentalDnsProber},
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
```
ka3de, line 96: I think we might need to add something else to this test, because right now we are basically asserting that calling `ReportProbe()` does not return an error, but I understand our main goal is to assert that the configured endpoint is called with the data provided.

For that purpose, since the test cases are sequential, we could:
- Define a local var for `TestUsageReporter_Reporter` test function such as `var gotReport report`
- In the test server handler, read and deserialize the request body and assign it to `gotReport`. Any error during this process returns 500, which would make our current test impl. fail.
- Add an `expReport report` field to the test case struct definition.
- In the `t.Run()`, reset the `gotReport = report{}` at the start, and then assert that it matches the report we expect `require.Equal(t, tt.expReport, gotReport)`.

Let me know what do you think.

cmd/synthetic-monitoring-agent/main.go
ka3de: If there's no use case for end users, I wonder if it would be simpler to not have an option for this, and just overwrite it completely for local dev purposes, even if that sounds "ugly" also... But no strong opinion about it, whatever you choose LGTM. :+1:

Pokom: Realistically, the only time this needs to be changed is if we're making changes to the usage stats report. If that happens more then once a year, I'd be surprised. So yeah I agree, let's remove it and let the engineer forcefully change it in code. If this comes up _many_ times, we can reevaluate adding it

internal/usage/usage_test.go
Pokom: Yep, that sounds good to me. That's a much better test then what I started with 😅 I think I've got enough direction to add a meaningful test here

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -209,6 +212,15 @@ func run(args []string, stdout io.Writer) error {
 		zl.Info().Msg("disabling k6 checks")
 	}
 
+	var usageReporter usage.Reporter
+	switch {
+	case config.DisableUsageReports:
+		usageReporter = usage.NewNoOPReporter()
+	default:
+		usageReporter = usage.NewHTTPReporter(usage.ProdStatsEndpoint)
```
mem, line 220: ah, interesting...

Does this mean that the agents in dev clusters will be reporting to the prod instance of this?

internal/usage/usage.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ReportProbe creates a report from the probe and then sends the report to the stats api endpoint via the report method.
+func (hr *HTTPReporter) ReportProbe(ctx context.Context, probe sm.Probe, features feature.Collection) error {
+	r := &report{
+		Report:       probe.String(),
+		CreatedAt:    time.Now().Format(time.RFC3339),
+		OS:           runtime.GOOS,
+		Arch:         runtime.GOARCH,
+		Version:      probe.Version,
+		UsageStatsId: uuid.NewString(),
```
mem: double checking: each individual report gets a different ID?

internal/usage/usage.go
```diff
@@ -0,0 +1,109 @@
...
+		Features:     features.String(),
+		Public:       probe.Public,
+		TenantID:     probe.TenantId,
```
mem, line 99: I don't recall, but I suspect this number is going to be a global tenant ID. See `internal/model/model.go`.

For public probes, we extend the tenant ID so that it becomes globally unique across all service regions. Private probes don't see that, they get the actual ID of their tenant.

At startup, the agent gets a parameter (`GrpcApiServerAddr`) which should be reported here because that's the only way we can tell tenant 1 from region A appart from tenant 1 from region Z. The probes themselves don't have a need for this, but looking at this, I think we should add OrgId and StackId to the Probe message, so that we can include that in logs and here. That's a much larger change becuase it requires modifying the protobuf definition and updating the code in the API.

When we first starting working on this, this was not true, but nowadays there's a 1:1 mapping between stack ID and tenant ID. At some point we should migrate from one to the other, I think.

tmp/build-errors.log
mem, line 1: how did this end up here?

cmd/synthetic-monitoring-agent/main.go
Pokom: Yep that's the case. Databases and other projects do this as well, so we're not misaligned. I do understand how that fact would be surprising. There exists a concept of a staging environment for usage stats, but when looking at `deployment_tools`, it has not been updated in ages. Personally, I don't see shipping our dev probes usage reports to be problematic as they are installations that are running out in the wild and doing _something_. 

With that context, I see a few options:
1. update usage stats to have a reliable staging environment to ship data off too
2. disable usage reports in our development environments with the `-disable-usage-stats` flag
3. let usage data flow from dev clusters into prod usage stats

My personal preference is 3 and then 2.

internal/usage/usage.go
Pokom: That's a good catch, and yes. Would it make more sense to generate a uuid when the probe launches and pass this along to the usage reporters `SendReport` method? This ensures a the UsageStatsId is common across all reports send by a probe during it's lifetime.

tmp/build-errors.log
Pokom: 🤦🏻 I had run `air` in agent forgetting that there did not exist an `.air.toml` configuration. I'll clean this up

internal/usage/usage.go
Pokom: Ah interesting, and thanks for the context! I had made the assumption that tenantId is unique regionally, and that the combination of tenantId and region would give us some bit of uniqueness. 

Should I capture this context as a comment in the codebase? Or is there a larger change that you're suggested to make?

internal/usage/usage_test.go
ka3de: Only a _nit_ that we could add a `return` [here](https://github.com/grafana/synthetic-monitoring-agent/blob/80fc7d2d8ac979321b68899e053afc458a511b74/internal/usage/usage_test.go#L26). That does not change the result of the test, though.

internal/usage/usage.go
ka3de: Could we generate the ID based on an FNV hash from some of the probe properties sent from the API?
E.g.: `{id,tenantId, name, region, public?}` ? Updating some of these fields, for example if a user changes the _name_ or the _region_ would, to effect of the usage stats, be represented as different report. Not completely sure this would be a better fit for usage stats, or if there's a better alternative.
I'm avoiding just using the obvious _id_ and/or _tenantId_ because these are always `1` for public probes, see [impl](https://github.com/grafana/sm-proxy/blob/45ce1cf8fe117d9bff762347046b13337175a93f/internal/proxy/upstream.go#L251-L263). and [metrics](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22gfz%22%3A%7B%22datasource%22%3A%22000000134%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22sum+by+%28id%29%28sm_agent_info%29%22%2C%22range%22%3Atrue%2C%22instant%22%3Atrue%2C%22datasource%22%3A%7B%22type%22%3A%22prometheus%22%2C%22uid%22%3A%22000000134%22%7D%2C%22editorMode%22%3A%22code%22%2C%22legendFormat%22%3A%22__auto%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1).

Pokom: Oh that's a solid idea @ka3de. As long as the hashing algorithm is deterministic, then I think that's far better then generating a random UUID. It preserves anonymity while also providing the ability to track usage of a particular probe over time. Let me take a look at what Go stdlib provides so I'm not reinventing something.

Pokom: @mem + @ka3de can you take a peak at the last diff? I've implemented a hash method for the probe based upon the following attributes
- Id
- TenantId
- Name
- Region
- Public

It's a bit naive, but with tests I've demonstrated that the method is at least deterministic. Given this change, the UsageStatsId should stay the same for probes.

cmd/synthetic-monitoring-agent/main.go
mem: OK, that's fine. Just double checking this was not overlooked. Thanks.

internal/usage/usage.go
mem: The tenantId is indeed unique for a region, but since there's a single instance of the stats service, we will see the same tenantId across probes from multiple regions. Because of the way we manipulate the tenant ID for agents behind the proxy, you end up with a negative large number.

Maybe the answer to this is just to disable the report for public probes? Dev is OK for validation and testing purposes, but I'm not sure we would be collecting new information. Also consider that because individual agents behind a proxy think they are the same entity (they get the same name, the same tenant ID, the same probe ID, ...) we are going to end up with duplicate data from the different instances.

mem: And this is a good place as any to say this.

When defining new flags, we need to consider the flag that an unknown flag causes the program to not run.

What that means in practice is that in order to introduce a new flag, we have to do it in steps:

- Add the flag, with some reasonable default value, without an implementation
- Merge, release, so that all instances of the program in all clusters understand the flag
- Modify deployment tools to pass the correct value of the flag for our case (it's still doing nothing, the flag is defined, but the implementation is not there)
- Add the implementation, merge, release

In principle, you can use a different approach: modify deployment tools, the new pods would fail to start, deploy the new image that understands the flag, the deployments work. But that's highly incompatible with the way we do rollouts.

If the default value of the flag is useful in our case, you can add the implementation together with the flag. That's for example not the case with the `-k6-uri` flag in the agent. We need to use a different value.

I mention this because if we decide that we should disable the report in our instances, when we should add the flag in deployment tools, and that cannot happen if the flag is not defined.

internal/usage/usage.go
mem: I think that's fine.

Note that the result is going to be the same for all our agent instances in the same location because all those values are the same for all instances. I say this because of the inclusion of "public" in this set.

The result is _likely_ to be different for private probes from different tenants, but there's a chance all those values are the same for two instances of different tenants in different regions (prod-xxx vs prod-yyy). The _region_ that you get here is *not* related to prod-xxx. Because it's highly unlikely that all these values match, I think this is OK. It would theoretically clash if we set up a private probe of our own in each of the regions because the method we would use for that is more likely to land in that situation (e.g. we add multiple instances of a probe with the same name in a single location, and we end up with the same ID by chance, because when we do it just after setting something up).

The morale of the story is that we should replace those IDs with UUIDs, but that's outside of the scope of this discussion.

mem (approved):
LGTM

ka3de (dismissed):
LGTM :tada:

internal/usage/usage.go
Pokom: > Maybe the answer to this is just to disable the report for public probes?

That sounds pretty reasonable to me. I don't think we'll gain much value in sending usage reports for our public probes, as we are already collecting telemetry. Taken that with your next comment raises an interesting question. As of now, the default value for `DisableUsageReports` is [false](https://github.com/grafana/synthetic-monitoring-agent/pull/1366/files#diff-5c2ea9d477a93115f8fbe86695012cf1e847c53eaf30efdd179b2ec6819eeb2eR78), which implies that an [HTTPReporter](https://github.com/grafana/synthetic-monitoring-agent/pull/1366/files#diff-5c2ea9d477a93115f8fbe86695012cf1e847c53eaf30efdd179b2ec6819eeb2eR220) instance is created and will send telemetry. 

The implication is that
1. New instances of the agent will run with the new flag's default value without modification of runtime pragma's
2. Usage reports are sent over when new deployments happen
3. Those wanting to opt out will need to update their runtime pragma's

> I mention this because if we decide that we should disable the report in our instances, when we should add the flag in deployment tools, and that cannot happen if the flag is not defined.

Let me take a peak at the jsonnet codebase and see if we can conditionally add the new flag based upon the images release version. This would allow us to stage the change in deployment_tools _prior_ to the release.

PR #1392, opened 2025-07-17
fix: runner: report script errors in the `RunResponse` object

nadiamoe:
The contract for the Runner is such that errors encountered while running the script, as opposed to those encountered trying to start k6, should be reported as part of the `RunResponse` object, using the more-or-less human-readable `Error` and `ErrorCode` fields. If the `Run` method returns an error, this is understood as a failure trying to run k6 whatsoever, and no logs are reported to loki.

This commit fixes a bug where the result of `errorFromLogs`, which by definition is always going to be an error with the script, was returned as an error trying to start k6 instead of in the `Error` and `ErrorCode` fields.

Fixes https://github.com/grafana/synthetic-monitoring/issues/330

### DIFF
--- a/internal/k6runner/local.go
+++ b/internal/k6runner/local.go
@@ -180,7 +180,14 @@ func (r Local) Run(ctx context.Context, script Script, secretStore SecretStore) 
 		fmt.Fprintf(logs, `level=error msg="Metrics output truncated at %d bytes"`+"\n", maxMetricsSizeBytes)
 	}
 
-	return &RunResponse{Metrics: metrics.Bytes(), Logs: logs.Bytes()}, errors.Join(err, errorFromLogs(logs.Bytes()))
+	rr := &RunResponse{Metrics: metrics.Bytes(), Logs: logs.Bytes()}
+	if err := errorFromLogs(logs.Bytes()); err != nil {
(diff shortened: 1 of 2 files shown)


internal/k6runner/local.go
```diff
@@ -180,7 +180,14 @@ func (r Local) Run(ctx context.Context, script Script, secretStore SecretStore)
...
 
-	return &RunResponse{Metrics: metrics.Bytes(), Logs: logs.Bytes()}, errors.Join(err, errorFromLogs(logs.Bytes()))
+	rr := &RunResponse{Metrics: metrics.Bytes(), Logs: logs.Bytes()}
+	if err := errorFromLogs(logs.Bytes()); err != nil {
+		// k6 was run currectly, but the script failed. We capture that failure in the RunResponse.
+		rr.ErrorCode = errorType(err)
+		rr.Error = err.Error()
+	}
+
+	return rr, nil
```
nadiamoe, line 190: This is the highlight of this PR: Instead of returning `errorFromLogs`, it is handled and its result gets put into `RunResponse`. Everything else is instrumental.

mem (approved):
thank you!

PR #1393, opened 2025-07-17
fix: error out early if grpc address doesn't have a port

nadiamoe:
For nearly all users, the port needs to be specified and needs to be `443`. If skipped, GRPC will fall back to a different port that may not respond to SYN TCP segments. As a result, the agent process will just hang without logging anything.

This change causes the agent to error out if the port is omitted, which allegedly should provide earlier feedback about this and other common mistakes.

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -144,6 +144,12 @@ func run(args []string, stdout io.Writer) error {
 		}
 	}
 
+	if _, _, err := net.SplitHostPort(config.GrpcApiServerAddr); err != nil {
+		// SplitHostPort errors if the address has no port. This is intended, as omitting the port in the address is
+		// almost likely a user error that is hard to troubleshoot otherwise.
+		return fmt.Errorf("parsing GRPC api server address %q: %w", config.GrpcApiServerAddr, err)
+	}
+
 	// If the token is provided on the command line, prefer that. Otherwise
 	// pull it from the environment variable SM_AGENT_API_TOKEN. If that's
 	// not available, fallback to API_TOKEN, which was the environment


cmd/synthetic-monitoring-agent/grpc.go
```diff
@@ -14,6 +16,13 @@ import (
 )
 
 func dialAPIServer(ctx context.Context, addr string, allowInsecure bool, apiToken string) (*grpc.ClientConn, error) {
+	_, _, err := net.SplitHostPort(addr)
```
mem: This is likely too late to be doing this validation.

This should be done when the agent is processing arguments, so that it avoid passing such an address to this function.

The validation here can remain as a last stop, but the code shouldn't depend on this happening.

nadiamoe: Makes sense, I've just moved it to `main.go` right after some other validations.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -144,6 +144,12 @@ func run(args []string, stdout io.Writer) error {
 		}
 	}
 
+	if _, _, err := net.SplitHostPort(config.GrpcApiServerAddr); err != nil {
```
mem, line 147: Thank you.

Please note that, for whatever reason, `net.SplitHostPort` accepts an empty host, an empty port, or both. In other words, none of these calls will return an error:

```
net.SplitHostPort("host:")
net.SplitHostPort(":port")
net.SplitHostPort(":")
```

It does require the `:` 🤷🏽, in other words, these do return errors:

```
net.SplitHostPort("host")
net.SplitHostPort("")
```

if the goal is to require a port (and presumably a host), you need:

```
if host, port, err := ...; err != nil || host == "" || port == "" {
```

but note that it will happily accept spaces as either a host or port. IOW, this is valid:

```
net.SplitHostPort("1.1.1.1: ")
```

and it returns ` ` (a single space) instead of an empty string.

How far you have to go depends on what kind of error you want to return to the user. At the very least, the code should try to stop obvious errors that will result in more obscure messages down the road (if it gets to the point where it's calling Dial, it will simply fail with "host not found", "port not found", etc).

nadiamoe: At this point I'm only interested in improving the current situation and catching the obvious "forgot the port, got no error" use case. As this validation is purely additive, I think this should be better than the current state and thus enough IMO.

If we find out other corner cases we want to protect against, we can always add them later.

mem (approved):

PR #1402, opened 2025-07-24
feat: runner/local: run k6 with `--throw` to raise common errors as exceptions

nadiamoe:
Namely, failures making HTTP requests and other common runtime faults will throw exceptions instead of logging a warning and continuing.

Ref:
https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#throw

### DIFF
--- a/internal/k6runner/local.go
+++ b/internal/k6runner/local.go
@@ -211,6 +211,7 @@ func (r Local) buildK6Args(script Script, metricsFn, logsFn, scriptFn, configFil
 		"--no-color",
 		"--no-summary",
 		"--verbose",
+		"--throw", // Abort with an exception on certain abnormal cases: https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#throw
 	}
 
 	// Add secretStore configuration if available


nadiamoe:
Spend a bit of time digging to see where this option is used. So far I've found the following uses:
- Invalid values for metrics: https://github.com/grafana/k6/blob/cdf27ce2b46b3763105da7be42bb38bb8ff151b1/internal/js/modules/k6/metrics/metrics.go#L83
- Errors dialing ws connections: https://github.com/grafana/k6/blob/cdf27ce2b46b3763105da7be42bb38bb8ff151b1/internal/js/modules/k6/ws/ws.go#L140
- Miscellaneous building http requests, such as parsing URLs, headers, request bodies, etc: https://github.com/grafana/k6/blob/cdf27ce2b46b3763105da7be42bb38bb8ff151b1/js/modules/k6/http/request.go#L73
- Errors returned by `http.Client.Do`, such as network errors, name resolution, etc.: https://github.com/grafana/k6/blob/cdf27ce2b46b3763105da7be42bb38bb8ff151b1/lib/netext/httpext/request.go#L328

So I believe this should be safe to apply retroactively.

image

d0ugal (approved):

The-9880 (approved):

PR #1415, opened 2025-08-07
feat: add secret manager boolean to HTTP checks

d0ugal:
For blackbox exporter checks we need are introducing a boolean to enable support for the secrets manager. This allows us to support a prefix convention in fields and maintain backwards compatibility. The following table shows the expected behaviour

| secretManagerEnabled | value              | resolved to            |
|----------------------|--------------------|------------------------|
| false                | gsm:name           | "gsm:name"             |
| false                | password           | "password"             |
| true                 | gsm:name           | value of "name" in gsm |
| true                 | plaintext:password | "password"             |
| true                 | password           | validation error       |


mem (approved):
Thank you!

PR #1468, opened 2025-08-28
feat: add core secret infrastructure and string interpolation system

d0ugal:
Add secret store to blackbox exporter checks.


internal/prober/interpolation/interpolation.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	// Then resolve variables
+	if r.variableProvider != nil {
+		resolvedValue, err := r.resolveVariables(value)
+		if err != nil {
+			return "", err
+		}
+		value = resolvedValue
+	}
```
mem: I'm not following.

Why does resolving the secret and then using *that* value to resolve variables make sense?

mem (commented):
This looks good in general, but there's that thing I'm not follwoing.

internal/prober/interpolation/interpolation.go
d0ugal: Its essentially a two pass on the string because, in theory, it could contain a secret and a variable. Maybe we don't want to support this case but I think it could happen in multihttp.

```
		"mixed secret and variable": {
			input:          "Bearer ${secrets.api-token} for ${username}@${domain}",
			secretEnabled:  true,
			expectedOutput: "Bearer secret-token-123 for admin@example.com",
			expectError:    false,
		},
```
${secrets.api-token} = 1234
${username} = admin
${domain} = grafana.com

1. `"Bearer ${secrets.api-token} for ${username}@${domain}"`
2. `"Bearer 1234 for ${username}@${domain}"`
2. `"Bearer 1234 for admin@grafana.com"`

I don't really like this however so I am working on an improvement to do this in one pass.

mem: OK. That makes sense.

Shouldn't this be the other way around then?

I'm not a 100% sure I want to support this, but I'm going to write this:

```
The secret value is ${secret_name} -> The secret value is ${secrets.api-token}
The secret value is ${secrets.api-token} -> The secret value is secret-token-123
```

Do you see what I mean?

I cannot come up with a realistic scenario doing this (a variable containing the name of the secret to use), but I think the other way around is worse (the secret is the name of the variable).

mem:
Also, can you rebase this (and the other PRs you have)? I would like to validate that the fix for the test failures is actually working.

internal/prober/interpolation/interpolation.go
d0ugal: I got something that works, but i don't love the code. It got a bit messy. I did add test cases for variables within variables though, so that isn't possible now.

Its on this PR in the second commit https://github.com/grafana/synthetic-monitoring-agent/pull/1468/commits/680aebe783880b555b7b0f95c2d33c22abf766b6

mem (approved):
LGTM!

PR #1469, opened 2025-08-28
feat: integrate secret resolution into HTTP prober and agent components

d0ugal:
Add secret store to http checks


internal/prober/http/http.go
d0ugal, line 1: @mem this is the file that I mentioned in our call

PR #1473, opened 2025-08-28
feat: Add IsUsingSecrets to synthetic_monitoring.Check

mem:
This method provides a single way to determine whether a check is using secrets or not. For some types, this will be straighforward (either because they do not support secrets, or because using secrets in that case is very explicit). In other cases, this involves some guesswork, and the extra scripts in this commit are mostly about validating that guesswork.

### DIFF
--- a/pkg/pb/synthetic_monitoring/checks_extra.go
+++ b/pkg/pb/synthetic_monitoring/checks_extra.go
@@ -418,6 +418,69 @@ func (c Check) ConfigVersion() string {
 	return strconv.FormatInt(int64(c.Modified*1000000000), 10)
 }
 
+var (
+	// This pattern matches "k6/secrets" or 'k6/secrets'. This is the basic canary to tell us if
+	// a script might be using secrets.
+	secretsCanary = `(?:'k6/secrets'|"k6/secrets")`
+
+	usesSecretsModuleRe = regexp.MustCompile(
+		`(?m:` +
+			// Default import: import secrets from 'k6/secrets'
+			`^\s*import\s+secrets\s+from\s+` + secretsCanary +
+			`|` +
+			// Named import: import { } from 'k6/secrets'
+			// Named import: import { secrets } from 'k6/secrets'
+			// Named import: import { secrets, foo } from 'k6/secrets'
+			// Named import: import { foo, secrets } from 'k6/secrets'
+			// Named import: import { foo, secrets, bar } from 'k6/secrets'
+			// Named import: import { secrets as s } from 'k6/secrets'
+			`^\s*import\s*{[^}]*}\s+from\s+` + secretsCanary +
+			`|` +
+			// Namespace import: import * as secrets from 'k6/secrets
+			`^\s*import\s+\*\s+as\s+\S+\s+from\s+` + secretsCanary +
+			`|` +
+			// Require syntax: const secrets = require('k6/secrets')
+			`^\s*const\s+secrets\s*=\s*require\s*\(\s*` + secretsCanary + `\s*\)` +
+			`)`)
+)
+
+// IsUsingSecrets returns true if the check uses secrets, false otherwise.
+//
+// Depending on the specific check type, the result is completely accurate or just a guess.
+func (c Check) IsUsingSecrets() bool {
+	switch {
+	case c.Settings.Dns != nil:
+		return false
+
+	case c.Settings.Grpc != nil:
+		return false
+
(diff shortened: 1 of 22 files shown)


pkg/pb/synthetic_monitoring/testdata/script_not_using_secrets_commented_out_import.js
mem, line 1: @d0ugal sorry to make this PR harder to understand.

What's important here is the *name* of the file. That should provide a hint as to the intention. For example this one is NOT using secrets, and it has the import bit commented out.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -418,6 +418,69 @@ func (c Check) ConfigVersion() string {
...
+			// Named import: import { secrets as s } from 'k6/secrets'
+			`^\s*import\s*{[^}]*}\s+from\s+` + secretsCanary +
+			`|` +
+			// Namespace import: import * as secrets from 'k6/secrets
+			`^\s*import\s+\*\s+as\s+\S+\s+from\s+` + secretsCanary +
+			`|` +
+			// Require syntax: const secrets = require('k6/secrets')
+			`^\s*const\s+secrets\s*=\s*require\s*\(\s*` + secretsCanary + `\s*\)` +
+			`)`)
+)
```
mem, line 445: This is the bit that is actually interesting.

I have tried to keep the regular expression simple.

It's matching lines (the `?m:` bit at the start, so it's possible to write `^\s*import`, meaning it's OK to have blanks before `import`, but `import` must be the first thing on the line.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -418,6 +418,69 @@ func (c Check) ConfigVersion() string {
...
+			// Named import: import { foo, secrets } from 'k6/secrets'
+			// Named import: import { foo, secrets, bar } from 'k6/secrets'
+			// Named import: import { secrets as s } from 'k6/secrets'
+			`^\s*import\s*{[^}]*}\s+from\s+` + secretsCanary +
+			`|` +
+			// Namespace import: import * as secrets from 'k6/secrets
+			`^\s*import\s+\*\s+as\s+\S+\s+from\s+` + secretsCanary +
+			`|` +
+			// Require syntax: const secrets = require('k6/secrets')
+			`^\s*const\s+secrets\s*=\s*require\s*\(\s*` + secretsCanary + `\s*\)` +
```
mem, line 443: I don't think I have a test for this, do I? How is this actually used?

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -418,6 +418,69 @@ func (c Check) ConfigVersion() string {
...
+func (c Check) IsUsingSecrets() bool {
+	switch {
+	case c.Settings.Dns != nil:
+		return false
+
+	case c.Settings.Grpc != nil:
+		return false
+
+	case c.Settings.Ping != nil:
+		return false
```
mem, line 459: this a good example of the intention: ping will never use secrets, so it will always return false, unconditionally.

OTOH, HTTP will support secrets, so this needs to be updated in the future.

pkg/pb/synthetic_monitoring/checks_extra_test.go
```diff
@@ -2045,3 +2049,139 @@ func TestRemoteInfoMarshalJSON(t *testing.T) {
...
+			script, err := os.ReadFile(filepath.Join(testdataDir, filename))
+			require.NoError(t, err)
+			assert.False(t, usesSecretsModuleRe.Match(script), "%s should report false", filename)
+
+		default: // skip
+		}
+	}
+}
+
+func TestCheckIsUsingSecrets(t *testing.T) {
```
mem, line 2084: This test is not trying to be exhaustive (that's why the above test exists). This test is just trying to exercise the switch.

mem:
> This looks good. I do wonder if it would be better to improve detection on the API side and send the metadata down to the agent?

That's fair. Are you thinking that at some point the API will have a better / more precise way of determining if a script is using secrets?

What I'm thinking here is that for non-scripted checks, if we have a way to ask whether a check is using secrets, we have have a piece of generic code that decides whether to take different actions based on that. The obvious one is refreshing a token. If we can improve the way we detect that, we might end up adding a field to specific check settings stating that, and this method ends up using that to return an answer.

PR #1503, opened 2025-09-12
fix: Modernize the agent

Pokom:
I get a solid amount of warnings from gopls or Goland when working within the project, and they're often related to either newer features being available or what I would consider nit's. An experimental tool from Go exists called
[modernize](https://pkg.go.dev/golang.org/x/tools/gopls@v0.20.0/internal/analysis/modernize) that allows you to automatically find and fix the code in a safe manner. I was curious to see the diff after running it across the entire project with the following command:

```
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...
```

I've reviewed most of the changes and they all seem reasonably safe. All unit tests appear to be passing as well.

### DIFF
--- a/internal/scraper/scraper_test.go
+++ b/internal/scraper/scraper_test.go
@@ -8,7 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	maps0 "maps"
+	"maps"
 	"net"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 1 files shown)


internal/adhoc/adhoc_test.go
```diff
@@ -523,7 +523,6 @@ func TestDefaultRunnerFactory(t *testing.T) {
 	}
 
 	for name, tc := range testcases {
-		tc := tc
```
nadiamoe, line 526: I know these are not needed nowadays but still get PTSD when I don't see it.

internal/scraper/scraper_test.go
```diff
@@ -1700,9 +1697,7 @@ func (l *testLogger) Log(kv ...interface{}) error {
 func mergeMaps(maps ...map[string]string) map[string]string {
```
nadiamoe: We can probably rename this argument and avoid the ugly `maps0` import.

Pokom: Yep, agreed with that one. I was tracing why it was called `maps0` and can push a fix

The-9880 (approved):

PR #1512, opened 2025-09-17
fix: Do not block during probe registration

mem:
In d3c3c889fe583126564825bd8c97d18a076ba6c1 we introduced an usage reporter. This is working fine, except in cases when the report cannot be sent. If the HTTP client blocks for too long, the agent will fail to complete the registration handshake.

Move sending the report to the end of that process, and run it from a goroutine in order to avoid blocking.


Pokom (approved):
Thanks for catching this and patching it. Having the usage report sent in a non blocking way makes sense 👍🏻

PR #1513, opened 2025-09-18
Add explicit darwin/arm64 platform support

d0ugal:
- Add darwin/arm64 to PLATFORMS in config.mk to ensure it's built and published
- This enables ARM64 macOS builds in GitHub Actions workflows
- Cross-compilation from Linux runners will produce darwin-arm64 binaries

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -3,4 +3,4 @@
 
 DOCKER_TAG = grafana/synthetic-monitoring-agent
 
-PLATFORMS = $(sort $(HOST_OS)/$(HOST_ARCH) linux/amd64 linux/arm64)
+PLATFORMS = $(sort $(HOST_OS)/$(HOST_ARCH) linux/amd64 linux/arm64 darwin/arm64)


mem:
> * Add darwin/arm64 to PLATFORMS in config.mk to ensure it's built and published

It's going to be built, but it's not going to be published.

> * This enables ARM64 macOS builds in GitHub Actions workflows
> * Cross-compilation from Linux runners will produce darwin-arm64 binaries

Note that [this](https://github.com/grafana/synthetic-monitoring-agent/blob/main/.github/workflows/build_and_publish.yaml#L106-L110) will only build the native binaries, and that's running in linux-amd64 and linux-arm64.

From there, the other job uses that output [to push files to the release](https://github.com/grafana/synthetic-monitoring-agent/blob/main/.github/workflows/build_and_publish.yaml#L275-L286), but since the darwin binary is not there, it doesn't get pushed.

I'm guessing the build (as in "make build") doesn't work in Darwin, does it? I would be surprised if that workflow worked without changes in that case.

PR #1533, opened 2025-10-01
fix: add warning when tenant received without secret store details

d0ugal:
Add a log line if we receive tenants without secret store. This will help identify any issues related to missing secret configuration.

### DIFF
--- a/internal/tenants/manager.go
+++ b/internal/tenants/manager.go
@@ -126,6 +126,12 @@ func (tm *Manager) updateTenant(tenant sm.Tenant) {
 
 	info.mutex.Lock()
 	if info.tenant == nil || info.tenant.Modified < tenant.Modified {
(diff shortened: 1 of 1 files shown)


internal/tenants/manager.go
```diff
@@ -126,6 +126,12 @@ func (tm *Manager) updateTenant(tenant sm.Tenant) {
 
 	info.mutex.Lock()
 	if info.tenant == nil || info.tenant.Modified < tenant.Modified {
+		if tenant.SecretStore == nil {
+			tm.logger.Warn().
+				Int64("tenantId", tenant.Id).
+				Msg("tenant received from API without secret store details")
+		}
```
Pokom, line 133: Do warn's get logged automatically in

Pokom (approved):

PR #1542, opened 2025-10-06
chore: add further debug logging around secrets

d0ugal:


internal/secrets/tenant.go
```diff
@@ -34,10 +34,12 @@ func NewTenantSecrets(tp TenantProvider, logger zerolog.Logger) *TenantSecrets {
 
 // GetSecretCredentials gets the secret store configuration for a tenant (backward compatibility)
 func (ts *TenantSecrets) GetSecretCredentials(ctx context.Context, tenantID model.GlobalID) (*sm.SecretStore, error) {
-	if ts.logger.GetLevel() <= zerolog.DebugLevel {
-		tenantID, regionID := model.GetLocalAndRegionIDs(tenantID)
-		ts.logger.Debug().Int("regionID", regionID).Int64("tenantId", tenantID).Msg("getting secret credentials")
-	}
```
d0ugal, line 40: I don't fully understand why it was done this way previously with the if statement.

nadiamoe: Maybe to save the call to `model.GetLocalAndRegionIDs`? I'm not sure if that one is particularly expensive.

nadiamoe: Made a quick check and `model.GetLocalAndRegionIDs` seems to be simple arithmetic, so doing it unconditionally looks good to me.

nadiamoe (approved):

internal/secrets/tenant.go
d0ugal: Thanks!

PR #1544, opened 2025-10-06
feat: Pull cost attribution labels for tenant

Pokom:
Following a similar path as limits to pull cost attribution labels for
tenants. Adds a new internal library `cals/cals.go` that provides an
interface that wraps the `TenantProvider` client. Updates all of the
wiring from Updated => Scraper to pass along the new cal client.

Starts the process of using the library in `scraper.go`, which right now
finds the cost attribution labels for a specific tenant. Returns an
empty slice if nothing is found.

This will be dependent upon the API changes being released to get cost
attribution labels defined at the tenant level.

My goal is to have a follow up PR that updates the telemeter bits to track executions by cost attribution labels. 

- relies on https://github.com/grafana/synthetic-monitoring-api/pull/1750

### DIFF
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -93,7 +93,7 @@ type Factory func(
 	labelsLimiter LabelsLimiter,
 	telemeter *telemetry.Telemeter,
 	secretStore secrets.SecretProvider,
-	costAttributionLabels TenantCals,
+	cals TenantCals,
 ) (*Scraper, error)
 
 type (
--- a/internal/checks/checks_test.go
+++ b/internal/checks/checks_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/grafana/synthetic-monitoring-agent/internal/cals"
 	"github.com/grafana/synthetic-monitoring-agent/internal/secrets"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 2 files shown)


internal/scraper/scraper.go
```diff
@@ -88,6 +94,7 @@ type Factory func(
 	labelsLimiter LabelsLimiter,
 	telemeter *telemetry.Telemeter,
 	secretStore *secrets.TenantSecrets,
+	costAttributionLabels *cals.CostAttributionLabels,
```
The-9880: You can use the interface here as well:

```suggestion
	costAttributionLabels TenantCals,
```

internal/scraper/scraper.go
```diff
@@ -122,28 +129,31 @@ func New(
 	labelsLimiter LabelsLimiter,
 	telemeter *telemetry.Telemeter,
 	secretStore *secrets.TenantSecrets,
+	cals *cals.CostAttributionLabels,
```
The-9880: You can use the interface here as well:

```suggestion
	cals TenantCals,
```

internal/cals/cals_test.go
```diff
@@ -0,0 +1,55 @@
...
+			expectedCostAttributionLabels: []string{},
+			expectError:                   true,
+		},
+	}
+	for name, testcase := range testcases {
+		t.Run(name, func(t *testing.T) {
+			tcal := NewCostAttributionLabels(testcase.tenantProvider)
+			cals, err := tcal.CostAttributionLabels(context.Background(), 1)
+			if testcase.expectError {
+				require.NotNil(t, err)
```
The-9880: If you're interested in a typed assertion here, you could define the following error in `cals.go`:
```
var (
    ErrTenantProvider := errors.New("fetching tenant data")
)
```

And update the `CostAttributionLabels(...)` method to wrap the tenant provider error:
```
if err != nil {
	return nil, fmt.Errorf("%w: %v", ErrTenantProvider, err)
}
```

Which you could assert for here like so:
```
if testcase.expectError {
	require.ErrorIs(t, err, TenantProviderError)
}
```

PR #1576, opened 2025-10-29
feat: Apply jitter to tenants cache

ka3de:
Tenant data is mostly requested from scrapers when a check execution happens. The actual scraper start time has a jitter to randomize checks start up, as most checks are received as a bulk on initial connection. However that jitter is kept small on purpose in order to not delay checks, that makes more sense particularly for new checks.

Therefore this commit applies a jitter at the tenants cache level of up to 25% of the entry timeout (currently 15m). This is only applied when there are no expiration considerations tied to the secrets functionality.

Updates grafana/synthetic-monitoring/issues/378

---
If we take a look at the requests to fetch tenant data that are made from probes, looking per cluster level we can see a certain pattern that matches the current default timeout ([query](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22uvt%22:%7B%22datasource%22:%22000000134%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%20by%28cluster%29%28rate%28server_tenant_request_duration_seconds_count%7Bcluster%3D~%5C%22pop-prod-.%2A%5C%22,%20job%3D%5C%22sm-proxy%2Fsm-proxy%5C%22%7D%5B$__rate_interval%5D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22000000134%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%221761710482733%22,%22to%22:%221761732023692%22%7D,%22compact%22:false%7D%7D&orgId=1)):
Screenshot from 2025-10-29 11-03-12
This change aims to randomize and spread that distribution.

### DIFF
--- a/internal/tenants/manager.go
+++ b/internal/tenants/manager.go
@@ -3,12 +3,19 @@ package tenants
 import (
 	"context"
 	"math"
+	"math/rand/v2"
 	"sync"
 	"time"
 
 	"github.com/grafana/synthetic-monitoring-agent/internal/pusher"
 	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
 	"github.com/rs/zerolog"
+)
+
+const (
+	// DefaultCacheTimeout is the default timeout for the tenant cache.
+	DefaultCacheTimeout = 15 * time.Minute
+	defaultCacheJitter  = 0.25 // 25% jitter
 )
 
 // isSecretStoreConfigured returns true if the SecretStore has both URL and token configured.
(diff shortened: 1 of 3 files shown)


internal/tenants/manager.go
```diff
@@ -141,11 +148,16 @@ func (tm *Manager) calculateValidUntil(tenant *sm.Tenant) time.Time {
 		}
 	}
 
-	validUntil := now.Add(tm.timeout)
+	// In case where the expiration time is set based on the default timeout,
+	// apply a jitter to avoid many agents from hitting the API at the same time.
+	jitterDuration := time.Duration(rand.Float64() * defaultCacheJitter * float64(tm.timeout))
```
Pokom, line 153: Is there any value in logging out the jitterDuration? I'm not sure if it would be a helpful signal for us to watch for in case

ka3de: I don't think there's value on it, all we would get is a number between 0s and 225s. Same applies if we log the full expiration time set for the entry. There are more particular cases for when the secret store expiration time has an impact in the entry expiration time calculation, but these are already covered by logging (that's debug level logging, but I think that's the right scope).
What is your reasoning to consider logging the jitter time?

Pokom: > all we would get is a number between 0s and 225s.

That makes sense. My main reasoning for logging the value  was basically "What if we got the jitter calculation wrong, how would we know?". It would be more practical to write a test to ensure the boundaries are always enforced than it would be to log it out 👍🏻

ka3de: There are tests. See https://github.com/grafana/synthetic-monitoring-agent/blob/9c79f1d3eaf04c8bd6b9ea9d09c9e2e4850acf3b/internal/tenants/manager_test.go#L168

Issue #1350, opened 2025-06-17
Adhoc check 504s give the impression something is wrong


peterschretlen:
Internal feedback:  In private probes, it's common to see logs with `504 (Gateway Timeout)` every 5 minutes. The logs looks like this:
```
{"level":"info","program":"synthetic-monitoring-agent","subsystem":"adhoc","time":1749842581207,"message":"fetching ad-hoc checks from synthetic-monitoring-api"}
{"level":"info","program":"synthetic-monitoring-agent","subsystem":"adhoc","probe id":18626,"probe name":"prd-apps","time":1749842581227,"message":"registered ad-hoc probe with synthetic-monitoring-api"}
{"level":"error","program":"synthetic-monitoring-agent","subsystem":"adhoc","error":"rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 504 (Gateway Timeout); transport: received unexpected content-type \"text/html\"","time":1749842881240,"message":"receiving ad-hoc check"}
{"level":"warn","program":"synthetic-monitoring-agent","subsystem":"adhoc","error":"rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 504 (Gateway Timeout); transport: received unexpected content-type \"text/html\"","connection_state":"READY","time":1749842881240,"message":"handling check changes"}
```

This creates the impression that something is wrong, but it's not actually a problem.   There is a 5 minute timeout on the connection if it's idle (on the server side).  Since the adhoc "test" button is not used very often, the connection is going to be idle more often that not, which is why these show up in the logs.  

There's nothing wrong in the agent, but private probe users may mistake these logs for a problem.  Is there a way to separate "normal" gateway timeouts from errors that probe operators should pay attention to?

peterschretlen:
Here is an [example](https://github.com/grafana/support-escalations/issues/16928#issuecomment-2992311194) (internal only link) of a customer thinking this was related to a problem they were having with a private probe:

PR #1578, opened 2025-10-29
fix: Handle 504 Gateway Timeout false alarms in adhoc checks

d0ugal:
Fixes https://github.com/grafana/synthetic-monitoring-agent/issues/1350

This handles timeouts in the adhoc subsystem, logging an info message and reconnecting. This should reduce the excessive 504 error logs the agent currently outputs.

### DIFF
--- a/internal/adhoc/adhoc_test.go
+++ b/internal/adhoc/adhoc_test.go
@@ -136,6 +136,14 @@ func (testBackoff) Duration() time.Duration {
 
 func (testBackoff) Reset() {}
 
+type timeoutBackoff struct{}
+
+func (timeoutBackoff) Duration() time.Duration {
+	return 100 * time.Millisecond
+}
+
+func (timeoutBackoff) Reset() {}
+
 type grpcTestConn struct {
 }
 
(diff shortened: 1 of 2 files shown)


internal/adhoc/adhoc_test.go
```diff
@@ -352,6 +371,61 @@ func (p *testProber) Probe(ctx context.Context, target string, registry *prometh
...
+	h, err := NewHandler(opts)
+	require.NoError(t, err)
+	require.NotNil(t, h)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	runErr := h.Run(ctx)
+	require.Error(t, runErr)
+	require.Contains(t, runErr.Error(), "context deadline exceeded")
```
nadiamoe, line 419: I think you should be able to use errors.Is (through require) here:

```suggestion
	require.ErrorIs(t, runErr, context.DeadlineExceeded)
```

nadiamoe (approved):
Looking good, just left a suggestion in case comparing the error as a string was not done on purpose.

PR #1590, opened 2025-11-04
Add k6 version repository for dynamic binary selection

nadiamoe:
This PR introduces the version Repository, which scans a specific folder for k6 binaries and allows mapping semver constraints to them.

Related to https://github.com/grafana/synthetic-monitoring/issues/413, https://github.com/grafana/synthetic-monitoring/issues/416

### DIFF
--- a/internal/k6runner/version/repository.go
+++ b/internal/k6runner/version/repository.go
@@ -35,6 +35,40 @@ type Repository struct {
 	entries []Entry
 }
 
+// NewRepository creates a new Repository object.
+func NewRepository(root, override string) (*Repository, error) {
+	if root == "" && override == "" {
+		return nil, fmt.Errorf("either root or override should be provided")
+	}
+
+	if root != "" {
+		rootStat, err := os.Stat(root)
+		if err != nil {
+			return nil, fmt.Errorf("could not stat repository root: %w", err)
+		}
+
+		if !rootStat.IsDir() {
+			return nil, fmt.Errorf("repository root is not a directory")
+		}
+	}
+
+	if override != "" {
+		overrideStat, err := os.Stat(override)
+		if err != nil {
+			return nil, fmt.Errorf("could not stat repository override: %w", err)
+		}
+
+		if overrideStat.IsDir() || overrideStat.Mode()&0o100 == 0 {
+			return nil, fmt.Errorf("repository override should be a single executable file")
+		}
+	}
+
+	return &Repository{
+		Root:     root,
+		Override: override,
+	}, nil
+}
+
 const binaryMustContain = "k6"
 
 // Entry stores the path to a k6 binary, and the parsed semantic version that binary returns.
--- a/internal/k6runner/version/repository_test.go
+++ b/internal/k6runner/version/repository_test.go
@@ -11,9 +11,9 @@ import (
 func TestOverride(t *testing.T) {
 	t.Parallel()
 
-	repo := version.Repository{
-		Root:     "./testdata/",
-		Override: "./testdata/override/sm-k6-custom",
+	repo, err := version.NewRepository("./testdata/", "./testdata/override/sm-k6-custom")
+	if err != nil {
+		t.Fatalf("could not create repository: %v", err)
 	}
 
 	versions, err := repo.Entries()
@@ -33,8 +33,9 @@ func TestOverride(t *testing.T) {
 func TestVersions(t *testing.T) {
 	t.Parallel()
 
-	repo := version.Repository{
-		Root: "./testdata/",
+	repo, err := version.NewRepository("./testdata/", "")
+	if err != nil {
+		t.Fatalf("could not create repository: %v", err)
 	}
 
 	versions, err := repo.Entries()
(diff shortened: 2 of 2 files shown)


internal/k6runner/version/repository.go
```diff
@@ -0,0 +1,239 @@
...
+
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	if len(r.entries) == 0 {
+		return nil, nil
+	}
+
+	entries := make([]Entry, len(r.entries))
+	copy(entries, r.entries)
```
ka3de, line 140: Notice that, since `Entry.Version` is `*semver.Version` this will do just a shallow copy and therefore the caller has access to the underlying version definition. That might be OK if we don't expect callers mutating the data, just consuming it.

internal/k6runner/version/repository.go
```diff
@@ -0,0 +1,239 @@
...
+	Version   string `json:"version"`
+}
+
+// ErrUnsatisfiable is returned when no binary matching a constraint is found.
+var ErrUnsatisfiable = errors.New("no compatible k6 version found")
+
+// BinaryFor walks the repository and returns the path to the binary with the highest version that matches
+// constraintStr. If no binary matching the constraint is found, ErrUnsatisfiable is returned.
+// BinaryFor scans the root folder if needed.
+func (r *Repository) BinaryFor(constraintStr string) (string, error) {
```
ka3de, line 94: To understand better the usage of this method, how do we expect to use this? Currently AFAIK the k6 runner implementation accepts a static "k6 path". I guess that will be changed so the path is resolved as part of the [`Run()`](https://github.com/grafana/synthetic-monitoring-agent/blob/191471cc59055fc665567118ee66dc3481831f0c/internal/k6runner/k6runner.go#L94) method? And therefore _run_ will be extended (possible [`Settings`](https://github.com/grafana/synthetic-monitoring-agent/blob/191471cc59055fc665567118ee66dc3481831f0c/internal/k6runner/k6runner.go#L29) to accept the specified k6 channel?

internal/k6runner/version/repository.go
```diff
@@ -0,0 +1,239 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/rs/zerolog"
+)
+
+// Repository points to a collection of k6 binaries.
+type Repository struct {
```
ka3de, line 23: Do we need a constructor for this?

internal/k6runner/version/repository_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	repo := version.Repository{
+		Root:     "./testdata/",
+		Override: "./testdata/override/sm-k6-custom",
+	}
+
+	versions, err := repo.Entries()
+	if err != nil {
+		t.Fatalf("retrieving entries: %v", err)
+	}
```
ka3de, line 22: We generally use `github.com/stretchr/testify/` and `require` in this case. Should we apply that to this and the other unit tests for this?

ka3de (commented):
Looking good! Just some small comments.

internal/k6runner/version/repository.go
nadiamoe: Ah, well spotted, the decision to return the copy predates using semver pointers for this... I think there's still some purpose to returning a copy, so callers can safely `append` and such. I'll document that callers shouldn't modify the underlying objects.

internal/k6runner/version/repository.go
nadiamoe: Good point, this is worth explaining. What I'm thinking about is to keep the `--k6-path` flag, but change the default value to the empty string. This flag will become `Repository.Override`, which if empty (default) will do nothing. Independently, we'll add another flag whose name I haven't decided yet, but can be something like `--k6-repository-path` or something along those lines, which will become the `Repository.Root`. It will default to `/usr/libexec/sm-k6/` or something like that, where we will place k6 binaries.

internal/k6runner/version/repository.go
nadiamoe: Do we? I was thinking about instantiating the struct setting the properties. There's no magic to be done.

internal/k6runner/version/repository_test.go
nadiamoe: This feels like a style choice to me. You're welcome to change it if you have a strong opinion about it (I don't), feel free to open a PR off this one and do so 😊

internal/k6runner/version/repository.go
ka3de: My rule of thumb is _"if the zero value of the struct is not usable, you should use a constructor"_.
I understand your comment that there is not extra magic in this case, which is true, but I'm also wondering if there should be? If we used a constructor we could validate:
- Either root or override must be provided.
- Root/Override must be a valid path (If we want to go "the extra mile" we could potentially validate that root is a path to a directory, and override to an executable binary, but probably not critical).

mem:
Does something else need to happen here, @nadiamoe ?

internal/k6runner/version/repository.go
nadiamoe: Added the constructor plus some sanity checks in https://github.com/grafana/synthetic-monitoring-agent/pull/1590/commits/b74ec9ca3dcd492895dc23953c17ae5dfc094e95

Issue #1508, opened 2025-09-15
Log duration_seconds and metric probe_duration_seconds show different values


ckbedwell:
In some instances (I haven't worked out the correlation yet -- it's not consistent) the `duration_seconds` value in the ending logs label (e.g. logs where the `msg` value is either 'Check succeeded' or `Check failed') and the `probe_duration_seconds` metric published to Mimir do not match.

**Some examples in ops:**
[Explore Mimir probe_duration_seconds](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22gwc%22:%7B%22datasource%22:%22grafanacloud-prom%22,%22queries%22:%5B%7B%22expr%22:%22probe_duration_seconds%7Bjob%3D%5C%22login%5C%22,%20instance%3D%5C%22https:%2F%2Fgrafana.com%2Fapi%2Flogin%5C%22,%20probe%3D%5C%22Calgary%5C%22%7D%5B$__range%5D%22,%22instant%22:true,%22format%22:%22heatmap%22,%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22grafanacloud-prom%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%221757929157000%22,%22to%22:%221757929164000%22%7D,%22compact%22:false%7D%7D&orgId=1): 0.140
[...]

Issue #1568, opened 2025-10-22
Inconsistent duration reporting


mem:
Depending on the type of check, we are reporting different values for the `probe_duration_seconds` metric.

There's [collectData](https://github.com/grafana/synthetic-monitoring-agent/blob/2ca5380f6c06354256c8fb6286ef969c75e2bd21/internal/scraper/scraper.go#L473) that is messing with the duration info via calling [patchDuration](https://github.com/grafana/synthetic-monitoring-agent/blob/2ca5380f6c06354256c8fb6286ef969c75e2bd21/internal/scraper/scraper.go#L589) (this function modifies `probe_duration_seconds`).

[collectData](https://github.com/grafana/synthetic-monitoring-agent/blob/2ca5380f6c06354256c8fb6286ef969c75e2bd21/internal/scraper/scraper.go#L473) -> [getProbeMetrics](https://github.com/grafana/synthetic-monitoring-agent/blob/2ca5380f6c06354256c8fb6286ef969c75e2bd21/internal/scraper/scraper.go#L623) -> [runProber](https://github.com/grafana/synthetic-monitoring-agent/blob/2ca5380f6c06354256c8fb6286ef969c75e2bd21/internal/scraper/scraper.go#L659) -> [prober.Probe](https://github.com/grafana/synthetic-monitoring-agent/blob/2ca5380f6c06354256c8fb6286ef969c75e2bd21/internal/prober/scripted/scripted.go#L67) (or any of the other implementations)

That last call is reporting a duration and if that bit of data is not available, it uses the time between calling `probe.Prober` and that function returning (this will include k6 overhead, etc).

`runProber` is emitting a log line with that duration:

```
level=error target=test probe=London region=EMEA instance=test job="weird metrics" check_name=scripted source=synthetic-monitoring-agent msg="Check failed" duration_seconds=5.18794805 walltime_seconds=5.18794805
```

but the duration there is not the same as the "patched" version.

The-9880:
Closing in favour of existing PR

PR #1593, opened 2025-11-06
feat: make `probe_duration_seconds` match `probe_script_duration_seconds`

nadiamoe:
For the vast majority of check types, `probe_duration_seconds` is reported as the wall time that took to run the prober. This is accurate for the vast majority of cases, but there's one in particular where it is not: k6-backed checks that use external runners.

External runners can take a certain time to process a request, which prior to this commit was counted as part of `probe_duration_seconds` while in fact it was not. This commit adds code that parses the `probe_script_duration_seconds` metric returned by k6 runners, and uses that as the duration of the check for both logs and `probe_duration_seconds`.

Fixes https://github.com/grafana/synthetic-monitoring-agent/issues/1508

### DIFF
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -313,8 +313,7 @@ func (dc *probeDurationCollector) process(_ *dto.MetricFamily, sample *model.Sam
 		return nil
 	}
 
-	// TODO: Is there a better way to convert from float seconds to time?
-	dc.duration = time.Duration(sample.Value*1000) * time.Millisecond // This truncates to the nearest millisecond.
+	dc.duration = time.Duration(float64(sample.Value) * float64(time.Second))
 
 	return nil
 }


internal/k6runner/k6runner.go
```diff
@@ -302,6 +304,21 @@ func (rc *checkResultCollector) process(mf *dto.MetricFamily, sample *model.Samp
...
+	duration time.Duration
+}
+
+func (dc *probeDurationCollector) process(_ *dto.MetricFamily, sample *model.Sample) error {
+	if sample.Metric[model.MetricNameLabel] != "probe_script_duration_seconds" {
+		return nil
+	}
+
+	// TODO: Is there a better way to convert from float seconds to time?
+	dc.duration = time.Duration(sample.Value*1000) * time.Millisecond // This truncates to the nearest millisecond.
```
nadiamoe: Open to suggestions here https://raintank-corp.slack.com/archives/C8AQC4F7T/p1762425129680709

internal/prober/scripted/scripted.go
```diff
@@ -78,14 +78,14 @@ func (p Prober) Probe(ctx context.Context, target string, registry *prometheus.R
...
 		Msg("secret store retrieved for scripted probe")
 
-	success, err := p.processor.Run(ctx, registry, logger, p.logger, secretStore)
+	success, duration, err := p.processor.Run(ctx, registry, logger, p.logger, secretStore)
 	if err != nil {
 		p.logger.Error().Err(err).Msg("running probe")
 		return false, 0
 	}
 
 	// TODO(mem): implement custom duration extraction.
```
The-9880: Looks like these `TODO`s can be removed from each of these k6runner `Probe` implementations

The-9880 (approved):
Nice - also closes https://github.com/grafana/synthetic-monitoring-agent/issues/1568.

internal/prober/scripted/scripted.go
nadiamoe: Done!

The-9880 (approved):

PR #1600, opened 2025-11-11
fix: Update DEVELOPMENT.md

mem:
This is a high level overview of the code, with some hints as to how to interact with it.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -56,7 +56,7 @@ configuration file targets v2.
 
 ### Code generation
 
-Some of the code and support files are generated, especifically:
+Some of the code and support files are generated, specifically:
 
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -1,87 +1,294 @@
...
 ```
 
-The documentation for other targets is available thru:
+Again, the linting tools are executed in a docker container, to ensure that
+versions match. You can use `golangci-lint` directly, but keep in mind that the
+configuration file targets v2.
+
+### Code generation
+
+Some of the code and support files are generated, especifically:
```
d0ugal: specifically

mem: argh, thank you!

I spell checked the whole file.

d0ugal (approved):

PR #1601, opened 2025-11-12
feat: Rework tenant cache

mem:
In order to have more resilience in the face of outages, implement an out-of-process cache using memcached. In order to support the existing mode of operation, add an in-process-cache using Otter. Rework the tenant cache to use this new system.

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -501,7 +501,7 @@ func setupCache(cacheType cache.Kind, memcachedServers []string, localCapacity i
 	// auto + servers provided -> memcached -> local -> noop
 	// auto + no servers -> local -> noop
 	effectiveType := cacheType
-	if cacheType == "auto" {
+	if cacheType == cache.KindAuto {
 		if len(memcachedServers) > 0 {
 			effectiveType = cache.KindMemcached
 		} else {


internal/tenants/manager.go
mem, line 1: This is the bit that needs the most attention.

internal/cache/cache.go
```diff
@@ -0,0 +1,81 @@
...
+	// a cache miss and other errors (network issues, serialization failures, etc.).
+	ErrCacheMiss = error_types.BasicError("cache miss")
+)
+
+// Cache defines the interface for cache operations.
+// All cache implementations (memcached, noop, local, etc.) must implement this interface.
+//
+// This interface allows the agent to use different cache backends interchangeably,
+// including a no-op implementation that eliminates the need for nil checks in client code.
+type Cache interface {
```
mem, line 65: There are three implementations of this interface.

A memcached one. This is the one I would like to use eventually.

A local one, based on Otter.

A noop one, which is a fallback in case the other two are not available.

internal/cache/cache.go
```diff
@@ -0,0 +1,81 @@
...
+	// Set stores a value in the cache with the specified expiration time.
+	// Returns an error if the operation fails.
+	Set(ctx context.Context, key string, value any, expiration time.Duration) error
```
mem, line 68: This is the bit where I struggled the most.

Otter uses generics to have a more strongly-typed cache. Since the goal is to use memcached, which does not have a strongly-typed API, I thought about multiple clients, one per type, but it was weird.

internal/cache/cache.go
mem: I did consider a multi-tiered implementation, too, using L1 with Otter and L2 with memcached. I would prefer to get some numbers before going that way.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -469,3 +491,83 @@ func setupGoMemLimit(ratio float64) error {
...
 }
+
+func setupCache(cacheType string, memcachedServers []string, localCapacity int, localTTL time.Duration, logger *zerolog.Logger) cache.Cache {
+	// Determine effective cache type with auto mode logic:
+	// auto + servers provided -> memcached -> local -> noop
+	// auto + no servers -> local -> noop
+	effectiveType := cacheType
+	if cacheType == "auto" {
+		if len(memcachedServers) > 0 {
+			effectiveType = "memcached"
```
The-9880: This may make sense as a typed string in the `cache` package, so the valid values are defined upfront and we don't need to handle the literals.

e.g:
```
package cache

type Type string

const (
	TypeMemcached Type = "memcached"
	TypeLocal     Type = "local"
	TypeNoop      Type = "noop"
)
```

internal/cache/local.go
```diff
@@ -0,0 +1,171 @@
...
+	if ttl == 0 {
+		ttl = l.defaultTTL
+	}
+
+	// Set the value
+	l.cache.Set(key, data)
+
+	// Set expiration if TTL is specified
+	if ttl > 0 {
+		l.cache.SetExpiresAfter(key, ttl)
```
The-9880: I think we can omit the `expiration` param and remove this bit - the `ExpiryCalculator` option already ensures we are using either the `config.DefaultTTL`, or else `24h` (if `config.DefaultTTL == 0`).

My only reason to keep this is if there's a need to set different per-item TTLs, which could make sense if we're storing different types of objects in the cache (currently, there's only `cachedTenant`).

internal/cache/local_test.go
```diff
@@ -0,0 +1,585 @@
...
+	cache, err := NewLocal(LocalConfig{
+		MaxCapacity: 1000,
+		Logger:      logger,
+	})
+	require.NoError(t, err)
+	defer cache.Close()
+
+	ctx := t.Context()
+
+	t.Run("flush clears all items", func(t *testing.T) {
```
The-9880: Is `t.Run()` here just to visually separate the setup? Or to annotate the test? I don't see a big advantage to it otherwise as there's only one subtest.

internal/cache/local_test.go
```diff
@@ -0,0 +1,585 @@
...
+		var result int
+		err = cache.Get(ctx, "test:flush:0", &result)
+		require.ErrorIs(t, err, ErrCacheMiss)
+	})
+}
+
+func TestLocalCacheCapacity(t *testing.T) {
+	logger := zerolog.New(io.Discard)
+
+	t.Run("respects max capacity", func(t *testing.T) {
```
The-9880: Similar q, why bother with `t.Run()` in this test?

internal/tenants/manager.go
```diff
@@ -162,128 +179,164 @@ func (tm *Manager) calculateValidUntil(tenant *sm.Tenant) time.Time {
...
+	mutex := tm.getFetchMutex(tenant.Id)
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	ctx := context.Background()
+	key := cacheKey(tenant.Id)
+
+	// Check if there's already a cached tenant
+	var existing cachedTenant
+	err := tm.cache.Get(ctx, key, &existing)
```
The-9880, line 192: Want to log this if it's non-nil?

internal/tenants/manager.go
```diff
@@ -162,128 +179,164 @@ func (tm *Manager) calculateValidUntil(tenant *sm.Tenant) time.Time {
...
+	// Create cached tenant
+	cached := cachedTenant{
+		Tenant:     &tenant,
+		ValidUntil: validUntil,
+		Modified:   tenant.Modified,
+	}
+
+	// Store in cache without TTL - we'll check ValidUntil manually
+	// This allows us to return stale data if the API is unavailable
+	if err := tm.cache.Set(ctx, key, cached, 0); err != nil {
```
The-9880, line 227: What do you think about defining typed caches on top of the `Cache` interface?

The assumption being that items of the same object class (e.g. `cachedTenant`) would be cached with the same `TTL` (`0` in this case). The counter-case would be if you ever wanted to have different TTLs for objects of the same kind, e.g. using `validUntil` as the TTL for `cachedTenant`.

This lets us omit `expiration` from `Set` by constructing a `Cache<cachedTenant>` which uses `0` for all its entries. You could also move the key namespace prefix (`tenant:`) to the typed cache's `Set` logic via its constructor.

So `NewManager` would accept a `CacheProvider` as an argument, and would initialize its own cache interface like:
```
tenantCache := cacheProvider.NewCache<cachedTenant>(namespace="tenant:", TTL=0)

tm := &Manager{
		tenantCh:      tenantCh,
		tenantsClient: tenantsClient,
		timeout:       timeout,
		cache:         tenantCache,
		logger:        logger,
		fetchMutexes:  xsync.NewMap[int64, *sync.Mutex](),
	}
```

If we end up caching other objects, they would initialize a separate `Cache<otherObject>` with its own TTL and key prefix.

The-9880, line 227: Another note: it's documented on the implementations, but `0` means something different between caching backends:
* `Local`: max of `24h` TTL, if `config.DefaultTTL` is not defined/valid.
* `Memcached`: as long as the memcached server is alive (no explicit expiry).

It makes sense, but it did trip me for a moment.

internal/cache/cache.go
The-9880: Yeah, I think I initially expected a multi-tiered implementation. It's not something we need to implement immediately, I'm fine with running this using one or the other first and getting some data out of it.

mem: The idea is to run with Otter by default and test with memcached in a couple of places.

cmd/synthetic-monitoring-agent/main.go
mem: I've added that.

internal/cache/local_test.go
mem: I was actually thinking of adding something else, but ended up implementing just that one test.

Removed.

internal/cache/local_test.go
mem: Habit...

Before writing the tests, I actually write a plan (in the form of `t.Run("do something something...", ...)`).

internal/tenants/manager.go
mem: That sounds reasonable, but it's a "later" thing.

internal/tenants/manager.go
mem: hmm... here the expected error would be a cache miss. In this path, the key should be valid. The other possible error is failing to contact memcached (in that scenario) but that error is logged at debug level _in_ the MemcachedClient implementation. The other error (that is also logged in the implementation) is failing to deserialize (which should mean the cache is corrupt).

I'm OK with bubbling up the logging to this layer, but the reason why it's in the implementation (for now) is that it is somewhat type-specific. In this code, the only condition that we should care about is the fact that there wasn't an error. This is not common, but in my head that's because this is a cache: cache misses should not be treated as exceptional situations.

mem:
Apologies for rebasing. I didn't realize multiple things had gone in since the original PR.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -473,3 +495,84 @@ func setupGoMemLimit(ratio float64) error {
 
 	return nil
 }
+
+func setupCache(cacheType cache.Kind, memcachedServers []string, localCapacity int, localTTL time.Duration, logger *zerolog.Logger) cache.Cache {
+	// Determine effective cache type with auto mode logic:
+	// auto + servers provided -> memcached -> local -> noop
+	// auto + no servers -> local -> noop
+	effectiveType := cacheType
+	if cacheType == "auto" {
```
The-9880:
```suggestion
	if cacheType == cache.KindAuto {
```

The-9880 (dismissed):

PR #1604, opened 2025-11-14
prober/traceroute: log traceroute entries in increasing hop order

nadiamoe:
Previously, results of traceroute hops were logged in random order as an artifact of them being stored in a map. This PR does not change the data structure, but ensures we walk it in ascending key order. Functionality should not change, but hopefully logs will be more readable as a result.


mem (approved):

nadiamoe:
Rebased and added the missing commit prefix.

Issue #738, opened 2024-06-21
Change gRPC client code to use `grpc.NewClient`


roobre:
[`grpc.Dial`](https://pkg.go.dev/google.golang.org/grpc#Dial) and [`grpc.DialContext`](https://pkg.go.dev/google.golang.org/grpc#DialContext) are deprecated, although they will be supported throughout `v1.x.x`.

We use this to dial the API gRPC server during the agent initialization. The suggested way to do this is to replace this:
https://github.com/grafana/synthetic-monitoring-agent/blob/83f8603977c4785e611fa381a14eea933b80826b/cmd/synthetic-monitoring-agent/grpc.go#L43
With something such as:
```
return grpc.NewClient(addr, opts...)
```
Or:
```
con, err := grpc.NewClient(addr, opts...)
if err != nil {
    return nil, fmt.Errorf("building client: %w", err)
}
con.Connect() // Although this does not block waiting for conn, and therefore never returns error
return con, nil
```
But this modifies our current behavior where we [block](https://github.com/grafana/synthetic-monitoring-agent/blob/83f8603977c4785e611fa381a14eea933b80826b/cmd/synthetic-monitoring-agent/grpc.go#L20) and force the connection with the API before returning from the [`dialAPIServer`](https://github.com/grafana/synthetic-monitoring-agent/blob/83f8603977c4785e611fa381a14eea933b80826b/cmd/synthetic-monitoring-agent/grpc.go#L16) function. This seems to be an [anti pattern](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md) and we should just expect the connection to be established during the first rpc call, and be able to handle connection errors as a result of that call, which I believe we already do.
So, we can probably remove the blocking behavior from main and just expect the first gRPC service call to connect, because the agent will try to fetch data from the API when starting, that will happen early when initializing also, but I'm keeping this PR open by now to review this better.

_Originally posted by @ka3de in https://github.com/grafana/synthetic-monitoring-agent/issues/671#issuecomment-2039564231_

roobre:
This also applies to the proxy here: https://github.com/grafana/sm-proxy/pull/106

PR #1632, opened 2025-11-27
fix: Migrate from gprc.DialContext to grpc.NewClient

mem:
This moves to the non-deprecated API.

Fixes: https://github.com/grafana/synthetic-monitoring-agent/issues/738

### DIFF
--- a/docs/GRPC.md
+++ b/docs/GRPC.md
@@ -1,0 +1,734 @@
+# gRPC Connection Management in Synthetic Monitoring Agent
+
+This document provides a comprehensive overview of how gRPC connections are handled in the
+synthetic-monitoring-agent, including architecture, configuration options, and best practices
+implementation.
+
+---
+
+## Table of Contents
+
+1. [Architecture Overview](#architecture-overview)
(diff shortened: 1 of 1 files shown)


internal/adhoc/adhoc.go
```diff
@@ -174,6 +180,43 @@ func NewHandler(opts HandlerOpts) (*Handler, error) {
...
+		tlsVersion = "TLS 1.0"
+	case tls.VersionTLS11:
+		tlsVersion = "TLS 1.1"
+	case tls.VersionTLS12:
+		tlsVersion = "TLS 1.2"
+	case tls.VersionTLS13:
+		tlsVersion = "TLS 1.3"
+	default:
+		tlsVersion = fmt.Sprintf("unknown (0x%04x)", tlsInfo.State.Version)
+	}
```
d0ugal: Can we use tls.VersionName? https://pkg.go.dev/crypto/tls#VersionName

mem: ah, thank you!

d0ugal (dismissed):

PR #1636, opened 2025-12-01
feat: Report k6 timestamp on k6 execution logs

The-9880:
For discussion - the `time` label reports the actual timestamp of the log as k6 encountered it, while the `ts` label corresponds to when the agent extracted the logs:
image

This change uses the `time` reported by k6 as the Timestamp for Loki entries, rather than the time of log extraction.

### DIFF
--- a/internal/scraper/scraper_test.go
+++ b/internal/scraper/scraper_test.go
@@ -2150,6 +2150,24 @@ func TestExtractLogsK6TimeOverridesDefaultTimestamp(t *testing.T) {
 			},
 		},
 		{
+			name: "k6 time field before ts (order independence)",
+			// Defensive case: time= appears before ts= in the log line. The guard on the ts= case
+			// (if !t.IsZero() { continue }) ensures ts= does not overwrite a time= already seen.
+			logs: `time="2023-06-01T13:40:26-06:00" ts=2023-06-01T20:00:00Z level=info msg="k6 message"`,
+			sharedLabels: []labelPair{
+				{name: "probe", value: "test-probe"},
+			},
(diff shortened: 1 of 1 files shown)


internal/scraper/scraper.go
```diff
@@ -816,6 +820,15 @@ RECORD:
 					continue RECORD
 				}
 
+			// k6 sets a 'time' field on its logs using RFC3339; if present, report it instead of the default timestamp
```
mem, line 852: I might be misunderstanding something here.

In principle, it's possible for the `ts` and `time` to show up in any order. I think L809 is trying to cover that and prefer the `time` field over the `ts` one. There's no test for that, though.

Question: why is it better to handle that here and not in the extension? Or is it the case that it was handled in the extension and that is no longer true?

The-9880: I updated this with a corresponding test for:
>I think L809 is trying to cover that and prefer the time field over the ts one. There's no test for that, though.

We should always see `ts` before `time` based on the docs for `go-kit/log log.With()` which say our labels should be pre-pended ([ref](https://github.com/go-kit/log/blob/main/log.go#L18)), but the tests account for either case anyways.

re: the question, I think it's a bigger approach to try to tackle this from the extension - k6 uses `logrus` with the default `FieldKeytime` (`"time"`) for the timestamp field. If the ask is to have the extension re-map this, that would be a bigger change. AFAICT, output extensions are only meant to manipulate the shape of metrics and not logs.

The other side of that is if we can get `ts` from the extension, the agent will still pre-pend its own `ts` when parsing logs. To avoid doing so, we would need to use a separate logger to parse logs from k6.

PR #1642, opened 2025-12-08
feat: track executions by cals

Pokom:
Resolves https://github.com/grafana/synthetic-monitoring/issues/391

Extend the region pusher codebase to track executions by Cost Attribution Labels(CAL's) by adding an addition map to the telemetry attribution. The `telemetry` map now is now:

```go
// key int64: tenant id
// middle string: check class
// last string: serialized cost attribution label
telemetry   map[int64]map[sm.CheckClass]map[string]*sm.CheckClassTelemetry
```

CAL's are stored as `[]sm.CostAttributionLabel`, which can not be indexed in a map. Two methods were introduced:
1. `serializeCal`
2. `deserializeCal`

The methods allows the program to go from a slice to a string for the key and back. 

To preserve existing state, if no cost attribution labels are set, then the executions are tracked with the key `__MISSING__`. When executions are pushed to the API, this gets deserialized to `[]sm.CostAttributionLabel{}` which preserves the API's ability to emit the telemetry as it does today. A PR will be made to the API which will handle the scenarios where no cost attribution labels are set(existing state today), and when the slice is set.

### DIFF
--- a/internal/telemetry/telemeter_test.go
+++ b/internal/telemetry/telemeter_test.go
@@ -24,33 +24,45 @@ func TestTelemeterAddExecution(t *testing.T) {
 			p, ok := tele.pushers[regionID]
 			require.True(t, ok)
 
-			// sum expected executions data
-			regionTele := make(map[int64]map[sm.CheckClass]*sm.CheckClassTelemetry)
+			// sum expected executions data, indexed by tenant -> checkClass -> calKey
+			regionTele := make(map[int64]map[sm.CheckClass]map[string]*sm.CheckClassTelemetry)
 			for _, e := range ee {
 				tenantTele, ok := regionTele[e.LocalTenantID]
 				if !ok {
-					tenantTele = make(map[sm.CheckClass]*sm.CheckClassTelemetry)
+					tenantTele = make(map[sm.CheckClass]map[string]*sm.CheckClassTelemetry)
 					regionTele[e.LocalTenantID] = tenantTele
 				}
-				if _, ok := tenantTele[e.CheckClass]; !ok {
-					tenantTele[e.CheckClass] = &sm.CheckClassTelemetry{CheckClass: e.CheckClass}
+				clTele, ok := tenantTele[e.CheckClass]
+				if !ok {
+					clTele = make(map[string]*sm.CheckClassTelemetry)
+					tenantTele[e.CheckClass] = clTele
 				}
-				tenantTele[e.CheckClass].Executions++
-				tenantTele[e.CheckClass].Duration += float32(e.Duration.Seconds())
+				calKey := serializeCALs(e.CostAttributionLabels)
+				calTele, ok := clTele[calKey]
+				if !ok {
+					calTele = &sm.CheckClassTelemetry{
+						CheckClass:            e.CheckClass,
+						CostAttributionLabels: deserializeCals(calKey),
+					}
+					clTele[calKey] = calTele
+				}
+				calTele.Executions++
+				calTele.Duration += float32(e.Duration.Seconds())
 			}
 
 			// verify
 			for tenantID, expTTele := range regionTele {
 				gotTTele, ok := p.telemetry[tenantID]
-				require.True(t, ok, "telemetry not found for tenant")
-				for _, expCCTele := range expTTele {
-					gotCCTele, ok := gotTTele[expCCTele.CheckClass]
-
-					require.True(t, ok, "cal telemetry not found for tenant")
-					for _, gotCalTele := range gotCCTele {
-						require.True(t, ok, "telemetry not found for check class")
-						require.Equal(t, expCCTele.Executions, gotCalTele.Executions)
-						require.Equal(t, expCCTele.Duration, gotCalTele.Duration)
+				require.True(t, ok, "telemetry not found for tenant %d", tenantID)
+				for checkClass, expCLTele := range expTTele {
+					gotCLTele, ok := gotTTele[checkClass]
+					require.True(t, ok, "telemetry not found for check class %v", checkClass)
+					for calKey, expCalTele := range expCLTele {
+						gotCalTele, ok := gotCLTele[calKey]
+						require.True(t, ok, "telemetry not found for cal key %s", calKey)
+						require.Equal(t, expCalTele.Executions, gotCalTele.Executions, "executions mismatch for cal key %s", calKey)
+						require.Equal(t, expCalTele.Duration, gotCalTele.Duration, "duration mismatch for cal key %s", calKey)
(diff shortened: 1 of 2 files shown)


internal/scraper/scraper.go
```diff
@@ -604,6 +605,33 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
...
+		if idx >= 0 {
+			value = s.check.Labels[idx].Value
+		}
+		vals = append(vals, sm.CostAttributionLabel{
+			Name:  cal,
+			Value: value,
+		})
+	}
+	return vals, nil
+}
```
Pokom, line 634: There is a potential for this to be a performance bottleneck. This gets called every scrape, so we should consider precomputing this once the check comes online.

The-9880: The reason this is called every scrape is because the tenant data could've been updated with new CALs, right?

Every scraper for this tenant's check would independently compute this, per scrape.

Maybe the `tenantProvider` could instead cache a wrapper for `sm.Tenant` which includes a computed field for`[]sm.CostAttributionLabel` on it? Then you can derive this data once per tenant update and continue to fetch it from the cache (where it may have been updated).

internal/telemetry/region_pusher.go
```diff
@@ -197,19 +212,63 @@ func (p *RegionPusher) next() sm.RegionTelemetry {
...
 
+func serializeCALs(cals []sm.CostAttributionLabel) string {
+	if len(cals) == 0 || cals == nil {
+		return CalNilStringTerminator
+	}
+
+	sorted := make([]sm.CostAttributionLabel, len(cals))
+	copy(sorted, cals)
+	slices.SortFunc(sorted, func(a, b sm.CostAttributionLabel) int {
+		if a.Name != b.Name {
```
The-9880: Is it possible to have CALs with the same `Name`? I get an error when trying to set that up in CMAB:
```
denied the request: label names must be unique for each dimension
```

Also: would it be simpler to have the API provide a guarantee that CALs will be sorted by name?

internal/telemetry/region_pusher.go
```diff
@@ -197,19 +212,63 @@ func (p *RegionPusher) next() sm.RegionTelemetry {
...
+			}
 		}
 		m.Telemetry = append(m.Telemetry, tenantTele)
 	}
 
 	return m
 }
 
+func serializeCALs(cals []sm.CostAttributionLabel) string {
+	if len(cals) == 0 || cals == nil {
```
The-9880, line 233: I believe `len(slice)` returns `0` anyways if the slice is `nil`.

internal/telemetry/region_pusher.go
```diff
@@ -197,19 +212,63 @@ func (p *RegionPusher) next() sm.RegionTelemetry {
...
-			})
+			for calKey, calTele := range clTele {
+				tenantTele.Telemetry = append(tenantTele.Telemetry, &sm.CheckClassTelemetry{
+					CheckClass:        calTele.CheckClass,
+					Executions:        calTele.Executions,
+					Duration:          calTele.Duration,
+					SampledExecutions: calTele.SampledExecutions,
+					// deserializeCals converts the key(string) into a slice of cost attribution labels
+					// This mitigates the need for the API to perform any operations on the cal's
+					CostAttributionLabels: deserializeCals(calKey),
```
The-9880: Why not use the value that is already on `calTele`:

```suggestion
					CostAttributionLabels: calTele.CostAttributionLabels,
```

The only difference I see is that `deserializeCals()` might sort them differently - would that make a difference in the API?

internal/telemetry/telemeter_test.go
```diff
@@ -45,9 +45,13 @@ func TestTelemeterAddExecution(t *testing.T) {
...
 					gotCCTele, ok := gotTTele[expCCTele.CheckClass]
-					require.True(t, ok, "telemetry not found for check class")
-					require.Equal(t, expCCTele.Executions, gotCCTele.Executions)
-					require.Equal(t, expCCTele.Duration, gotCCTele.Duration)
+
+					require.True(t, ok, "cal telemetry not found for tenant")
+					for _, gotCalTele := range gotCCTele {
+						require.True(t, ok, "telemetry not found for check class")
+						require.Equal(t, expCCTele.Executions, gotCalTele.Executions)
+						require.Equal(t, expCCTele.Duration, gotCalTele.Duration)
```
The-9880: Something for `CostAttributionLabels` would be nice here - at least a non-nil check?

internal/scraper/scraper.go
Pokom: > The reason this is called every scrape is because the tenant data could've been updated with new CALs, right?

Correct

> Maybe the tenantProvider could instead cache a wrapper for sm.Tenant which includes a computed field for[]sm.CostAttributionLabel on it? Then you can derive this data once per tenant update and continue to fetch it from the cache (where it may have been updated).

That's an interesting idea! I would need to think through how to modify this method to do that, but I think that could be done.

The other thing to consider is that the checks are effectively immutable. Whenever a check gets created/updated/deleted, the Updater(https://github.com/grafana/synthetic-monitoring-agent/blob/feat/track-executions-by-cals/internal/checks/checks.go#L852-L852) will delete the existing scraper and create a new one. 

So another way we could solve this is to compute the value of the cal's once on initiating. I _looked_ for an easy place to put this, but the code tree is a bit complex. I'm likely going to get this into dev and measure performance before opting to solve it.

internal/telemetry/region_pusher.go
Pokom: That's fair. This was more defensive programming, but I believe the way the database is setup today we can not store two cal's with the same name. I'll remove the comparison. 

> Also: would it be simpler to have the API provide a guarantee that CALs will be sorted by name?

Yeah that would probably be a nice to have.

Pokom: I'll leave this in place until we get the API in place to return them sorted.

internal/telemetry/region_pusher.go
Pokom: Ah very close to being able to do that, but the tests fail with that change. Let me figure out why and that removes the need for us to call deserializeCals()

Pokom: Arg yeah it's because the cals aren't sorted heh. I can simplify this a bit but not remove the deserializeCals call. I'll push up a commit soon and retag a review, then I can look at the API side

internal/telemetry/telemeter_test.go
Pokom: Oh really good catch! That entire test block needed to be updated to support cals 😬

PR #1649, opened 2025-12-26
fix: prober/multihttp: fix updated headers being carried over

nadiamoe:
We've got a report where, for multihttp checks, http headers set by users for one request are being carried over to the next request. After some debugging, I think I've found the root cause: A variable is being reused in the `augmentHttpHeaders`'s function main loop, causing headers from being copied over.

This PR adds some tests for `augmentHttpHeaders` and fixes this behavior.


Pokom:
@nadiamoe I'm going to close this out for now until we can prioritize a long term fix.

PR #1689, opened 2026-01-28
k6runner: propagate k6 version manifest to remote runners

nadiamoe:
This PR adds the plumbing bits that propagate the k6 channel manifest down the call stack to runners. For now, the local runner does not do anything with this information (that will come in a separate PR), but the HTTP runner does propagate this information.

Part of: https://github.com/grafana/synthetic-monitoring/issues/416

### DIFF
--- a/internal/k6runner/http.go
+++ b/internal/k6runner/http.go
@@ -69,6 +69,11 @@ func (r HttpRunner) Run(ctx context.Context, script Script, secretStore SecretSt
 		panic("zero backoff, runner is misconfigured, refusing to DoS")
 	}
 
+	// FIXME: Make this an error after we fully roll out k6 versioning.
+	if script.K6ChannelManifest == "" || script.K6ChannelManifest == "*" {
+		r.logger.Debug().Msg("Check does not define a channel, latest k6 version will be used and it may change without warning")
+	}
+
 	if deadline, hasDeadline := ctx.Deadline(); !hasDeadline {
 		defaultAllRetriesTimeout := time.Duration(script.Settings.Timeout) * time.Millisecond * 2
 		r.logger.Error().
--- a/internal/prober/browser/browser.go
+++ b/internal/prober/browser/browser.go
@@ -45,6 +45,18 @@ func NewProber(ctx context.Context, check model.Check, logger zerolog.Logger, ru
 				Timeout: check.Timeout,
 			},
 			CheckInfo: k6runner.CheckInfoFromSM(check),
+			K6ChannelManifest: func(ch *sm.Channels) string {
+				if ch == nil || ch.K6 == nil {
+					checkInfo := k6runner.CheckInfoFromSM(check)
+					logger.Warn().
+						Object("checkInfo", &checkInfo).
+						Msg("Channel manifest unexpectedly empty for check. Please select a version for it.")
+
+					return ""
+				}
+
+				return ch.K6.Manifest
+			}(check.Channels),
 		},
 	}
 
--- a/internal/prober/browser/browser.go
+++ b/internal/prober/browser/browser.go
@@ -45,20 +45,21 @@ func NewProber(ctx context.Context, check model.Check, logger zerolog.Logger, ru
 				Timeout: check.Timeout,
 			},
 			CheckInfo: k6runner.CheckInfoFromSM(check),
-			K6ChannelManifest: func(ch *sm.Channels) string {
-				if ch == nil || ch.K6 == nil {
-					checkInfo := k6runner.CheckInfoFromSM(check)
-					logger.Warn().
-						Object("checkInfo", &checkInfo).
-						Msg("Channel manifest unexpectedly empty for check. Please select a version for it.")
-
-					return ""
-				}
-
-				return ch.K6.Manifest
-			}(check.Channels),
 		},
 	}
+
+	// Set channel manifest to empty string handling potential nils in the check model.
+	p.module.Script.K6ChannelManifest = func() string {
+		if check.Channels == nil || check.Channels.K6 == nil {
+			logger.Warn().
+				Object("checkInfo", &p.module.Script.CheckInfo).
+				Msg("Channel manifest unexpectedly empty for check. Please select a version for it.")
+
+			return ""
+		}
+
+		return check.Channels.K6.Manifest
+	}()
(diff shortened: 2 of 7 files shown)


internal/k6runner/http.go
```diff
@@ -69,6 +69,11 @@ func (r HttpRunner) Run(ctx context.Context, script Script, secretStore SecretSt
 		panic("zero backoff, runner is misconfigured, refusing to DoS")
 	}
 
+	// FIXME: Make this an error after we fully roll out k6 versioning.
+	if script.K6ChannelManifest == "" || script.K6ChannelManifest == "*" {
+		r.logger.Debug().Msg("Check does not define a channel, latest k6 version will be used and it may change without warning")
```
mem, line 74: this should probably be a warning. If that's the case, this is going to be noisy, which is a good indication that this belongs elsewhere. In the prober maybe?

nadiamoe: That will definitely be a warning (or error, as the comment says) once all DB-side migrations are done and all checks are assigned to v1 in the database. But for now that will be incredibly noisy, that's why I set it to debug on this first PR.

nadiamoe: Welp, it seems like I actually already added that to the prober? https://github.com/grafana/synthetic-monitoring-agent/blob/2f16cdf959b26c22558ebd0dd58a5fa40eb3871b/internal/prober/scripted/scripted.go#L51-L53

internal/prober/browser/browser.go
```diff
@@ -45,6 +45,18 @@ func NewProber(ctx context.Context, check model.Check, logger zerolog.Logger, ru
 				Timeout: check.Timeout,
 			},
 			CheckInfo: k6runner.CheckInfoFromSM(check),
+			K6ChannelManifest: func(ch *sm.Channels) string {
+				if ch == nil || ch.K6 == nil {
+					checkInfo := k6runner.CheckInfoFromSM(check)
+					logger.Warn().
+						Object("checkInfo", &checkInfo).
+						Msg("Channel manifest unexpectedly empty for check. Please select a version for it.")
```
nadiamoe: @ka3de Could use your insight here: Will this warning trigger for all checks if we merge this PR now? Trying to figure out if I should downgrade this to debug for the time being.

internal/k6runner/http.go
mem: OK, makes sense.

internal/prober/multihttp/multihttp.go
```diff
@@ -52,14 +52,17 @@ func NewProber(ctx context.Context, check model.Check, logger zerolog.Logger, ru
 		return p, err
 	}
 
+	const multiHTTPManifest = "^1.0.0" // From v1 and up to, but _not_ including, v2.
```
ka3de, line 55: Correlating this with how the k6 channel is parsed from the runner [here](https://github.com/grafana/synthetic-monitoring-agent/blob/2f16cdf959b26c22558ebd0dd58a5fa40eb3871b/internal/k6runner/http.go#L72-L75).
With the current definition it obligates us to remember this implementation and change it, probably when v2 is released, and as mandatory when v1 is stopped being supported.

Is there a safer approach that we could take? Could we just pass `*` so the runner uses the latest (default) version available? In that case the problem is that the runner does not distinguish between check types, so as of now we would log the warning from [here](https://github.com/grafana/synthetic-monitoring-agent/blob/2f16cdf959b26c22558ebd0dd58a5fa40eb3871b/internal/k6runner/http.go#L72-L75).

What are your thoughts on this?

internal/prober/browser/browser.go
ka3de: No. As of now, every scripted and browser check should be associated with the k6 channel `v1`, and that should already be propagated to agents.

internal/prober/browser/browser.go
```diff
@@ -45,6 +45,18 @@ func NewProber(ctx context.Context, check model.Check, logger zerolog.Logger, ru
 				Timeout: check.Timeout,
 			},
 			CheckInfo: k6runner.CheckInfoFromSM(check),
+			K6ChannelManifest: func(ch *sm.Channels) string {
+				if ch == nil || ch.K6 == nil {
+					checkInfo := k6runner.CheckInfoFromSM(check)
```
ka3de: nit: We could avoid this allocation by initializing `K6ChannelManifest` after the initialization of `p.module`, reusing `CheckInfo`.

<details><summary>Something like this</summary>

```go
  p.module = Module{
		Prober: sm.CheckTypeBrowser.String(),
		Script: k6runner.Script{
			Script: check.Settings.Browser.Script,
			Settings: k6runner.Settings{
				Timeout: check.Timeout,
			},
			CheckInfo: k6runner.CheckInfoFromSM(check),
		},
	}

	p.module.Script.K6ChannelManifest = func(ch *sm.Channels, checkInfo k6runner.CheckInfo) string {
		if ch == nil || ch.K6 == nil {
			logger.Warn().
				Object("checkInfo", &checkInfo).
				Msg("Channel manifest unexpectedly empty for check. Please select a version for it.")

			return ""
		}

		return ch.K6.Manifest
	}(check.Channels, p.module.Script.CheckInfo)
```

</details>

ka3de (commented):
LGTM. I mainly just have a comment on how are we gonna handle the multihttp k6 channel manifest that is passed to the runner.

internal/prober/multihttp/multihttp.go
nadiamoe: I don't have a problem sending `*`, which can be seen as safer in the sense of "we cannot forget to change it". However, I think it is also less safe in terms of multihttp becoming less deterministic. If we add a `v2` beta binary to the runner, then that will be automatically used by all multihttp checks in existence, and our multihttp script generator may not be compatible with that version.

So, with `*` we're effectively trading the risk of forgetting to update this, with the risk of forgetting to test multihttp scripts in new runner versions.

ka3de: We have decided to keep `v1` by now. We'll implement an e2e test in order catch the version mismatch in case it is produced. It's probably not the ideal solution, but other options have other drawbacks or require a more complex implementation.

ka3de (approved):

PR #1749, opened 2026-03-25
poc(metamonitoring): Publish internal metrics to remote

Pokom:
This is a proof of concept to demonstrate a way of publishing internal metrics from the agent to the tenants remote endpoints. This builds another gatherer called `metametrics` that periodically scrapes the internal registry and converts the metric families into a format that can be sent over the wire to a prometheus backend. There's still quite a bit of work that needs to be done, but in the simplest form this allowed me to publish metrics to a personal stack.


cmd/synthetic-monitoring-agent/main.go
```diff
@@ -391,6 +393,20 @@ func run(args []string, stdout io.Writer) error {
...
+			Publisher: publisher,
+			TenantID:  1, // TODO: set to actual tenant ID
+		})
+		g.Go(func() error {
+			return metaMetricsHandler.Run()
+		})
+		if err != nil {
+			return fmt.Errorf("connot create metamonitoring handler: %s", err)
+		}
+	}
```
Pokom: Gotta refactor this jsut a bit. I'm not convinced we want to halt the launch of the agent if this handler fails to launch. Also it's weird that the return is after launching the go routine.

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+				ts = append(ts, prompb.TimeSeries{
+					Labels:  labels,
+					Samples: []prompb.Sample{{Timestamp: stamp, Value: *value}},
+				})
+			}
+		}
+	}
+
+	return ts
+}
```
Pokom: This is a copied from the publisher codebase. I need to figure out if I want to share the code, _and_ figure out how to handle histograms/summaries which might be more complex.

internal/metametrics/metametrics.go
```diff
@@ -0,0 +1,120 @@
...
+	return &metaMetricsHandler{
+		logger:    opts.Logger,
+		registry:  opts.Registry,
+		publisher: opts.Publisher,
+		tenantID:  opts.TenantID,
+	}, nil
+}
+
+func (m metaMetricsHandler) Run() error {
+	ticker := time.NewTicker(15 * time.Second)
```
Pokom: This should be configurable, with the default being 1 minute. This would be 4DPM, which would incur a higher cost then most customers would expect.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -391,6 +397,29 @@ func run(args []string, stdout io.Writer) error {
...
 	})
 
+	if config.PushMetrics {
+		g.Go(func() error {
+			tenantID, err := tm.WaitForTenant(ctx)
+			if err != nil {
+				return err
+			}
+
+			zl.Info().Int64("tenantID", tenantID).Msg("metamonitoring: tenant discovered, starting metrics handler")
```
Pokom: I've tested this mechanism against a dev and produciton personal stack and validated that the proper tenant is found and metrics are published to it.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -97,6 +101,8 @@ func run(args []string, stdout io.Writer) error {
 			CacheType:          cache.KindAuto,
 			CacheLocalCapacity: 10000,
 			CacheLocalTTL:      5 * time.Minute,
+			PushMetrics:        true,
+			MetricsInterval:    time.Minute,
```
Pokom, line 103: Defaulting to 1 minute so that that DPM is 1, which aligns with recommended practices for Grafana Cloud. 4 DPM could cause an unexpected spike in costs.

internal/tenants/manager.go
```diff
@@ -178,7 +183,25 @@ func (tm *Manager) calculateValidUntil(tenant *sm.Tenant) time.Time {
...
+		return tm.firstTenantID, nil
+	case <-ctx.Done():
+		return 0, ctx.Err()
+	}
+}
+
 func (tm *Manager) updateTenant(tenant sm.Tenant) {
+	tm.firstOnce.Do(func() {
+		tm.firstTenantID = tenant.Id
+		close(tm.firstTenantCh)
```
Pokom: Close the channel so that it does not block. We only need to read once the tenantID and then we can assume it doesn't change.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -97,6 +101,8 @@ func run(args []string, stdout io.Writer) error {
 			CacheType:          cache.KindAuto,
 			CacheLocalCapacity: 10000,
 			CacheLocalTTL:      5 * time.Minute,
+			PushMetrics:        true,
```
mem: In order to give us a chance to test this internally, I would prefer to have this default to false and register a command line flag to enable it (label it "experimental").

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+
+type metricsHandler struct {
+	logger    zerolog.Logger
+	registry  *prometheus.Registry
+	publisher pusher.Publisher
+	tenantID  model.GlobalID
+	interval  time.Duration
+}
+
+func NewHandler(opts HandlerOpts) (*metricsHandler, error) {
```
mem: there's no way for this to fail, remove the error from the return values.

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case t := <-ticker.C:
+			if err := m.reportUsage(); err != nil {
+				m.logger.Error().Err(err).Time("t", t).Msg("failed to report metrics")
+			} else {
+				m.logger.Info().Time("t", t).Msg("reported metrics")
```
mem: having this every minute is a little spammy.

Add an entry outside the loop that says it's reporting metrics and log errors.

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	Logger    zerolog.Logger
+	Registry  *prometheus.Registry
+	Publisher pusher.Publisher
+	TenantID  model.GlobalID
+	Interval  time.Duration
+}
+
+type metricsHandler struct {
+	logger    zerolog.Logger
+	registry  *prometheus.Registry
```
mem: you normally want a `prometheus.Gatherer` or a `prometheus.Registerer` here, not the concrete type.

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+
+	m.publisher.Publish(&payload{
+		tenantID: m.tenantID,
+		metrics:  ts,
+	})
+
+	return nil
+}
+
+func mfsToTimeseries(t time.Time, mfs []*dto.MetricFamily) []prompb.TimeSeries {
```
mem: can you please add a comment around here along the lines of:

```go
// Histograms and summaries are intentionally skipped: they decompose
// into multiple series (_bucket, _sum, _count) that require special
// handling. Only scalar metric types (counter, gauge, untyped) are
// forwarded as-is.
```

internal/metamonitoring/metrics.go
mem: Also, while this pattern exists in some parts of the code, returning unexported types is problematic.

Either export a type or define a small interface that this implements and report that, e.g.:

```go
type Handler interface {
    Run(ctx context.Context) error
}

File: internal/tenants/manager.go [hunk 12]
```diff
@@ -33,6 +33,10 @@ type Manager struct {
 	cache         cache.Cache
 	fetchMutexes  *xsync.Map[int64, *sync.Mutex] // for fetch deduplication.
 	logger        zerolog.Logger
+
+	firstTenantID int64
+	firstTenantCh chan struct{} // closed when the first tenant is observed
```
mem: isn't it simpler to just send the tenant ID over the channel?

internal/tenants/manager_test.go
```diff
@@ -173,6 +173,108 @@ func TestTenantManagerGetTenant(t *testing.T) {
...
+		})
+		require.NoError(t, err)
+		defer localCache.Close()
+
+		tc := &testTenantsClient{
+			tenants:      map[int64]sm.Tenant{},
+			requestCount: make(map[int64]int),
+		}
+
+		tm := NewManager(ctx, tc, tenantCh, DefaultCacheTimeout, localCache, logger)
```
mem: this bit here is duplicated across all tests, maybe move it to a separate inline function?

internal/tenants/manager_test.go
```diff
@@ -173,6 +173,108 @@ func TestTenantManagerGetTenant(t *testing.T) {
 	require.Equal(t, t3, *tenant)
 }
 
+func TestWaitForTenant(t *testing.T) {
```
mem: I understand it's hard to test (I would expect not so much, but I haven't looked at this closely)...

This is not testing the `GetTenant` code path. If testing is not possible, add a comment as to why.

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -391,6 +397,29 @@ func run(args []string, stdout io.Writer) error {
 		return adhocHandler.Run(ctx)
 	})
 
+	if config.PushMetrics {
```
mem: the bit that makes me uneasy here is that this waits for whatever tenant comes first. In a multitenant deployment, that will be whatever tenant and internal agent metrics will get published there.

internal/metamonitoring/metrics_test.go
```diff
@@ -0,0 +1,344 @@
...
+		handler, err := NewHandler(HandlerOpts{
+			Logger:    zerolog.New(zerolog.NewTestWriter(t)),
+			Registry:  registry,
+			Publisher: pub,
+			TenantID:  1,
+			Interval:  50 * time.Millisecond,
+		})
+		require.NoError(t, err)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 175*time.Millisecond)
```
mem: the combination of 50 ms interval plus 175 ms timeout is what causes the expect between 2 and 4 elements in the payload.

That should be documented.

Ideally, that should be removed. Is it possible to use `testing/synctest` here?

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -366,7 +373,6 @@ func run(args []string, stdout io.Writer) error {
 	if err != nil {
 		return fmt.Errorf("cannot create checks updater: %w", err)
 	}
-
```
mem, line 369: this is unrelated to this change

mem (changes requested):
This looks good. I would prefer to solve the `PushMetrics: true` bit before merging to avoid accidents.

internal/metamonitoring/metrics.go
mem: Ah, that explains the lack of histogram / summary support.

The publisher code never sees histograms / summaries because those are a collection of counters.

internal/tenants/manager.go
mem: This is well defined:

```go
package main

import (
        "fmt"
        "sync"
)

func main() {
        var wg sync.WaitGroup

        ch := make(chan int)

        wg.Add(1)
        go publisher(&wg, ch)

        wg.Add(1)
        go consumer(&wg, ch)

        wg.Wait()
}

func publisher(wg *sync.WaitGroup, ch chan<- int) {
        defer wg.Done()

        ch <- 42
        close(ch)
}

func consumer(wg *sync.WaitGroup, ch <-chan int) {
        defer wg.Done()

        v := <-ch
        fmt.Println("Got", v)
}
```

Pokom: I'm trying to thread this comment with your comment later

> the bit that makes me uneasy here is that this waits for whatever tenant comes first. In a multitenant deployment, that will be whatever tenant and internal agent metrics will get published there.

Two questions:
1. is there a scenario where a private probe would be running in a multitenant deployment? Or is this concern more about general safety to ensure it could never happen
2. Is the suggestion have `firstTenantCh chan int` that only communicates the tenant ID?

Both math sense, just trying to make sure I understand the full concern and address it. If the first is about safety, I think a better alternative would be to communicate the tenant ID when register probe is called(https://github.com/grafana/synthetic-monitoring-agent/blob/poc/metamonitoring-setup/internal/checks/checks.go#L381-L381). Let me know if that makes sense!

cmd/synthetic-monitoring-agent/main.go
```diff
@@ -391,6 +400,19 @@ func run(args []string, stdout io.Writer) error {
...
+		g.Go(func() error {
+			metricsHandler := metamonitoring.NewHandler(metamonitoring.HandlerOpts{
+				Logger:        zl.With().Str("subsystem", "metamonitoring").Logger(),
+				Registry:      promRegisterer,
+				Publisher:     publisher,
+				Interval:      config.MetricsInterval,
+				ProbeTenantCh: probeTenantCh,
+			})
+
+			return metricsHandler.Run(ctx)
```
Pokom, line 412: I've moved over the logic to fetching the tenant ID into the metamonitoring component. We still have the ability to pass one in with `HandlerOpts` in case we want to support the user defining this value ahead of time(think sending data to a different tenants stack).

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil
+	case <-ctx.Done():
+		return errTenantTimeout
+	}
+}
+
+func (m *metricsHandler) reportUsage() error {
+	mfs, err := m.registry.Gather()
+	if err != nil {
+		return err
```
mem, line 99: I'm not sure this is correct. `Gather` will return an error *and* possibly some data. It should only happen when there are metrics with errors, which should never happen in this case. Add comment?

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+	now := time.Now()
+	ts := mfsToTimeseries(now, mfs)
+	if len(ts) == 0 {
+		return nil
+	}
+
+	m.publisher.Publish(&payload{
+		tenantID: model.GlobalID(m.tenantID),
+		metrics:  ts,
+	})
```
mem, line 111: I'm munching over this...

When *not* running behind the proxy, this is OK. When running behind the proxy, the proxy will try to extract the region from the tenant ID it sees so that it knows which API to contact for `GetTenant`.

Because this is a private probe, it will never be behind the proxy today. In the future this might or might not be true. Basically the question is what is being sent over the channel: a global ID or a local ID? If it's a local one, why is it not a global one?

internal/metamonitoring/metrics.go
mem: I think what you need to do here is log the error, and process mfs if `len(mfs) > 0`.

internal/metamonitoring/metrics.go
mem: The cast is necessary because the channel is not passing a `model.GlobalID`. The tenant ID received from the API *should be* a Global ID: if it comes from the proxy, it is a GlobalID; if it comes directly from the API, it's local. Casting to a `model.GlobalID` should happen on the other side of the channel unless I'm missing something.

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func (m *metricsHandler) waitForTenantID(ctx context.Context) error {
+	select {
+	case tenantID := <-m.probeTenantCh:
+		m.tenantID = tenantID
+		return nil
+	case <-ctx.Done():
+		return errTenantTimeout
```
mem: The context has been cancelled. There should be a `ctx.Err()` with the cancellation reason. Possibly annotate that: `fmt.Errorf("waiting for probe tenant ID: %w", ctx.Err())`.

mem: Or simply shut down and return `nil`. Returning `nil` is bad because `Run` is blocked waiting for this to finish and if it gets `nil`. It will exit itself because the context is cancelled on line L73.

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func (m *metricsHandler) reportUsage() error {
```
mem, line 96: I've been mulling over this.

Note there's one issue that needs addressing: internal metrics have _some_ labels, but notably they do not have labels to separate different instances of the agent. Consider [this](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22v8q%22:%7B%22datasource%22:%22grafanacloud-prom%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sm_agent_api_connection_status%7Bcluster%3D%5C%22pop-prod-aws-calgary-1%5C%22,%20namespace%3D%5C%22sm-proxy%5C%22,%20container%3D%5C%22agent%5C%22%7D%22,%22range%22:false,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22grafanacloud-prom%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22exemplar%22:false,%22format%22:%22table%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&orgId=1&var-cluster=prod-us-central-0). `(cluster, pod, instance)` is keeping these metrics from overwriting each other. In our case `cluster` is a 1:1 relationship with the probe's name. If we didn't have those labels, metrics from different locations would clash.

If I'm following the code correctly, because the internal metrics are expected to be scraped, the scraper is adding those labels. Without scraping, no extra labels are added. Basically, we have to add this labels. In the normal path, that happens after building the timeseries IIRC.

internal/checks/checks.go
```diff
@@ -528,6 +534,17 @@ func (c *Updater) validateProbeCapabilities(capabilities *sm.Probe_Capabilities)
 	return nil
 }
 
+// notifyProbeTenant will the probes tenantID to the probeTenantCh at most once and then close the channel.
```
mem: will _something missing_ the probes

notify?

internal/metamonitoring/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+	var ts []prompb.TimeSeries
+
+	for _, mf := range mfs {
+		name := mf.GetName()
+		mType := mf.GetType()
+
+		for _, metric := range mf.GetMetric() {
+			ml := metric.GetLabel()
+			labels := make([]prompb.Label, 0, 1+len(ml))
+			labels = append(labels, prompb.Label{Name: "__name__", Value: name})
```
mem, line 127: this might be a good place to add the extra identifying labels.

internal/metamonitoring/metrics.go
Pokom: That is true, but the comment from the Gather method says we should almost always want to discard metrics in that case:

```go
//   ...it is almost always better to not
	// expose an incomplete result and instead disregard the returned
	// MetricFamily protobufs in case the returned error is non-nil.
```

Let me think on this a bit and decide what to do.

internal/metamonitoring/metrics.go
Pokom: > Casting to a model.GlobalID should happen on the other side of the channel unless I'm missing something.

Which channel are thinking the cast should happen on? Are thinking in the metamonitoring package when we receive the `int64` tenantID here: https://github.com/grafana/synthetic-monitoring-agent/blob/poc/metamonitoring-setup/internal/metamonitoring/metrics.go#L84-L86

I think that's reasonable. I believe I that's the way the metrics package was setup at first and I changed it to an int64 with the refactor.

internal/metamonitoring/metrics.go
Pokom: Yeah this is completely fair and a bit of an oversight for the metrics. I covered this in my PR for logs(https://github.com/grafana/synthetic-monitoring-agent/pull/1759/changes#diff-5c2ea9d477a93115f8fbe86695012cf1e847c53eaf30efdd179b2ec6819eeb2eR429). The main change we'll need to make is what is sent to metamonitoring after the RegisterProbe call finished. Currently we're sending just the tenantID, we'll need to send the tenantID and probe name. Any objections on broadcasting the probe struct so that we're creating an intermediary type to communicate the data?

mem (approved):
Let's go with this.

PR #1771, opened 2026-04-13
feat: Add support for xk6-sm v2

ka3de:
Updates build and release files in order to include xk6-sm v2 binary bundled within the same directory as v1.
Also updates renovate configuration in order to keep both dependencies up to date.

Depends on: grafana/xk6-sm/pull/397
Updates grafana/synthetic-monitoring/issues/491

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes build/release packaging and Docker image contents by adding a second `k6` binary, which can break artifacts or image size/paths if misconfigured; runtime probe logic is largely unchanged (multihttp still pins v1).
> 
> **Overview**
> Adds bundling of **xk6-sm v2** alongside v1 across release artifacts and container images.
> 
> Updates Dockerfiles, `goreleaser`/`nfpm`, and packaging Make targets to download/build and ship both `/usr/libexec/sm-k6/k6-v1` and `k6-v2` (and include `k6-v2` in archives/tarballs). Renovate config is split into two tracked deps (`grafana/xk6-sm-v1` and `grafana/xk6-sm-v2`) with version ranges and regex managers that update `K6_V1_VERSION`/`K6_V2_VERSION` variables.
> 
> Tests are adjusted to skip running multiHTTP probes against the v2 binary since multiHTTP is currently forced to use v1.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23893e94a0c2d6506aa59a4f5e27e95e0c2fbd5c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/renovate.yaml
+++ b/.github/renovate.yaml
@@ -23,10 +23,18 @@ packageRules:
       - "ghcr.io/grafana/chromium-swiftshader-alpine"
     versioning: loose
 
+  - matchDepNames:
+      - "grafana/xk6-sm-v1"
+    allowedVersions: "<2.0.0"
+
+  - matchDepNames:
+      - "grafana/xk6-sm-v2"
+    allowedVersions: ">=2.0.0 <3.0.0"
+
 customManagers:
-  - # Update xk6-sm
-    customType: "regex"
-    depNameTemplate: "grafana/xk6-sm"
+  - customType: "regex"
+    depNameTemplate: "grafana/xk6-sm-v1"
+    packageNameTemplate: "grafana/xk6-sm"
     datasourceTemplate: "github-releases"
     versioningTemplate: "semver"
     managerFilePatterns:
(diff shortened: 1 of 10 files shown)


ka3de:
@cursor bugbot review please

.github/renovate.yaml
```diff
@@ -23,15 +23,35 @@ packageRules:
...
+    depNameTemplate: "grafana/xk6-sm-v1"
+    packageNameTemplate: "grafana/xk6-sm"
+    datasourceTemplate: "github-releases"
+    versioningTemplate: "semver"
+    managerFilePatterns:
+      - "**/Dockerfile"
+      - "**/Dockerfile.*"
+      - "**/*.mk"
+    matchStrings:
+      - "https://github.com/grafana/xk6-sm/releases/download/(?<currentValue>[^/]+)/[^\\s]+\\s.*k6-v1"
```
mem: this is easier if you introduce a variable in the makefile that sets the k6 version we want:

```make
K6_V1_VERSION := v1.1.0
```

and then your match string is something like:

```yaml
"K6_V1_VERSION := (?<currentValue>v1\\.[0-9]+\\.[0-9]+)"
```

adjust the regular expression as needed.

And then in the other part of the makefile you just have:

```make
curl -sSL "https://github.com/grafana/xk6-sm/releases/download/$(K6_V1_VERSION)/sm-k6-$(1)-$(2)" -o "$$@"
```

mem (commented):
other than the minor tweak to injecting the k6 versions, this looks good

.github/renovate.yaml
ka3de: The problem is that the current match string is also used for the Dockerfiles. So if we want to make this change, we either have to have separate definitions for each case in renovate (because the `K6_V1_VERSION := ..` will not apply for the Dockerfiles), or we need to apply other changes on how the version is passed to the Dockerfiles.

I understand the inline URL with the specific version might not be ideal, but it allows us to have a single pattern and single match string in renovate.

mem (dismissed):

mem (dismissed):

mem (approved):

Pokom (approved):

PR #1782, opened 2026-04-20
feat: Unit + Integration tests for check execution metrics

The-9880:
Relates to https://github.com/grafana/synthetic-monitoring/issues/539

## Summary

This PR adds an in-repo metric catalogue contract for the scraper and surfaces an initially non-blocking runtime validation in CI.

The intent is to catch metric/label drift prior to PR merge instead of solely relying on E2E validation (see https://github.com/grafana/sm-e2e/pull/81).

This currently achieves **catalogue parity** with E2E. The next step would be to attain parity with the exhaustiveness of different check configurations tested.

The gap left is that `traceroute` metrics are not validated  in CI here.

### 1. Align local scraper fixtures with the catalogue surface

The scraper fixtures now intentionally emit the metrics needed for the catalogue contract:

- DNS fixture now exercises the SOA branch so `probe_dns_serial` is covered.
- TLS HTTP fixture now emits `Last-Modified`, which surfaces `probe_http_last_modified_timestamp_seconds`.
- Browser fixture now performs a real interaction for the integration test so the interaction-driven browser web vitals are emitted.

### 2. Codify two catalogue contracts

This PR introduces two expected catalogues:

- **Fixture catalogue**
  - Protects the deterministic local scraper fixture/replay path used by the regular scraper tests.
- **Runtime catalogue**
  - Protects the richer k6/browser execution path when integration test checks run with a real `sm-k6` binary and browser-capable runtime.

Those are validated by separate tests because they intentionally represent different execution contracts.

### 3. Surface runtime drift in CI

`validate_pr.yaml` now runs the runtime catalogue integration as a **non-blocking** (initially) CI signal on `ubuntu-latest`. We will evaluate the step for a short time before making it blocking.

- The browser-enabled image is built and loaded locally in the job.
- The runtime catalogue integration runs once per CI on `ubuntu-latest`.
- The step uses `continue-on-error: true`, so it provides signal without blocking merges yet.

The regular browser build validation remains on both matrix legs.

## Commit guide

Review in commit order:

1. `fix(scraper): align local fixtures with catalogue coverage`
   - Fixture behavior changes and the minimal golden/accounting updates that follow from them.

2. `test(scraper): codify fixture and runtime metric catalogues`
   - Catalogue reduction helpers, expected catalogue files, and the fixture/runtime contract tests.

3. `ci(validate): surface runtime catalogue drift`
   - Non-blocking CI wiring for the runtime catalogue integration.

## Validation

Locally verified:

- `ROOTDIR=$PWD GOOS=linux GOARCH=amd64 ./scripts/docker-run go test ./internal/scraper`
- `./scripts/ci/run-runtime-catalogue-integration.sh` inside the browser-capable image

## Known boundaries

- `traceroute` is still excluded from this agent-side contract because the runner environment may not reliably provide the required capabilities.

---

> [!NOTE]
> **Medium Risk**
> Changes observable Prometheus metric names/labels contracts and billing-related active series counts; runtime CI is non-blocking but wrong catalogues could block future merges once enforced.
> 
> **Overview**
> Adds **metric/label catalogue contracts** for scraper check output so PRs can catch drift before merge, with a separate **fixture** path (deterministic local/replay) and **runtime** path (real `sm-k6` / browser).
> 
> **Fixture alignment** updates local probe setups and goldens so the contract covers more surface: DNS uses **SOA** (including `probe_dns_serial`), HTTP SSL serves **Last-Modified**, and scripted/multihttp/browser fixtures run richer k6 scripts (TLS, assertions, real page interaction) so catalogue tests match emitted metrics. Shared probe setup is centralized via `validateMetricSetups()`; `config_version` is a named constant in scraper label wiring.
> 
> **Tests** add catalogue extraction/compare helpers, expected JSON for fixture and runtime classes, contract tests (`TestExpectedMetricCatalogueMatchesFixtureOutputs`, opt-in generators), and `TestValidateMetricCatalogueIntegration` gated by `SM_RUNTIME_CATALOGUE_INTEGRATION` + `K6_PATH`. **Accounting** series counts bump slightly for dns/http_ssl after new metrics.
> 
> **CI** (`validate_pr.yaml`): on `ubuntu-latest` only, build/load a browser image, compile the integration test binary, and run runtime catalogue tests in Docker with **`continue-on-error: true`** (signal only for now). Traceroute remains outside the catalogue matrix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3a5c1ec7fa53855ef3724155edb397e26f051ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


The-9880:
Need to fix the integration test rigging...

The-9880:
Fixed, catalogue tests are now running 👍

internal/scraper/catalogue_helpers_test.go
mem, line 1: This is not a `_test.go` file, the package is not called `_test`.

Is this code really part of what should be built along with the rest of the code?

internal/scraper/catalogue_fixture_test.go
```diff
@@ -0,0 +1,90 @@
...
+		"multihttp_basic": {setup: setupMultiHTTPProbe, basicMetricsOnly: true},
+		"ping":            {setup: setupPingProbe, basicMetricsOnly: false},
+		"ping_basic":      {setup: setupPingProbe, basicMetricsOnly: true},
+		"scripted":        {setup: setupScriptedProbe, basicMetricsOnly: false},
+		"scripted_basic":  {setup: setupScriptedProbe, basicMetricsOnly: true},
+		"tcp":             {setup: setupTCPProbe, basicMetricsOnly: false},
+		"tcp_basic":       {setup: setupTCPProbe, basicMetricsOnly: true},
+		"tcp_ssl":         {setup: setupTCPSSLProbe, basicMetricsOnly: false},
+		"tcp_ssl_basic":   {setup: setupTCPSSLProbe, basicMetricsOnly: true},
+	}
```
mem: this list already exists in `TestValidateMetrics` in almost the same form. Having to maintain it in two places does not sound appealing.

mem: this list already exists in `TestValidateMetrics` in almost the same form. Having to maintain it in two places does not sound appealing.

internal/scraper/expected_fixture_metric_catalogue.go
mem, line 1: same issue with the name of the file

internal/scraper/expected_runtime_metric_catalogue.go
mem, line 1: same issue with the name of the file

internal/scraper/expected_fixture_metric_catalogue.go
mem: also, unless I'm missing something, I don't see what this needs to be code. It seems you can have this as YAML or JSON file that is loaded when needed.

internal/scraper/expected_runtime_metric_catalogue.go
mem: same thing with this file, this can be a YAML or JSON file.

scripts/ci/run-runtime-catalogue-integration.sh
```diff
@@ -0,0 +1,21 @@
...
+: "${GOMODCACHE:=/tmp/go-mod-cache}"
+: "${HOME:=/tmp}"
+
+export K6_PATH
+export SM_RUNTIME_CATALOGUE_INTEGRATION
+export GOCACHE
+export GOMODCACHE
+export HOME
+
+apk add --no-cache go git build-base >/dev/null
```
mem: This is a big no. Why do you need to get "random Go version" from Alpine? And where is the Alpine assumption materialized?

What's the actual need here?

scripts/ci/run-runtime-catalogue-integration.sh
```diff
@@ -0,0 +1,21 @@
...
+
+exec go test -run TestValidateMetricCatalogueIntegration -count=1 -v ./internal/scraper
```
mem: If the need to be able to run the test, you can compile the test and use that binary elsewhere.

You can compile the test with:

```
go test -c -o xxx ./internal/scraper/
```

and then you run it with:

```
./xxx -test.run TestValidateMetricCatalogueIntegration -test.count 1 -test.v
```

mem: Note also that this allows you to define and pass other flags. Look for `updateGolden` in the existing scraper_test.go file.

internal/scraper/catalogue_integration_test.go
```diff
@@ -0,0 +1,49 @@
...
+		t.Skipf("set %s=1 to run runtime catalogue integration tests", runtimeCatalogueIntegrationEnv)
+	}
+
+	k6Path := os.Getenv("K6_PATH")
+	if k6Path == "" {
+		t.Fatalf("K6_PATH must be set when %s is enabled", runtimeCatalogueIntegrationEnv)
+	}
+	if _, err := os.Stat(k6Path); err != nil {
+		t.Fatalf("K6_PATH %q is not usable: %v", k6Path, err)
+	}
```
mem, line 25: this guard is OK

internal/scraper/expected_fixture_metric_catalogue_generate_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"sort"
+	"strings"
+	"testing"
+)
+
+const generateFixtureCatalogueEnv = "SM_GENERATE_FIXTURE_CATALOGUE"
+
+// TestPrintExpectedMetricCatalogue is an opt-in generator for refreshing the
+// fixture catalogue from the current deterministic scraper fixture matrix.
+func TestPrintExpectedMetricCatalogue(t *testing.T) {
```
mem: Why is the guard from https://github.com/grafana/synthetic-monitoring-agent/pull/1782/changes#diff-6385e7a3a46b70e9a718175b85d5806fdfdb2beea51c46822e7679158bf84346R19-R25 not present here?

mem: My confusion comes mostly from the fact that this is not just printing something, it's generating it and then printing it. Is that correct?

.github/workflows/validate_pr.yaml
```diff
@@ -106,3 +106,37 @@ jobs:
...
+          file: Dockerfile.build
+          load: true
+
+      - name: run runtime catalogue integration tests
+        if: matrix.runner == 'ubuntu-latest'
+        # This runs the browser-capable runtime catalogue tests as a best-effort
+        # integration signal while the job shape is still being proven out.
+        continue-on-error: true
+        run: |
+          docker run --rm \
```
mem, line 142: you can build the test image (either as part of the regular build, or in a dedicated step above) and the inject the binary into the container:

```
  ...
  -e XXX=yyy \
  -v /.../catalogue-integration:/catalogue-integration \
  ...
  /catalogue-integration -test.run TestValidateMetricCatalogueIntegration -test.count 1 -test.v
```

internal/scraper/catalogue.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func CatalogueFromReader(r io.Reader) (MetricLabelCatalogue, error) {
+	dec := expfmt.NewDecoder(r, expfmt.NewFormat(expfmt.TypeTextPlain))
+	mfs := make([]*dto.MetricFamily, 0)
+
+	for {
+		mf := new(dto.MetricFamily)
+		err := dec.Decode(mf)
+		if err == io.EOF {
```
mem: nit: `errors.Is(err, io.EOF)`

mem (dismissed):

scripts/ci/run-runtime-catalogue-integration.sh
The-9880: Addressed in https://github.com/grafana/synthetic-monitoring-agent/pull/1782/changes/e9a7ad4475ddf2fb4ba826e55490f50e47b86eb5, this now uses `go` from `gbt` to build a test binary from source to execute the integration tests.

internal/scraper/expected_fixture_metric_catalogue_generate_test.go
The-9880: This doesn't use the same guard because this is the `fixture` test set.

The `fixture` tests are able to be tested locally and collect metrics by modelling different request/response patterns.

The `runtime` tests are the ones that are k6-enabled - if you look at the runtime metric catalogue, they're only k6-powered tests and require the browser-enabled image to execute and collect metrics.

The `generate` files are useful for updating the catalogues, e.g. if a diff is detected and we want to match the diff.

.github/workflows/validate_pr.yaml
The-9880: I've addressed this in the latest commit, let me know if you think it should be tweaked slightly

vmgrlab:
The probe-level metrics contract tests makes a lot of sense, since they are always emitted by the agent regardless of script content. I'm less sure about the HTTP and browser-specific metrics. The emitted metric set is really a function of script content rather than check type, on example gRPC-heavy script would produce something completely different. Am I missing something ?

.github/workflows/validate_pr.yaml
```diff
@@ -106,3 +106,44 @@ jobs:
             type=sha,prefix=sha-,suffix=-browser,format=short
             latest-browser
           target: with-browser
+
+      # Run the runtime catalogue validation once per CI on ubuntu-latest rather
+      # than repeating the same loaded-image test on both matrix legs.
+      - name: build browser image for runtime catalogue tests
+        if: matrix.runner == 'ubuntu-latest'
```
mem: what are you selecting for here? We should stop using those particular runners and move to the self-hosted ones.

You can make this more resilient by doing something like:

```
matrix:
  runner: [a, b]
  inlcude:
    - runner: a
      build_catalogue: true
```

that gives you a matrix that has:

```
- runner: a
  build_catalogue: true
- runner: b
```

and then you test for that.

internal/scraper/catalogue_fixture_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	probeImpl, check, stop := spec.setup(ctx, t)
+	defer stop()
+
+	check.Job = name + " job"
+	check.Frequency = 60000
+	check.Modified = 42
```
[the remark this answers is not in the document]
The-9880: This is fine; if the test takes long enough to cancel the context, we should revisit it.

.github/workflows/validate_pr.yaml
The-9880: Added! We can move to self-hosted runners separately, I may create an issue/PR for that as a follow-up.

mem (approved):

internal/scraper/expected_fixture_metric_catalogue_test.go
```diff
@@ -0,0 +1,33 @@
...
+		expected, ok := expectedCatalogues[class]
+		if !ok {
+			t.Fatalf("missing expected catalogue for accounting class %s", class)
+		}
+
+		observed := collectFixtureCatalogue(t, class, spec)
+		result := CompareMetricCatalogue(expected, observed)
+		if !result.Success() {
+			t.Fatalf("catalogue mismatch for %s: %s", class, result.Summary())
+		}
```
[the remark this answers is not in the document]
The-9880: Not an issue in CI, _could_ be a mild tripping hazard if you set `K6_PATH` locally - non-blocking.

The-9880:
Force-pushed to sign all commits - no diff.

vtorosyan (approved):

PR #1783, opened 2026-04-21
feat: Set k6 ref. id for browser checks run by local runner

ka3de:
Sets the `K6_CLOUD_PUSH_REF_ID` environment variable within the k6 test script runtime for browser checks, referencing the check's _job_, _instance_, and _probe_ (by now as a JSON object serialization). This variable is automatically parsed by k6 runtime and set as `window.k6` property within the web page DOM for every page visited by the browser check execution.

The purpose of this is to establish a correlation between a FE O11y browser session and Synthetic Monitoring check execution.

This PR also propagates the updated `CheckInfo`, containing the check's `job`, `instance`, and `probe` to the remote HTTP runner.

Depends on grafana/synthetic-monitoring/issues/505
Updates grafana/synthetic-monitoring/issues/506

---

> [!NOTE]
> **Medium Risk**
> Touches the `ProberFactory`/`NewProber` call chain and the k6 local runner invocation, so regressions could prevent checks (especially k6-backed ones) from starting or could change runner behavior for browser executions.
> 
> **Overview**
> Adds probe/job/instance context to k6 executions by extending `k6runner.CheckInfo` and threading a `probeName` parameter through `ProberFactory.New` and all k6-backed probers (scripted, multihttp, browser), including scraper and ad-hoc flows.
> 
> For **local** k6 browser checks, `Local.buildK6Args` now injects `K6_CLOUD_PUSH_REF_ID` (JSON containing `job`, `instance`, `probe`) via `-e`, enabling frontend observability correlation; non-browser checks keep the existing single-iteration flags. Tests are updated/added to validate the new fields and argument behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3df23f580c723869f81d6643d93306052ec9ac9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/checks/checks_test.go
+++ b/internal/checks/checks_test.go
@@ -477,7 +477,7 @@ func (testProber) Probe(ctx context.Context, target string, registry *prometheus
 
 type testProbeFactory struct{}
 
-func (f testProbeFactory) New(ctx context.Context, logger zerolog.Logger, check model.Check) (prober.Prober, string, error) {
+func (f testProbeFactory) New(ctx context.Context, logger zerolog.Logger, check model.Check, probeName string) (prober.Prober, string, error) {
 	return testProber{}, check.Target, nil
 }
 
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -60,6 +60,10 @@ type Settings struct {
 type CheckInfo struct {
 	// Type is the string representation of the check type this script belongs to (browser, scripted, multihttp, etc.)
 	Type string `json:"type"`
+	// Job is the check's job label.
+	Job string `json:"job"`
+	// Instance is the check's instance label.
+	Instance string `json:"instance"`
 	// Metadata is a collection of key/value pairs containing information about this check, such as check and tenant ID.
 	// It is loosely typed on purpose: Metadata should only be used for informational properties that will make its way
 	// into telemetry, and not for making decision on it.
@@ -73,6 +77,8 @@ func CheckInfoFromSM(smc smmodel.Check) CheckInfo {
 	}
 
 	ci.Type = smc.Type().String()
+	ci.Job = smc.Job
+	ci.Instance = smc.Target
 	ci.Metadata["id"] = smc.Id
 	ci.Metadata["tenantID"] = smc.TenantId
 	ci.Metadata["regionID"] = smc.RegionId
@@ -85,6 +91,8 @@ func CheckInfoFromSM(smc smmodel.Check) CheckInfo {
 // MarshalZerologObject implements zerolog.LogObjectMarshaler so it can be logged in a friendly way.
 func (ci *CheckInfo) MarshalZerologObject(e *zerolog.Event) {
 	e.Str("type", ci.Type)
+	e.Str("job", ci.Job)
+	e.Str("instance", ci.Instance)
 	for k, v := range ci.Metadata {
 		e.Any(k, v)
 	}
--- a/internal/k6runner/k6runner.go
+++ b/internal/k6runner/k6runner.go
@@ -64,14 +64,16 @@ type CheckInfo struct {
 	Job string `json:"job"`
 	// Instance is the check's instance label.
 	Instance string `json:"instance"`
+	// Probe is the name of the probe executing the check.
+	Probe string `json:"probe"`
 	// Metadata is a collection of key/value pairs containing information about this check, such as check and tenant ID.
 	// It is loosely typed on purpose: Metadata should only be used for informational properties that will make its way
 	// into telemetry, and not for making decision on it.
 	Metadata map[string]any `json:"metadata"`
 }
 
-// CheckInfoFromSM returns a CheckInfo from the information of the given SM check.
-func CheckInfoFromSM(smc smmodel.Check) CheckInfo {
+// CheckInfoFromSM returns a CheckInfo from the information of the given SM check and probe name.
+func CheckInfoFromSM(smc smmodel.Check, probeName string) CheckInfo {
 	ci := CheckInfo{
 		Metadata: map[string]any{},
 	}
@@ -79,6 +81,7 @@ func CheckInfoFromSM(smc smmodel.Check) CheckInfo {
 	ci.Type = smc.Type().String()
 	ci.Job = smc.Job
 	ci.Instance = smc.Target
+	ci.Probe = probeName
 	ci.Metadata["id"] = smc.Id
 	ci.Metadata["tenantID"] = smc.TenantId
 	ci.Metadata["regionID"] = smc.RegionId
@@ -93,6 +96,7 @@ func (ci *CheckInfo) MarshalZerologObject(e *zerolog.Event) {
 	e.Str("type", ci.Type)
 	e.Str("job", ci.Job)
 	e.Str("instance", ci.Instance)
+	e.Str("probe", ci.Probe)
 	for k, v := range ci.Metadata {
 		e.Any(k, v)
 	}
--- a/internal/k6runner/local.go
+++ b/internal/k6runner/local.go
@@ -18,6 +18,13 @@ import (
 	"github.com/spf13/afero"
 )
 
+// k6CloudPushRefIDEnvVar is the name of the k6 runtime environment variable used to
+// tag browser tests as originated from Synthetic Monitoring. It contains a reference
+// to the check's job, instance, and probe associated with the execution.
+// k6 will propagate this data to the window.k6 property, which is parsed by FE O11y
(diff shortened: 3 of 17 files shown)


internal/checks/checks_test.go
```diff
@@ -477,7 +477,7 @@ func (testProber) Probe(ctx context.Context, target string, registry *prometheus
 
 type testProbeFactory struct{}
 
-func (f testProbeFactory) New(ctx context.Context, logger zerolog.Logger, check model.Check) (prober.Prober, string, error) {
+func (f testProbeFactory) New(ctx context.Context, logger zerolog.Logger, check model.Check, probeName string) (prober.Prober, string, error) {
```
ka3de, line 480: I don't particularly like this addition, but I don't see any better way to do this.
This will be ignored by any prober that is not k6-based (scripted/browser/multihttp). For these, this information is propagated to the runner, but only takes effect within the k6 test run for browser checks.

internal/k6runner/k6runner.go
```diff
@@ -60,19 +60,28 @@ type Settings struct {
 type CheckInfo struct {
 	// Type is the string representation of the check type this script belongs to (browser, scripted, multihttp, etc.)
 	Type string `json:"type"`
+	// Job is the check's job label.
+	Job string `json:"job"`
+	// Instance is the check's instance label.
+	Instance string `json:"instance"`
+	// Probe is the name of the probe executing the check.
+	Probe string `json:"probe"`
```
mem, line 68: why doesn't this fit into the `metadata` bucket below?

internal/k6runner/local.go
mem, line 44: What the bot is complaining about elsewhere is this. It's not wrong.

`logger` is not guaranteed to be set, you can instantiate `Local` and it will not have that. The runner is normally instantiated with `k6runner.New` which ensures a logger is set. We have this `WithLogger` contraption that switches the logger that is used. This is why the code is peppered with `if r.logger == nil` (because `WithLogger` allows the logger to be set to `nil`).

internal/k6runner/local.go
```diff
@@ -301,6 +315,18 @@ func (r Local) buildK6Args(script Script, metricsFn, logsFn, scriptFn, configFil
 	return args, nil
 }
 
+func buildK6RefID(ci CheckInfo) (string, error) {
```
mem, line 332: it's not obvious to me that this produces a compact JSON object. Can you please add a test to pin down the output?

```go
func TestBuildK6RefID(t *testing.T) {
	checkInfo := CheckInfo{
		Job:      "a_job",
		Instance: "an_instance",
		Probe:    "a_probe",
	}

	id, err := buildK6RefID(checkInfo)
	require.NoError(t, err)
	require.Equal(t, `{"job":"a_job","instance":"an_instance","probe":"a_probe"}`, id)
}
```

mem: Also note that job and instance have very few restrictions placed on them (non-empty, max length, IIRC). How is this value going to be communicated to the rest of the system? Is it important that it is a JSON object? Can it be base64-encoded?

mem: Ah, sorry, this is tested indirectly by testing buildk6args...

internal/k6runner/local.go
```diff
@@ -18,6 +18,13 @@ import (
 	"github.com/spf13/afero"
 )
 
+// k6CloudPushRefIDEnvVar is the name of the k6 runtime environment variable used to
+// tag browser tests as originated from Synthetic Monitoring. It contains a reference
+// to the check's job, instance, and probe associated with the execution.
+// k6 will propagate this data to the window.k6 property, which is parsed by FE O11y
+// as a means to correlate between a specific browser session and a check execution.
```
mem, line 25: meaning you can choose to encode and decode the value in a particular way that k6 will handle?

mem (commented):
LGTM

The encoding of the ID is something I would like to see addressed.

internal/k6runner/local.go
```diff
@@ -289,7 +296,14 @@ func (r Local) buildK6Args(script Script, metricsFn, logsFn, scriptFn, configFil
 		r.logger.Debug().Msg("No secret source configuration to add to k6")
 	}
 
-	if script.CheckInfo.Type != synthetic_monitoring.CheckTypeBrowser.String() {
+	// For browser checks, set K6_CLOUD_PUSH_REF_ID env variable for FE O11y correlation.
+	if script.CheckInfo.Type == synthetic_monitoring.CheckTypeBrowser.String() {
+		if k6RefID, err := buildK6RefID(script.CheckInfo); err != nil {
+			r.logger.Warn().Err(err).Msg("error building k6RefID")
```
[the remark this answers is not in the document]
mem: also note this can be fixed by moving the check to the top of the function

```go
logger := /* init with noop logger */
if r.logger != nil {
    logger = r.logger
}
```

I *think* the reason why the guards are in place is because there's a test somewhere that constructs a runner without a logger.

internal/k6runner/k6runner.go
```diff
@@ -60,19 +60,28 @@ type Settings struct {
...
-// CheckInfoFromSM returns a CheckInfo from the information of the given SM check.
-func CheckInfoFromSM(smc smmodel.Check) CheckInfo {
+// CheckInfoFromSM returns a CheckInfo from the information of the given SM check and probe name.
+func CheckInfoFromSM(smc smmodel.Check, probeName string) CheckInfo {
 	ci := CheckInfo{
 		Metadata: map[string]any{},
 	}
 
 	ci.Type = smc.Type().String()
+	ci.Job = smc.Job
```
mem, line 82: note that for adhoc checks, which use this same code path, `job` is empty.

How should that case be handled?

internal/k6runner/local.go
mem: Also note the note above about ad-hoc checks: this code path assumes job is set. What happens if it isn't?

internal/k6runner/local.go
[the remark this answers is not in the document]
ka3de: The way the logger is handled for the runner seems a little bit messy, but I won't get into that refactor here. For this PR I've followed your recommendation in [da1c486](https://github.com/grafana/synthetic-monitoring-agent/pull/1783/commits/da1c48694772ab0cec1eebae8e8408af871e3a89).
I'm avoiding copying the logger explicitly.

internal/k6runner/k6runner.go
ka3de: Metadata was only included in order to have an unspecified set of fields that could be propagated up to the log vars. However I would argue we should not handle it, as it's an "opaque" structure, for data the system explicitly needs, such as in this case `instance` and `job`.

internal/k6runner/local.go
ka3de: Answered [above](https://github.com/grafana/synthetic-monitoring-agent/pull/1783#discussion_r3272754773).

internal/k6runner/local.go
ka3de: > Also note that job and instance have very few restrictions placed on them (non-empty, max length, IIRC). How is this value going to be communicated to the rest of the system? Is it important that it is a JSON object? Can it be base64-encoded?

We can decide the format, k6 does not care about this, it will just pass it along into the page DOM. Afterwards this will be parsed by other parts of DEM in order to establish the correlation, but "we" handle that part too.
We have a particular issue, [here](https://github.com/grafana/synthetic-monitoring/issues/505), to decide this that is set as dependency for this PR. Notice that this PR was being kept as _draft_ because some things required a second iteration.

internal/k6runner/local.go
ka3de: As I mentioned above, k6 does not care about this, it will just pass it verbatim.

internal/k6runner/k6runner.go
ka3de: That's another reason why this PR was being kept as draft. This is a discussion we are still having. One option was to propagate the ad-hoc UUID and include that in the reference. That will allow us to differentiate between ad-hoc vs non-ad-hoc and uniquely identify the ad-hoc request. I don't particularly like that option so I'm open for alternatives.

So far I would consider the support for ad-hoc checks as secondary. Therefore my proposal as first iteration is to only include the k6 reference when both, `job` and `instance` are properly set, which should identify non ad-hoc checks.

ka3de: I have included that in [af703ae](https://github.com/grafana/synthetic-monitoring-agent/pull/1783/commits/af703ae7f131233a9c54eb48473d0c731780bf69).

ka3de:
Closed in favor of #1853.

PR #1798, opened 2026-04-30
feat: Reject k6 scripts containing pragma directives

ka3de:
Pragma directives are not supported by Synthetic Monitoring, and frobidden by the Grafana Cloud app. This commits includes validation for the backend.

Updates grafana/synthetic-monitoring/issues/566

---
#### Notes for reviewer
The implementation uses the same regex and comment-stripping logic from [github.com/grafana/k6deps](https://github.com/grafana/k6deps/blob/d300092847f3f53239a89dfe32a0271df46c4802/dependencies.go#L13-L37).

The regex definitions are copied here rather than imported for two reasons:
- The k6deps UnmarshalJS API parses both _imports_ and _pragmas_ into the same `Dependencies` map without distinguishing them. So we can't reliably detect all pragma forms.
- The k6deps module transitively depends on esbuild (used for JS bundling in its Analyze() path). Since we only need a regex match, importing the module would add ~40MB to the module cache for no compiled benefit. The practical cost is small (cached download, no binary size impact), but unnecessary.

The `use k6 ...` pragma syntax is a stable part of the k6 spec, so drift risk from the copied regex is low.

An alternative would be adding an exported `ContainsPragma()` function to k6deps and importing it.

---

> [!NOTE]
> **Medium Risk**
> Tightens validation on user-supplied k6 scripts and can cause previously accepted scripted/browser checks to be rejected if they include pragmas; risk is limited to these validation paths and is covered by new tests.
> 
> **Overview**
> **Scripted and browser checks now reject k6 pragma directives.** `ScriptedSettings.Validate` and `BrowserSettings.Validate` fail with a new `ErrK6ScriptContainsPragma` when the script contains `"use k6 ..."` pragmas.
> 
> This introduces regex-based pragma detection (with normalization and comment stripping) to avoid false positives in `//` and `/* */` comments, and adds unit tests covering valid scripts, multiple pragma forms, and comment-handling behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4577b0ca78bb85ca87db7c0058fabdb013663a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/pb/synthetic_monitoring/checks_extra.go
+++ b/pkg/pb/synthetic_monitoring/checks_extra.go
@@ -20,6 +20,7 @@ package synthetic_monitoring
 //go:generate ../../../scripts/enumer -type=MultiHttpEntryAssertionType,MultiHttpEntryAssertionSubjectVariant,MultiHttpEntryAssertionConditionVariant,MultiHttpEntryVariableType -trimprefix=MultiHttpEntryAssertionType_,MultiHttpEntryAssertionSubjectVariant_,MultiHttpEntryAssertionConditionVariant_,MultiHttpEntryVariableType_ -transform=upper -output=multihttp_string.go
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -89,7 +90,8 @@ var (
 
 	ErrInvalidTracerouteHostname = errors.New("invalid traceroute hostname")
 
-	ErrInvalidK6Script = errors.New("invalid K6 script")
+	ErrInvalidK6Script        = errors.New("invalid K6 script")
+	ErrK6ScriptContainsPragma = errors.New("k6 script contains unsupported pragma")
 
 	ErrInvalidMultiHttpTargets = errors.New("invalid multi-http targets")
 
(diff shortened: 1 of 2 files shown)


Pokom (commented):
Is the impact of this limited to creating new checks? Or would this impact the ability to run checks that may have already been created? If I understand the issue correctly, the app has validation on the frontend that would prevent checks from being created, but checks managed via terraform could potentially bypass this.

ka3de:
> Is the impact of this limited to creating new checks? Or would this impact the ability to run checks that may have already been created?

The intention was to only apply during check create/update validation. Reviewing it now I see that the actual check's `Validate()` method, which is extended here for scripted/browser, is also called when a check is handled by an agent. That means that, if there exists a check which currently has pragmas defined on it, it will stop running. That's probably not what we want, plus there is no need to do that also beyond initial validation. So what we can do is move this to the API during HTTP request validation.

> If I understand the issue correctly, the app has validation on the frontend that would prevent checks from being created, but checks managed via terraform could potentially bypass this.

That's correct. We ACK that some check update might fail, but validation has to be in place in order to be coherent with FE and inform users that pragmas are ignored.

vmgrlab:
> > Is the impact of this limited to creating new checks? Or would this impact the ability to run checks that may have already been created?
> 
> The intention was to only apply during check create/update validation. Reviewing it now I see that the actual check's `Validate()` method, which is extended here for scripted/browser, is also called when a check is handled by an agent. That means that, if there exists a check which currently has pragmas defined on it, it will stop running. That's probably not what we want, plus there is no need to do that also beyond initial validation. So what we can do is move this to the API during HTTP request validation.
> 

I agree, breaking checks silently is not a good idea. Will you close this PR ?

Pokom:
To be clear, I do not think it's a _problem_ to introduce a breaking change, it just needs to be intentional and communicated out. @ka3de and I were talking over DM, this is likely impacting a very minor set of users. What we would like to do is gather the check scripts and see if we can figure out customers that would have checks breaking and communicate it prior to releasing the change.

vmgrlab:
> To be clear, I do not think it's a _problem_ to introduce a breaking change, it just needs to be intentional and communicated out. @ka3de and I were talking over DM, this is likely impacting a very minor set of users. What we would like to do is gather the check scripts and see if we can figure out customers that would have checks breaking and communicate it prior to releasing the change.

Just curious — how did scripts with pragma end up in the DB? Was the frontend check enabled later? Since pragma is ignored, those scripts were likely never validated by customers, or may not be working as intended.

PR #1810, opened 2026-05-08
[DRAFT] feat: Control removal of label_ prefix

The-9880:
Introduces a LabelMode enum with three values:
  LABEL_MODE_PREFIXED   = 0  (default, legacy)
  LABEL_MODE_DUAL_WRITE = 1  (writes both label_foo and foo)
  LABEL_MODE_UNPREFIXED = 2  (writes foo only; irreversible)

The enum is added as field 10 on the Tenant proto message, which the API
will populate from the tenants.label_mode DB column and the agent reads
via GetTenant on each scrape cycle.

Also adds the private systemLabels map and the exported IsSystemLabel /
SystemLabelNames helpers. These enumerate the 32 label names reserved by
the agent (probe, instance, job, region, etc.) and are imported by the
API to enforce label constraints at check/probe creation time.


The-9880:
Superseded by https://github.com/grafana/synthetic-monitoring-agent/pull/1879

PR #1841, opened 2026-06-01
fix: Push to GAR, not dockerhub

mem:
We can no longer use dockerhub directly. Push to GAR, so that we can mirror from there to dockerhub.


VikaCep (approved):
LGTM

PR #1853, opened 2026-06-08
feat: Set k6 ref. id as check execution ID for browser checks run by local runner

ka3de:
Sets the `K6_CLOUD_PUSH_REF_ID` environment variable within the k6 test script runtime for browser checks, referencing the check's _execution ID_ . This variable is automatically parsed by k6 runtime and set as `window.k6 property` within the web page DOM for every page visited by the browser check execution.

Introduces a new check _execution ID_, an UUID that identifies each check execution, included as [structured metadata](https://grafana.com/docs/loki/latest/get-started/labels/structured-metadata/). This execution ID allows grouping all logs for each check execution.

The purpose of this is to establish a correlation between a FE O11y browser session and Synthetic Monitoring check execution. By including the execution ID, FE O11y can query

This PR also propagates the check's _execution ID_ to the remote HTTP runner.

Depends on https://github.com/grafana/synthetic-monitoring/issues/505
Updates https://github.com/grafana/synthetic-monitoring/issues/506
Alternative to https://github.com/grafana/synthetic-monitoring-agent/pull/1783

----
#### Pros and Cons
Compared to #1783, using a _check execution ID_ allows for better identification of the check execution logs associated with the FE O11y session without relying on timing. This implies also an additional side benefit to our current _Timepoint Explorer_ implementation, which can now do the same without relying on specific "log lines as sentinels".

The concern with this approach is whether some tenant's volume might be too much in order to rely only on structured metadata.
Fe O11y will only have this reference in order to query logs, therefore not using any labels to reduce the log stream results. Synthetic Monitoring volumes are small, but for tenants with tenths of thousands of checks total log size can reach significant volumes. Theoretically, structured metadata should allow the Loki query to select particular streams before retrieval, therefore reducing volume. We should understand and guarantee that this solution will be appropriate, even for big tenants.
How will Loki behave on a query for _execution ID_ for big tenants given a time window of 1/2/..4 weeks? Will Loki give up on the query due to total volume?

The alternative here could imply having a merge of both approaches, so the k6 ref. id includes a reference to `job` and `instance` along with the _execution ID_, keeping this last one as structured metadata. In this case FE o11y could use `job` and `instance` labels as part of the LogQL query.

---

> [!NOTE]
> **Medium Risk**
> Broad interface change across probers and k6 runners touches all check execution paths; browser checks depend on correct k6 env propagation and remote runners must accept executionId.
> 
> **Overview**
> Each check run now gets a new **UUID execution ID** that flows through scheduled scrapes, ad-hoc runs, and all k6-based probes.
> 
> Published Loki log entries attach that ID as **`execution_id` structured metadata** (not a stream label), so logs for one run can be grouped and queried without relying on timing alone.
> 
> The **`Prober.Probe`** and **`k6runner.Runner.Run`** APIs take an **`executionID`** argument end-to-end. Remote k6 runs send it as **`executionId`** on the HTTP run payload; the local runner sets **`K6_CLOUD_PUSH_REF_ID=sm:<executionID>`** for **browser** checks (and skips `--vus`/`--iterations` for those), so FE O11y can tie a browser session to a specific SM execution via `window.k6`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a38ee07d4f8ac174b9912e854cbe592e320405a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -570,6 +570,7 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		s.summaries, s.histograms,
 		sl,
 		s.check.BasicMetricsOnly,
+		executionID,
 	)
 	if err != nil {
 		return nil, 0, err
@@ -689,10 +690,11 @@ func getProbeMetrics(
 	histograms map[uint64]prometheus.Histogram,
 	logger kitlog.Logger,
 	basicMetricsOnly bool,
+	executionID string,
 ) (bool, []*dto.MetricFamily, error) {
 	registry := prometheus.NewRegistry()
 
-	success := runProber(ctx, prober, target, timeout, registry, checkInfoLabels, logger)
+	success := runProber(ctx, prober, target, timeout, registry, checkInfoLabels, logger, executionID)
 
 	mfs, err := registry.Gather()
 	if err != nil {
@@ -723,6 +725,7 @@ func runProber(
 	registry *prometheus.Registry,
 	checkInfoLabels map[string]string,
 	logger kitlog.Logger,
+	executionID string,
 ) bool {
 	start := time.Now()
 
@@ -731,7 +734,7 @@ func runProber(
 	checkCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	success, duration := prober.Probe(checkCtx, target, registry, logger)
+	success, duration := prober.Probe(checkCtx, target, registry, logger, executionID)
 
 	wallDuration := time.Since(start).Seconds()
 
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -12,11 +12,10 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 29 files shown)


internal/scraper/scraper.go
```diff
@@ -508,6 +508,8 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		userLabels = s.buildUserLabels()
 		// These labels are applied to the sm_check_info metric.
 		checkInfoLabels = s.buildCheckInfoLabels(userLabels)
+		// This is the execution ID for the check run.
+		executionID = uuid.New().String()
```
ka3de, line 512: By now I'm keeping the _execution ID_ generation in this PR along with the inclusion as k6 ref. ID for browser checks so we can discuss about implications of its format etc. Once we have an agreement we can separate _execution ID_ generation from k6 ref. id setting in different PRs.

mem (approved):

PR #1878, opened 2026-06-25
feat(proto): add LabelMode enum to Tenant proto

The-9880:
## Summary

Follows https://github.com/grafana/synthetic-monitoring/issues/541

Introduces the proto and shared-package definitions required to implement tenant-controlled label prefix removal. The agent's scraper implementation follows in a separate PR.

## What this adds

**Proto (`checks.proto`):**
- `LabelMode` enum with three values:
  - `LABEL_MODE_PREFIXED = 0` — default; all user labels get the `label_` prefix
  - `LABEL_MODE_DUAL_WRITE = 1` — writes both `label_foo` and `foo`; entry from PREFIXED is irreversible
  - `LABEL_MODE_UNPREFIXED = 2` — writes `foo` only; reversible back to DUAL_WRITE
- `LabelMode label_mode = 10` field on the `Tenant` message

**`checks_extra.go`:**
- Private `systemLabels` map — the 32 label names reserved by the agent (e.g. `probe`, `instance`, `job`, `region`)
- `IsSystemLabel(name string) bool` — exported for use by the API's reserved-label enforcement
- `SystemLabelNames() []string` — returns the sorted list for the settings UI

## State machine

```
PREFIXED ──────────────────► DUAL_WRITE ◄──────────────► UNPREFIXED
          (irreversible)                  (reversible)
```

PREFIXED→DUAL_WRITE is the one-way door because it is an additive, backward-query-compatible change (existing PromQL expressions continue to work via subset matching). DUAL_WRITE↔UNPREFIXED is reversible — tenants can return to dual-write if they have not finished migrating all their label-based policies.


pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -131,6 +131,51 @@ const (
...
+}
+
+// SystemLabelNames returns a sorted slice of all reserved system label names.
+// The result is a new slice on each call; callers must not modify it.
+func SystemLabelNames() []string {
+	names := make([]string, 0, len(systemLabels))
+	for n := range systemLabels {
+		names = append(names, n)
+	}
+	return names
```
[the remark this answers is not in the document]
The-9880: Ah, good catch on the doc string. I don't believe it matters if the list is sorted or not, either.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -131,6 +131,48 @@ const (
...
+}
+
+// SystemLabelNames returns a slice of all reserved system label names.
+// Order is not guaranteed. The result is a new slice on each call; callers must not modify it.
+func SystemLabelNames() []string {
+	names := make([]string, 0, len(systemLabels))
+	for n := range systemLabels {
+		names = append(names, n)
+	}
+	return names
```
mem: not sure what the use case is here, but doing this every time this function is called sounds a bit wasteful.

I wasn't going to comment on the formatting of the map (it's hard to read) but looking at this I think you can clean this up.

I'm going to push a commit on top of this.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -131,6 +132,83 @@ const (
...
+
+	for _, label := range systemLabels {
+		set[label] = struct{}{}
+	}
+
+	return set
+}()
+
+// SystemLabelCount is the number of reserved system label names.
+func SystemLabelCount() int { return len(systemLabels) }
```
mem, line 195: Go requires constant values to be constant at build time, so it's not possible to do `const foo = func() int {...}()`

mem: If you really really want `const SystemLabelCount = 27`, add a test to control for drift. `var SystemLabelCount` is a non-starter.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -131,6 +132,83 @@ const (
...
+	return found
+}
+
+// SystemLabelNames returns a slice of all reserved system label names.
+//
+// The returned slice is always sorted alphabetically.
+//
+// The result is a new slice on each call; callers are free to use it as they
+// see fit.
+func SystemLabelNames() []string { return slices.Clone(systemLabels) }
```
mem, line 210: not sure what the use case is here, but defining the list as a slice allows you to simplify this function. If you really want the map, use `maps.Keys(...)`.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -131,6 +132,83 @@ const (
...
+	"scenario",
+	"serialnumber",
+	"sm_region",
+	"status",
+	"subject",
+	"subjectalternative",
+	"tls_version",
+	"url",
+	"version",
+}
```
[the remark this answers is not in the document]
mem: That is not incorrect.

mem:
> **TruffleHog Scan Results**
> 
> **Summary:** Found 2 potential secrets (0 verified, 2 unverified)
> 
> * **Possible secret** (URI) at `pkg/pb/synthetic_monitoring/checks_extra_test.go:1080` → `http***.org`
> * **Possible secret** (URI) at `pkg/pb/synthetic_monitoring/checks_extra_test.go:1080` → `http***.org`

you are drunk, go home.

mem (approved):

The-9880:
@mem I'm unable to review your commits in this PR - I approve, but I must either re-push them and have you re-approve, or you can open a separate PR with these commits and I can approve them formally there

The-9880:
or, @ka3de if you would be so kind as to review _our_ PR, that would be excellent

ka3de (approved):

PR #1879, opened 2026-06-25
feat: Implement label mode behaviour

The-9880:
## Summary

Relates to https://github.com/grafana/synthetic-monitoring/issues/541

Implements the agent-side scraper behaviour for tenant-controlled label prefix removal. Depends on #1878 (`feat(proto)`), which must merge first.

## What this adds

**`TenantLabelMode` provider — separate from `LabelsLimiter`**

Follows the same pattern as `TenantCals` for cost attribution labels. A new `internal/labelmode` package exposes a `LabelMode` struct backed by a `TenantProvider` to return the label mode of a tenant by ID.

**`buildUserLabels()` — mode-aware label emission**

Accepts an explicit `LabelMode`. Behaviour by mode:

- `PREFIXED` (0, default): `env=prod` → `label_env="prod"` (unchanged)
- `DUAL_WRITE` (1): all prefixed forms first, then all un-prefixed — `[label_foo, label_bar, foo, bar]`. 
  - If the label limit for Loki is exceeded, remaining labels spill into the `StructuredMetadata`.
  - The ordering ensures the `label_`-prefixed labels are retained to avoid breaking dependencies on indexed labels.
- `UNPREFIXED` (2): `env=prod` → `env="prod"` only

Reserved system label names are **not filtered**. The API enforces that tenants clear all reserved-name collisions before entering `DUAL_WRITE`, so any post-migration collision can only be caused by a new label we add to the reserved set. In that case the user-defined value should win — preserving existing tenant behaviour is preferable to silently overwriting/breaking it.

**`userLabelsForExecution()` — user labels on check execution metrics**

In `DUAL_WRITE` and `UNPREFIXED` modes, un-prefixed user labels are appended to the shared label set applied to all check execution metrics (`probe_success`, `probe_duration_seconds`, `probe_http_*`, etc.). Only the un-prefixed form is added to execution metrics in `DUAL_WRITE` — the dual-write of both prefixed and un-prefixed forms applies only to `sm_check_info` and Loki streams.

Enables label-based cost attribution and direct PromQL filtering without a join to `sm_check_info`, e.g. `rate(probe_success{env="prod"}[5m])`.

**Log label overflow to `StructuredMetadata`**

When the combined system and user label count would exceed Loki's 15-stream-label cap, excess user labels spill into `logproto.Entry.StructuredMetadata` rather than being silently truncated. Structured metadata is queryable in LogQL but not indexed as stream labels.

**Unit tests**

- `TestBuildUserLabels_*` — all three modes, check-overrides-probe, DUAL_WRITE ordering (`[label_foo, label_bar, foo, bar]`), reserved name passthrough (user value preserved)
- `TestUserLabelsForExecution_*` — PREFIXED returns nil, DUAL_WRITE/UNPREFIXED return un-prefixed only, reserved names pass through, check overrides probe

## State machine note

The state machine (`PREFIXED → DUAL_WRITE` irreversible, `DUAL_WRITE ↔ UNPREFIXED` reversible) is enforced by the API, not the agent. The agent reads `LabelMode` from the tenant cache and applies it — no transition logic lives here.

## Stale series note

When a tenant enters `DUAL_WRITE`, existing execution metric series (without user labels) become stale and new series (with user labels) are created. This is expected — tenants entering `DUAL_WRITE` are in a label migration context and should anticipate schema changes to their series.

### DIFF
--- a/internal/labelmode/labelmode.go
+++ b/internal/labelmode/labelmode.go
@@ -28,6 +28,12 @@ func New(provider TenantProvider) *LabelMode {
 
 // ForTenant returns the label mode for the given tenant, defaulting to
 // LABEL_MODE_PREFIXED if the tenant cannot be fetched.
+//
+// The default value is never actually used - the LabelMode is fetched
+// during the scraper's `collectData` operation, which aborts overall
+// if the tenant is not able to be retrieved as the tenant's limits
+// are necessary to the scraper's operation. The default path is therefore
+// irrelevant for scrapes which report telemetry.
 func (lm *LabelMode) ForTenant(ctx context.Context, tenantID model.GlobalID) (sm.LabelMode, error) {
 	tenant, err := lm.provider.GetTenant(ctx, &sm.TenantInfo{Id: int64(tenantID)})
 	if err != nil {
--- a/internal/labelmode/labelmode.go
+++ b/internal/labelmode/labelmode.go
@@ -27,7 +27,7 @@ func New(provider TenantProvider) *LabelMode {
 }
 
 // ForTenant returns the label mode for the given tenant, defaulting to
-// LABEL_MODE_PREFIXED if the tenant cannot be fetched.
+// LABEL_MODE_UNPREFIXED if the tenant cannot be fetched.
 //
 // The default value is never actually used - the LabelMode is fetched
 // during the scraper's `collectData` operation, which aborts overall
@@ -37,7 +37,7 @@ func New(provider TenantProvider) *LabelMode {
 func (lm *LabelMode) ForTenant(ctx context.Context, tenantID model.GlobalID) (sm.LabelMode, error) {
 	tenant, err := lm.provider.GetTenant(ctx, &sm.TenantInfo{Id: int64(tenantID)})
 	if err != nil {
-		return sm.LabelMode_LABEL_MODE_PREFIXED, fmt.Errorf("fetching tenant label mode: %w", err)
+		return sm.LabelMode_LABEL_MODE_UNPREFIXED, fmt.Errorf("fetching tenant label mode: %w", err)
 	}
 	return tenant.LabelMode, nil
 }
--- a/internal/labelmode/labelmode.go
+++ b/internal/labelmode/labelmode.go
@@ -1,0 +1,37 @@
+package labelmode
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/grafana/synthetic-monitoring-agent/internal/model"
+	sm "github.com/grafana/synthetic-monitoring-agent/pkg/pb/synthetic_monitoring"
+)
+
+// TenantProvider is the subset of the tenant manager interface required to
+// look up a tenant's current label mode.
+type TenantProvider interface {
+	GetTenant(context.Context, *sm.TenantInfo) (*sm.Tenant, error)
+}
+
+// LabelMode provides the tenant's current label migration mode
+// (PREFIXED / DUAL_WRITE / UNPREFIXED) by reading it from the GetTenant
+// response, which is cached with a 15-minute TTL by the tenant manager.
+type LabelMode struct {
+	provider TenantProvider
+}
+
+// New creates a LabelMode provider backed by the given TenantProvider.
+func New(provider TenantProvider) *LabelMode {
+	return &LabelMode{provider: provider}
+}
+
+// ForTenant returns the label mode for the given tenant, defaulting to
+// LABEL_MODE_PREFIXED if the tenant cannot be fetched.
+func (lm *LabelMode) ForTenant(ctx context.Context, tenantID model.GlobalID) (sm.LabelMode, error) {
+	tenant, err := lm.provider.GetTenant(ctx, &sm.TenantInfo{Id: int64(tenantID)})
+	if err != nil {
+		return sm.LabelMode_LABEL_MODE_PREFIXED, fmt.Errorf("fetching tenant label mode: %w", err)
+	}
+	return tenant.LabelMode, nil
+}
--- a/internal/scraper/scraper.go
+++ b/internal/scraper/scraper.go
@@ -69,9 +69,6 @@ type TenantCals interface {
 	CostAttributionLabels(ctx context.Context, tenantID model.GlobalID) ([]string, error)
 }
 
-// TenantLabelMode provides the label migration mode for a tenant.
-// The implementation reads from the GetTenant response, which is cached by the
-// tenant manager, so calls within the cache TTL are fast local reads.
 type TenantLabelMode interface {
 	ForTenant(ctx context.Context, tenantID model.GlobalID) (sm.LabelMode, error)
 }
@@ -89,9 +86,9 @@ type Scraper struct {
 	check         model.Check
 	probe         sm.Probe
 	prober        prober.Prober
-	labelsLimiter  LabelsLimiter
-	labellingMode  TenantLabelMode
-	stop           chan struct{}
+	labelsLimiter LabelsLimiter
+	labellingMode TenantLabelMode
+	stop          chan struct{}
 	metrics       Metrics
 	summaries     map[uint64]prometheus.Summary
 	histograms    map[uint64]prometheus.Histogram
@@ -516,10 +513,16 @@ func tickWithOffset(
 }
 
 func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time.Duration, error) {
+	target := s.target
+
+	labelMode, err := s.labellingMode.ForTenant(ctx, s.check.GlobalTenantID())
+	if err != nil {
+		return nil, 0, fmt.Errorf("retrieving tenant label mode: %w", err)
+	}
+
 	var (
-		target = s.target
 		// These are the labels defined by the user.
-		userLabels = s.buildUserLabels()
+		userLabels = buildUserLabels(s.probe.Labels, s.check.Labels, labelMode)
 		// These labels are applied to the sm_check_info metric.
 		checkInfoLabels = s.buildCheckInfoLabels(userLabels)
 		// This is the execution ID for the check run.
@@ -608,6 +611,14 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		// {name: "source", value: CheckInfoSource}, // identify metrics that belong to synthetic-monitoring-agent
 	}
 
+	// In DUAL_WRITE and UNPREFIXED modes, append un-prefixed user labels to every
+	// execution metric (probe_success, probe_duration_seconds, probe_http_*, etc.).
+	// This enables label-based filtering and cost attribution directly on high-frequency
+	// series without requiring a join to sm_check_info.
+	// Only the un-prefixed form is added — the dual-write of prefixed+un-prefixed
+	// continues to apply to sm_check_info and Loki streams only.
+	metricLabels = append(metricLabels, userLabelsForExecution(s.probe.Labels, s.check.Labels, labelMode)...)
+
 	ts := s.extractTimeseries(t, mfs, metricLabels)
 
 	successValue := "1"
@@ -616,13 +627,27 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		successValue = "0"
 	}
 
-	if len(logLabels) >= maxLogLabels {
-		logLabels = logLabels[:maxLogLabels-1]
+	// Split logLabels into stream labels and structured metadata overflow.
+	// maxLogLabels - 1 reserves one slot for probe_success which is always appended.
+	var streamLogLabels []labelPair
+	var overflowMetadata []logproto.LabelAdapter
+	maxUserLogLabels := maxLogLabels - 1
+	if len(logLabels) > maxUserLogLabels {
+		streamLogLabels = logLabels[:maxUserLogLabels]
+		for _, lp := range logLabels[maxUserLogLabels:] {
+			overflowMetadata = append(overflowMetadata, logproto.LabelAdapter{Name: lp.name, Value: lp.value})
+		}
+	} else {
+		streamLogLabels = logLabels
 	}
-	logLabels = append(logLabels, labelPair{name: ProbeSuccessMetricName, value: successValue}) // identify log lines that are failures
+	streamLogLabels = append(streamLogLabels, labelPair{name: ProbeSuccessMetricName, value: successValue}) // identify log lines that are failures
+
+	// full set of structured metadata consists of any overflow labels
+	// and the execution_id
+	structuredMetadata := append(overflowMetadata, logproto.LabelAdapter{Name: "execution_id", Value: executionID})
 
 	// streams need to have all the labels applied to them because loki does not support joins
-	streams := s.extractLogs(t, logs.Bytes(), logLabels, executionID)
+	streams := s.extractLogs(t, logs.Bytes(), streamLogLabels, structuredMetadata)
 
 	return &probeData{ts: ts, streams: streams, tenantId: s.check.GlobalTenantID()}, duration, err
 }
@@ -835,7 +860,10 @@ func getDerivedMetrics(mfs []*dto.MetricFamily, summaries map[uint64]prometheus.
 	return nil
 }
 
-func (s Scraper) extractLogs(t time.Time, logs []byte, sharedLabels []labelPair, executionID string) Streams {
+// extractLogs parses logfmt-encoded log bytes and returns Loki streams using sharedLabels
+// as stream labels. Any labels in structuredMetadata are attached to each log entry as
+// Loki structured metadata (queryable but not indexed as stream labels).
+func (s Scraper) extractLogs(t time.Time, logs []byte, sharedLabels []labelPair, structuredMetadata []logproto.LabelAdapter) Streams {
 	var line strings.Builder
 
 	dec := logfmt.NewDecoder(bytes.NewReader(logs))
@@ -896,11 +924,9 @@ RECORD:
 			s.logger.Warn().Err(err).Msg("encoding logs")
 		}
 		entries = append(entries, logproto.Entry{
-			Timestamp: t,
-			Line:      line.String(),
-			StructuredMetadata: logproto.LabelsAdapter{
-				{Name: "execution_id", Value: executionID},
-			},
+			Timestamp:          t,
+			Line:               line.String(),
+			StructuredMetadata: structuredMetadata,
 		})
 	}
 
@@ -955,32 +981,89 @@ func (s Scraper) buildCheckInfoLabels(userLabels []labelPair) map[string]string 
 	return labels
 }
 
-func (s Scraper) buildUserLabels() []labelPair {
-	labels := []labelPair{}
-	idx := make(map[string]int)
+// buildUserLabels builds the user-defined label pairs from probe and check labels
+// according to the tenant's LabelMode.
+//
+// Reserved system label names are not filtered here. The API enforces that tenants
+// in DUAL_WRITE or UNPREFIXED mode cannot have user labels whose names collide with
+// the reserved set at the time of migration. Any collision that occurs post-migration
+// can only be caused by adding a new label to the reserved set, in which case the
+// user-defined value should win — preserving existing tenant behaviour is preferable
+// to silently breaking it.
+//
+// In DUAL_WRITE mode, all prefixed forms are emitted first, followed by all
+// un-prefixed forms: [label_foo, label_bar, foo, bar]. This ordering ensures that
+// in the event of log stream label overflow, the prefixed forms — which existing
+// policies depend on — are preserved as stream labels, while the un-prefixed forms
+// spill into StructuredMetadata.
+func buildUserLabels(probeLabels, checkLabels []sm.Label, mode sm.LabelMode) []labelPair {
+	// Collect unique labels in order, with check values overriding probe values.
+	collected := []labelPair{}
+	seen := make(map[string]int) // name → index in collected
 
-	// add probe labels
-	for _, l := range s.probe.Labels {
-		idx[l.Name] = len(labels)
-
-		labels = append(labels,
-			labelPair{name: "label_" + l.Name, value: l.Value})
-	}
-
-	// add check labels
-	for _, l := range s.check.Labels {
-		if where, found := idx[l.Name]; found {
-			// already there, update value
-			labels[where].value = l.Value
-			continue
+	for _, src := range [][]sm.Label{probeLabels, checkLabels} {
+		for _, l := range src {
+			if i, found := seen[l.Name]; found {
+				collected[i].value = l.Value
+				continue
+			}
+			seen[l.Name] = len(collected)
+			collected = append(collected, labelPair{name: l.Name, value: l.Value})
 		}
-
-		idx[l.Name] = len(labels)
-
-		labels = append(labels,
-			labelPair{name: "label_" + l.Name, value: l.Value})
 	}
 
+	switch mode {
+	case sm.LabelMode_LABEL_MODE_DUAL_WRITE:
+		// All prefixed forms first, then all un-prefixed forms.
+		labels := make([]labelPair, 0, 2*len(collected))
+		for _, e := range collected {
+			labels = append(labels, labelPair{name: "label_" + e.name, value: e.value})
+		}
+		for _, e := range collected {
+			labels = append(labels, labelPair{name: e.name, value: e.value})
+		}
+		return labels
+	case sm.LabelMode_LABEL_MODE_UNPREFIXED:
+		labels := make([]labelPair, 0, len(collected))
+		for _, e := range collected {
+			labels = append(labels, labelPair{name: e.name, value: e.value})
+		}
+		return labels
+	default: // LABEL_MODE_PREFIXED
+		labels := make([]labelPair, 0, len(collected))
+		for _, e := range collected {
+			labels = append(labels, labelPair{name: "label_" + e.name, value: e.value})
+		}
+		return labels
+	}
+}
+
+// userLabelsForExecution returns the un-prefixed user-defined labels to be added to
+// check execution metrics (probe_success, probe_duration_seconds, probe_http_*, etc.)
+// in DUAL_WRITE and UNPREFIXED modes.
+//
+// In PREFIXED mode this returns nil — execution metrics carry no user labels.
+//
+// In DUAL_WRITE and UNPREFIXED modes, only the un-prefixed form is returned — the
+// dual-write of prefixed+un-prefixed applies only to sm_check_info and Loki streams.
+func userLabelsForExecution(probeLabels, checkLabels []sm.Label, mode sm.LabelMode) []labelPair {
+	if mode == sm.LabelMode_LABEL_MODE_PREFIXED {
+		return nil
+	}
+
+	var labels []labelPair
+	seen := make(map[string]int) // base name → index in labels; check overrides probe
+
+	for _, src := range [][]sm.Label{probeLabels, checkLabels} {
+		for _, l := range src {
+			if idx, found := seen[l.Name]; found {
+				labels[idx].value = l.Value
+				continue
+			}
+			seen[l.Name] = len(labels)
+			labels = append(labels, labelPair{name: l.Name, value: l.Value})
+		}
+	}
 	return labels
 }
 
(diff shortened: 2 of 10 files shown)


internal/scraper/scraper.go
```diff
@@ -595,6 +615,10 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		// {name: "source", value: CheckInfoSource}, // identify metrics that belong to synthetic-monitoring-agent
 	}
 
+	// In DUAL_WRITE and UNPREFIXED modes, append un-prefixed user labels to every
+	// execution metric (probe_success, probe_duration_seconds, probe_http_*, etc.).
+	metricLabels = append(metricLabels, userLabelsForExecution(s.probe.Labels, s.check.Labels, labelMode)...)
```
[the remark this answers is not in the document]
The-9880: The API validation for transitioning into these states precludes this situation, as previously noted in other areas (PR description, code).

internal/scraper/scraper.go
```diff
@@ -595,6 +622,10 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
...
+
```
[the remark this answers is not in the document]
The-9880: This is fine - we've already validated via the catalogue that the label limit we set on the API in addition to the check metric labels will not exceed this limit.

mem:
As a general comment, `collectData` tests using `testLabellingMode{}` are testing the `PREFIXED` path (the zero value), and the non-PREFIXED path is not considered. The doubled limit, the execution-metric label append, and the DUAL_WRITE-ordering/overflow interaction are all untested end-to-end. Functions are covered, but the end-to-end wiring isn't.

I know this is a tall ask, but adding tests for one side of the code and not the other isn't going to work.

internal/scraper/scraper.go
```diff
@@ -522,7 +541,15 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
...
+	// Cap at 40 (the Mimir hard limit) to avoid false-firing while still
+	// guarding against genuinely invalid sizes.
+	checkInfoLimit := min(maxMetricLabels*2, 40)
+	if labelMode != sm.LabelMode_LABEL_MODE_DUAL_WRITE {
+		checkInfoLimit = maxMetricLabels
+	}
+	if len(checkInfoLabels) > checkInfoLimit {
 		// This should never happen.
 		return nil, 0, fmt.Errorf("invalid configuration, too many labels: %d", len(checkInfoLabels))
 	}
```
mem, line 571: If this guard fires, it causes collectData to return an error, which discards the entire probe results.

I know you have gone thru the existing tenants to make sure that none have maxMetricLabels greater than ~ 20.

That said, I'm not sure this works, or I don't understand why it should work beyond "this case does not exist in reality". There's a base of 5 labels IIRC, making the limit about 18 (5 + 2×18 = 41).

Is there something else we can do instead of hard failing in that case?

mem: Also, unless I'm missing something, the current Mimir limit is 60, with the caveat that there are three environments where it's something else. Grep for `max-label-names-per-series` in deployment_tools. 40 is the current mimimum minimorum.

internal/scraper/scraper.go
```diff
@@ -522,7 +541,15 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		return nil, 0, fmt.Errorf("retrieving tenant log labels limit: %w", err)
 	}
 
-	if len(checkInfoLabels) > maxMetricLabels {
+	// In DUAL_WRITE mode, buildUserLabels emits both prefixed and un-prefixed
+	// forms, so checkInfoLabels can contain up to 2× the user label entries.
+	// Cap at 40 (the Mimir hard limit) to avoid false-firing while still
+	// guarding against genuinely invalid sizes.
+	checkInfoLimit := min(maxMetricLabels*2, 40)
```
mem: Please give the magic constant 40 a name.

internal/scraper/scraper_test.go
```diff
@@ -1713,8 +1746,9 @@ func TestScraperCollectData(t *testing.T) {
 					maxMetricLabels: tc.maxMetricLabels,
 					maxLogLabels:    tc.maxLogLabels,
 				},
-				summaries:  make(map[uint64]prometheus.Summary),
-				histograms: make(map[uint64]prometheus.Histogram),
+				labellingMode: testLabellingMode{},
```
mem: This exists in multiple places, so I'm not going to flag all of them.

`testLabellingMode{}` is constructed with its zero value everywhere, so its `mode` field is `sm.LabelMode(0)` = `LABEL_MODE_PREFIXED`.

Consequently `TestScraperCollectData` and every other `collectData`-level test runs only in `PREFIXED` mode. The new mode-dependent logic added to `collectData` is therefore untested end-to-end.

The following paths are not tested:

- the doubled `checkInfoLimit` branch (`scraper.go:548-551`)
- the execution-metric label append via `userLabelsForExecution` (`scraper.go:627`), i.e. that `probe_success`/`probe_duration_seconds` actually receive non-prefixed user labels in `DUAL_WRITE`/`UNPREFIXED`;
- the interaction of `DUAL_WRITE` ordering (`[label_foo, …, foo, …]`) with to overflow-to-structured-metadata split, i.e. that prefixed forms are the ones retained as stream labels when overflow occurs.

The pure functions `buildUserLabels`/`userLabelsForExecution` are well covered in `labels_test.go`, but the wiring in `collectData` is not.

Add table cases that set mode to `DUAL_WRITE` and `UNPREFIXED`, asserting:

- execution timeseries carry the non-prefixed user labels
- with a large enough number of user labels that overflow `maxLogLabels`, the prefixed forms remain as stream labels while the non-prefixed forms land in structured metadata
- a `DUAL_WRITE` case near the label limit to lock down the intended behaviour (hard fail vs something else)

internal/scraper/scraper.go
```diff
@@ -942,32 +988,89 @@ func (s Scraper) buildCheckInfoLabels(userLabels []labelPair) map[string]string
...
+	for _, src := range [][]sm.Label{probeLabels, checkLabels} {
+		for _, l := range src {
+			if idx, found := seen[l.Name]; found {
+				labels[idx].value = l.Value
+				continue
+			}
+			seen[l.Name] = len(labels)
+			labels = append(labels, labelPair{name: l.Name, value: l.Value})
+		}
+	}
```
mem: sorry, I don't like to squint to find differences.

Isn't this the same code as lines 1011-1019 above?

internal/scraper/scraper.go
```diff
@@ -942,32 +988,89 @@ func (s Scraper) buildCheckInfoLabels(userLabels []labelPair) map[string]string
...
+// In PREFIXED mode this returns nil — execution metrics carry no user labels.
+//
+// In DUAL_WRITE and UNPREFIXED modes, only the un-prefixed form is returned — the
+// dual-write of prefixed+un-prefixed applies only to sm_check_info and Loki streams.
+func userLabelsForExecution(probeLabels, checkLabels []sm.Label, mode sm.LabelMode) []labelPair {
+	if mode == sm.LabelMode_LABEL_MODE_PREFIXED {
+		return nil
 	}
 
+	var labels []labelPair
```
mem: preallocate this, you have an upper limit for the size of this slice.

internal/scraper/scraper.go
mem: If I'm understanding the code correctly, this is:

```go
func appendUnique(slice1, slice2 []int) []int {
    if slice1 == nil && slice2 == nil {
        return nil
    }

	result := make([]int, 0, len(slice1)+len(slice2))
	seen := make(map[int]bool)

	addUnique := func(slice []int) {
		for _, val := range slice {
			if !seen[val] {
				seen[val] = true
				result = append(result, val)
			}
		}
	}

	addUnique(slice1)
	addUnique(slice2)

	return slices.Clip(result)
}
```

internal/labelmode/labelmode.go
```diff
@@ -0,0 +1,37 @@
...
+func New(provider TenantProvider) *LabelMode {
+	return &LabelMode{provider: provider}
+}
+
+// ForTenant returns the label mode for the given tenant, defaulting to
+// LABEL_MODE_PREFIXED if the tenant cannot be fetched.
+func (lm *LabelMode) ForTenant(ctx context.Context, tenantID model.GlobalID) (sm.LabelMode, error) {
+	tenant, err := lm.provider.GetTenant(ctx, &sm.TenantInfo{Id: int64(tenantID)})
+	if err != nil {
+		return sm.LabelMode_LABEL_MODE_PREFIXED, fmt.Errorf("fetching tenant label mode: %w", err)
```
mem: I'm not sure I like this.

If GetTenant fails to obtain the information for some reason, a migrate tenant that is already in `UNPREFIXED` mode all of a sudden falls back to `PREFIXED` causing some flapping.

GetTenant has a cache, so during the lifetime of the agent this situation should not occur. The problem happens if the value is not in the cache and the request fails (agent is disconnected from API). I don't think building another sticky cache on top of the existing one is the right thing to do.

internal/scraper/scraper.go
```diff
@@ -503,10 +513,19 @@ func tickWithOffset(
...
 
 func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time.Duration, error) {
+	target := s.target
+
+	labelMode, err := s.labellingMode.ForTenant(ctx, s.check.GlobalTenantID())
+	if err != nil {
+		// Fall back to PREFIXED on transient errors so that a temporary tenant
+		// cache miss does not abort the scrape.
+		s.logger.Warn().Err(err).Msg("could not retrieve tenant label mode, falling back to PREFIXED")
+		labelMode = sm.LabelMode_LABEL_MODE_PREFIXED
```
mem: This is reproducing the logic already present in ForTenant.

You should require ForTenant to return `PREFIXED` in case of erorr.

But please see the comment in that function.

Does switching the mode restart checks?

If yes, this should be a property of the scraper, instead of asking about it every time `collectData` runs.

internal/scraper/scraper.go
```diff
@@ -595,6 +622,10 @@ func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time
 		// {name: "source", value: CheckInfoSource}, // identify metrics that belong to synthetic-monitoring-agent
 	}
 
+	// In DUAL_WRITE and UNPREFIXED modes, append un-prefixed user labels to every
+	// execution metric (probe_success, probe_duration_seconds, probe_http_*, etc.).
+	metricLabels = append(metricLabels, userLabelsForExecution(s.probe.Labels, s.check.Labels, labelMode)...)
```
mem: given that userLabelsForExecution is already allocating a slice and trying to be smart about it, I think you should:

1. Give it a better name
2. Use it to merge the three sets of labels

I think there's a subtle thing here, which is I guess the reason why the second point is not feasable: is the set of user-defined labels overriding metric labels?

If probe labels happens to contain e.g. `instance`, who wins?

You have this:

```
append({instance: aaa}, {instance: bbb})
```

I honestly don't remember what happens in this case. If the first one wins, then it's fine. If the last one wins, it's not. If the sample is rejected because it has the same label twice, it also doesn't work.

All of that is fixed by running the same consolidation function using the three sets of labels as input, in order.

internal/scraper/scraper.go
```diff
@@ -942,32 +988,89 @@ func (s Scraper) buildCheckInfoLabels(userLabels []labelPair) map[string]string
...
+// can only be caused by adding a new label to the reserved set, in which case the
+// user-defined value should win — preserving existing tenant behaviour is preferable
+// to silently breaking it.
+//
+// In DUAL_WRITE mode, all prefixed forms are emitted first, followed by all
+// un-prefixed forms: [label_foo, label_bar, foo, bar]. This ordering ensures that
+// in the event of log stream label overflow, the prefixed forms — which existing
+// policies depend on — are preserved as stream labels, while the un-prefixed forms
+// spill into StructuredMetadata.
+func buildUserLabels(probeLabels, checkLabels []sm.Label, mode sm.LabelMode) []labelPair {
```
mem, line 1036: it took me a while to understand how these two functions are different.

userLabelsForExecution is basically buildUserLabels with mode set to `UNPREFIXED`. I honestly think you'd do better by just doing that.

The bit that is blowing my head is why is the `UNPREFIXED` case below doing _anything_. Shouldn't it simply return collected?

You want this basically (please add comments):

```go
var result []labelPair

switch mode {
case DUAL_WRITE:
    result = make([]labelPair, 0, 2*len(probeLabels), 2*len(checkLabels))

case UNPREFIXED, PREFIXED:
    result = make([]labelPair, 0, len(probeLabels), len(checkLabels))
}

seen := make(map[string]bool)

addUnique := func(slice []labelPair) {
	for _, val := range slice {
			if seen[val.name] {
                continue
            }

			seen[val.name] = true

			if mode != PREFIXED {
                result = append(result, val)
            }

            if mode != UNPREFIXED {
                result = append(result, labelPair{name: "label_"+val.name, value: val.Value)
			}
		}
	}
}

addUnique(probeLabels)
addUnique(checkLabels)

return slices.Clip(result)
```

and then:

```
func userLabelsForExecution(...) []labelPair {
    return buildUserLabels(UNPREFIXED, ...)
}

Comment (mem) [reply internal/scraper/scraper.go:1036]: To be clear: ignoring the many allocations, I'm not saying there's something wrong with the implementation. It's just really hard to read.

File: internal/scraper/scraper_test.go [hunk 12]
```diff
@@ -1443,6 +1452,7 @@ func (t testCalTenants) CostAttributionLabels(_ context.Context, tenantID model.
 	return t.costAttributionLabels, nil
 }
 
+//nolint:gocyclo
```
mem, line 1516: you'd do better adding a helper, e.g. `assertStructuredMetadata`

mem (changes requested):
Sorry about the messy comments, it took me a while to figure some things out and I think I got some of those wrong.

mem:
> Sorry about the messy comments, it took me a while to figure some things out and I think I got some of those wrong.

If I did get something wrong, use that as guidance as for what (kind of) comments are missing.

internal/scraper/scraper.go
mem: Related to that point, there's a comment somewhere in this code that states that some labels should be computed once and reapplied as needed.

The labels coming from metrics can change.

The probe and user labels should be constant.

The prefix mode should be constant thru the life of the scraper (the above question).

If those things are true, it should be possible to compute these things once per scraper. It's going to require additional memory, but my hunch is that that should be compensated by the extra allocations not happening. That also gives you the opportunity to build a shared pool of strings (e.g. per tenant) later on.

internal/labelmode/labelmode.go
The-9880: I'm trying to reason about when the `Tenant` would not be in the cache.

The agent never loads the `Tenant` until the first call to `GetTenant()` - this happens at the start of the scrape cycle: getting `LabelMode`, getting `TenantLimits`. It also happens later in the `Publisher`, but it happens _first_ here.

If the limits cannot be resolved then `collectData` aborts: https://github.com/grafana/synthetic-monitoring-agent/blob/main/internal/scraper/scraper.go#L516-L523. In which case, we wouldn't accrue any metrics/logs with the default `PREFIXED` state.

When the first `collectData` scrape _does_ complete, there must have been a `Tenant` value accessible for that scrape.

So I don't think there's a tangible risk of a tenant moving to `DUAL_WRITE` or `UNPREFIXED` and then flapping _back_ due to the encoded default, because a `Tenant` needs to be present for us to execute the scrape and accrue execution telemetry.

I think the only real "risk" is that a tenant may change state and that state may take some time to propagate to the agent - a sync delay is acceptable, the agent is meant to continue running in its current state if the connection with the API is lost transiently.

internal/scraper/scraper.go
The-9880: `sm_check_info` has a base of 8 labels: `job`, `instance`, `probe`, `config_version`, `check_name`, `region`, `frequency`, and `geohash`. Maximally 9, if you want to count `alert_sensitivity`.

No tenant has more than 10 custom labels - this limit comes from the calculation on the API and accounts for out-of-the-box labels consuming the tenant's `maxMetricLabels` limit - see [here](https://github.com/grafana/synthetic-monitoring-api/blob/main/internal/limits/labels_test.go#L16-L27), [here](https://github.com/grafana/synthetic-monitoring-api/blob/main/internal/db/migrations/mysql/20240305101321_increase_mimir_labels_limit.up.sql#L1), and [here](https://github.com/grafana/synthetic-monitoring-api/blob/main/internal/limits/labels.go#L20) (`maxMetricLabels == 22` by default).

So under `DUAL_WRITE`, we end up with (8 base labels + 10*2 custom labels) = 28 labels on `sm_check_info`. 29 if you want to include `alert_sensitivity`.

The `40` magic number comes from [here](https://grafana.com/docs/grafana-cloud/send-data/metrics/metrics-prometheus/ingestion-errors/#max-label-names-per-series) - it might not be the most recent as you note the latest limit is now `60`, but it should work for our current data + validation rules. I can bump this value to `50` and leave a comment to that effect, if you think we should be more cautious with this.

internal/scraper/scraper.go
The-9880: I've responded to that other comment separately.

re: maintaining the static labels, I found the [comment](https://github.com/grafana/synthetic-monitoring-agent/blob/main/internal/scraper/scraper.go#L588-L590) - I would do this in a separate PR to keep this diff focused.

internal/scraper/scraper.go
The-9880: Just a small nit with your code snippet:
```
if seen[val.name] {
    continue
}
```

Is not the exact intent.  It's:
```
if i, found := seen[val.Name]; found {
    collected[i].value = val.Value
	continue
}
```


The `probeLabels` always come first positionally in the labelset, and if there are duplicate check labels then those replace the probe label's value _in-place_.

That is how it is currently on `main`, anyways, so I try to preserve this logic.

internal/scraper/scraper.go
```diff
@@ -942,46 +1009,123 @@ func (s Scraper) buildCheckInfoLabels(userLabels []labelPair) map[string]string
...
+// customMetricLabels returns the un-prefixed user-defined labels to be added to
+// check execution metrics (probe_success, probe_duration_seconds, probe_http_*, etc.)
+// in DUAL_WRITE and UNPREFIXED modes.
+//
+// In PREFIXED mode this returns nil — execution metrics carry no custom labels.
+func customMetricLabels(probeLabels, checkLabels []sm.Label, mode sm.LabelMode) []labelPair {
+	if mode == sm.LabelMode_LABEL_MODE_PREFIXED {
+		return nil
+	}
+	return buildUserLabels(probeLabels, checkLabels, sm.LabelMode_LABEL_MODE_UNPREFIXED)
```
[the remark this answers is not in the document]
The-9880: This is intentional, the scope is fairly defined for what values this can take on.

The-9880:
@mem thanks for bearing with this - please take another look and let's discuss further feedback.

internal/labelmode/labelmode.go
mem: Could you please add that comment to the code. I think you are correct, and that's a substantial amount of information to be deriving every time somebody reads this code.

internal/scraper/scraper.go
```diff
@@ -903,85 +956,233 @@ RECORD:
...
-	idx := make(map[string]int)
+// buildUserLabels builds the user-defined label pairs from probe and check labels
+// according to the tenant's LabelMode.
+//
+// Reserved system label names are not filtered here. The API enforces that tenants
+// in DUAL_WRITE or UNPREFIXED mode cannot have user labels whose names collide with
+// the reserved set at the time of migration. Any collision that occurs post-migration
+// can only be caused by adding a new label to the reserved set, in which case the
+// user-defined value should win — preserving existing tenant behaviour is preferable
+// to silently breaking it.
```
mem: are you saying the API will call `sm.IsSystemLabel` to reject those?

mem: I'm asking because:

- `appendUniqueLabels(customMetricLabels(...), sysMetricLabels)` puts user labels *first*, and `makeTimeseries` de-dupes keeping the *first* occurrence (`scraper.go:1075-1110`), so the user label shadows the system one.
- `buildCheckInfoLabels` does the same via `appendUniqueLabels(userLabels, baseLabels)`.
- `buildUserLabels`' own doc comment (`scraper.go:1024-1035`) deliberately argues "the user-defined value should win."

For single-valued identity labels this is questionable (the `unprefixed system label collision` test even asserts a user `job=user-job` overrides the system `job` on execution metrics, the opposite of the "silently ignored"). But for **per-series intrinsic metric labels** (`le`, `method`, `phase`, `status`, `result`, `name`, `url`, ...) it is destructive.

`checks_extra.go:138-139` says that only covers "tenants in dual-write mode or created after the feature epoch". Grandfathered tenants may carry these labels, and the doc names the agent as the backstop. The agent isn't implementing that backstop, is it?

Put differently, should the agent also call `IsSystemLabel` somewhere?

mem: **Suggestion** (I cannot emphasize that word more):

Make the implementation match the documented contract: in `DUAL_WRITE` / `UNPREFIXED`, drop user labels for which `sm.IsSystemLabel(name)` is true (both
in `customMetricLabels` and in the un-prefixed forms added to `sm_check_info`),
rather than letting them win (which is the case now).

internal/scraper/scraper.go
```diff
@@ -503,12 +518,32 @@ func tickWithOffset(
 }
 
 func (s Scraper) collectData(ctx context.Context, t time.Time) (*probeData, time.Duration, error) {
+	target := s.target
+
+	labelMode, err := s.labellingMode.ForTenant(ctx, s.check.GlobalTenantID())
+	if err != nil {
+		// ForTenant falls back to PREFIXED on transient errors so that a temporary tenant
+		// cache miss does not abort the scrape.
+		s.logger.Warn().Err(err).Msg("could not retrieve tenant label mode, falling back to PREFIXED")
```
mem: help me here: why does falling back to `PREFIXED` make sense? Shouldn't this fall back to `UNPREFIXED`? If we agree the error should never happen, shouldn't we assume the tenant has been migrated?

I'm trying to figure out which one is the least-bad value to use.

mem (commented):
This is much better, thanks!

A handful of questions.

internal/scraper/scraper.go
The-9880: >the `unprefixed system label collision` test even asserts a user `job=user-job` overrides the system `job` on execution metrics, the opposite of the "silently ignored"

>But for per-series intrinsic metric labels (le, method, phase, status, result, name, url, ...) it is destructive.

The reason this is considered "acceptable" here is because when users move away from the `PREFIXED` state, none of these initial `systemLabels` can be in their custom labels due to API validations (and the API will prevent them from doing that going forward).

So in effect, even though the test says `job=user-job` will win that case will never actually occur for the `job` label.

The **per-series intrinsic metric labels** are part of the initial set of `systemLabels` we defined, so they will be protected too.

So the reason we aren't calling `IsSystemLabel` is to prioritize keeping the user's existing functionality stable when future changes threaten to break that.

We could later break out the _intrinsic metric labels_ as a separate list of reserved labels and make the order of priority `intrinsic metric labels` -> `user-defined labels` -> `system-defined labels` to be more explicit about that.

internal/scraper/scraper.go
The-9880: I lean towards `PREFIXED` for 2 reasons:
1. It's the state most tenants are in today
2. Agent v1's breaking changes will drop support for the legacy prefix altogether, operating on the assumption of `UNPREFIXED` tenants going forward.

Given that we're pushing for (2) soon, I think it makes sense to assume `PREFIXED` as a default for pre-v1 releases.

internal/labelmode/labelmode.go
The-9880: I've added this comment in the latest commit 👍

internal/scraper/scraper.go
mem: I'm not sure I follow that reasoning.

If the API is rejecting the system labels, where's the problem with validating that assumption on the agent side? When a check is received, we run `Validate()` on the settings, because we don't simply assume the API did it already. In the same sense, per your explanation, if we were to drop system labels from the label set present in the check when the check is received, we would drop nothing, because none of those labels should be there. We wouldn't be assuming that the API rejected those labels, we would be asserting it.

internal/scraper/scraper.go
mem: But your own previous reasoning is that we are never going to hit line 527.

1. If the tenant has migrated, falling back to `PREFIXED` is actually harmful (queries depending on the new behavior fail, labels disappear, ...) because the user already assumes the new state.
2. If the tenant has not migrated yet, falling back to `UNPREFIXED` will modify the emitted metrics and queries _might_ fail (joins have more labels than expected, etc), but since the user has not done any work on their side that assumes this state, when the issue clears, it goes back to the previous state, so the issue is transient.

internal/scraper/scraper.go
mem: With the extra benefit that if we do drop labels, we can emit a log line about that and alert on it.

The-9880: This part:
> if we were to drop system labels from the label set present in the check when the check is received, we would drop nothing, because none of those labels should be there

This is only fine for the current, initial set of `systemLabels`.

The problem is if we introduce a new addition to `systemLabels` in the future: we agreed that we will keep running the tenant's check as it was - i.e. without overwriting or dropping their current labels - and instead use other signals to alert them the check needs to be updated.

That's why this is written in a user-wins fashion, with the understanding that the user will never contend the initial `systemLabels` set. It is because future system labels _might_ contend with their label set, in which case we want to preserve any functionality relying on their existing label schema.

The-9880: Would it be clearer if I omit the definitions for `IsSystemLabel` and `systemLabels` from the agent and maintain them in the API? The agent can follow the user-wins approach with comments making the decision explicit.

internal/scraper/scraper.go
The-9880: > But your own previous reasoning is that we are never going to hit line 527

Yes, it's why I think this point is essentially moot.

> If the tenant has not migrated yet, falling back to UNPREFIXED will modify the emitted metrics and queries

Ah, I take your point - fair enough. The issue should be transient either way, but I think this breaks fewer things.

The-9880: Addressed in https://github.com/grafana/synthetic-monitoring-agent/pull/1879/commits/a0a58535e55b275159c05cc2a60151fdcd9bfd72

internal/scraper/scraper.go
mem: I can see we are not going to agree on anything on this path.

I have added a couple of tests. Make them pass by changing the code (not the tests, obviously) in a reasonable way.

internal/scraper/scraper.go
mem: Thank you.

internal/scraper/scraper.go
The-9880: I understand and am aligned that the intrinsic metric labels need to be protected.

For now, we can implement a blanket system-wins policy on the basis that there can be no overlap _today_.

When we eventually do gain another system label (e.g. update in BBE, new check type introduced, etc.), then the following priority must take place for collisions:
1. Intrinsic metric labels win first
2. User-defined labels win second
3. System-defined labels win third

And that will need to be implemented.

Does that sound reasonable?

mem: Since there's nothing in the code that makes the distinction between intrinsic metric labels and system labels today, and there's no validation on the agent side that rejects either, I cannot say whether that's reasonable or not. Furthermore, I'm looking at this with skepticism because the bit of code that is required to protect intrinsic labels from being overwritten is trivial, yet it's not there (and I realize now that `IsSystemLabel` should not have been exported, that logic should have been part of `Validate`, and that would have saved us this discussion, but that's besides the point).

So, as I said before, make those tests pass with some reasonable code change and we are good.

The-9880: Addressed in https://github.com/grafana/synthetic-monitoring-agent/pull/1879/changes/2b2a6c37c7c45465cf223630a5b1bbcd09c80617

mem (approved):
Thank you!

PR #1891, opened 2026-07-04
feat: Gate protocol secrets with the protocol-secrets feature

mem:
Protocol-secrets support was enabled through a bespoke environment variable, SM_AGENT_ENABLE_PROTOCOL_SECRETS, read via a one-off boolFromEnv helper. The agent already has a first-class feature-flag mechanism (the -features flag backed by feature.Collection), so use it: add a "protocol-secrets" feature and drive config.EnableProtocolSecrets from features.IsSet(feature.ProtocolSecrets).

The downstream wiring is unchanged: config.EnableProtocolSecrets still feeds SupportsProtocolSecrets into the checks updater and adhoc handler, which advertise it in ProbeInfo at registration.

Remove the now-unused environment variable and the boolFromEnv helper. The feature is pre-release (only ever set on dev probes), so there is no backward-compatible fallback. Enable it with `-features protocol-secrets`.

### DIFF
--- a/cmd/synthetic-monitoring-agent/main.go
+++ b/cmd/synthetic-monitoring-agent/main.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strconv"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


mem:
Part of https://github.com/grafana/synthetic-monitoring/issues/570

The-9880 (approved):

PR #1904, opened 2026-07-15
fix: Enable WSL linter

mem:
I'm starting to get tired of LLM-generated code smushing together code because to it it's basically the same. At the same time I don't want to be telling people to add blank lines all over the place because that gets annoying fast. So reach out for the best available alternative: have a linter check that formatting is consistent and emit diagnostics if it's not.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -40,6 +40,7 @@ linters:
     - wastedassign
     - whitespace
     - zerologlint
+    - wsl_v5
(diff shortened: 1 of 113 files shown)


The-9880 (approved):

ckbedwell (approved):
I LOVE BLANK LINES ❤️  (I wonder if my approval is good enough)

PR #1906, opened 2026-07-17
feat: expose single-execution telemetry collector

ckbedwell:
## Summary

- add a public `pkg/collector` adapter for running one caller-supplied probe through the agent's normal metric and log transformation pipeline
- expose non-publishing collection from the scraper while keeping execution ID generation internal to the agent
- thread a logical event timestamp through check start/end logs while continuing to measure real wall-clock execution time
- document the unscheduled collector path and its relationship to the existing scraper

## What this enables

This is the agent-side seam used by [grafana/dem-dev#43](https://github.com/grafana/dem-dev/pull/43).

That PR adds a deterministic Synthetic Monitoring scenario and historical-data generator. It supplies synthetic probers to this collector, asks the agent pipeline to collect at historical event times, and receives the same Prometheus time series and Loki streams that a normal agent execution would produce—without starting the scheduler or publishing live telemetry.

Keeping the transformation in the agent means the generator does not need to duplicate metric naming, label shaping, execution metadata, log encoding, or failed-probe behavior. Execution IDs also remain owned by the agent, matching production semantics.

For the normal scheduled path, callers continue to pass the current scrape time, so existing agent behavior is preserved.

## Validation

- Docker-backed Linux tests:
  - `pkg/collector`
  - `internal/scraper`
  - 90 tests passed, 6 expected capability/generator skips
  - 81.2% aggregate coverage for the affected run
- targeted `golangci-lint`: 0 issues
- `git diff --check`

## Related PR

- Enables [grafana/dem-dev#43 — deterministic scenario history generator](https://github.com/grafana/dem-dev/pull/43)


mem (dismissed):
LGTM

Pokom (changes requested):
I think there's an opportunity to refactor this in a way that enables the ability for an arbitrary timestamp to be added while not overloading the existing `collectData` method. My main concern with the approach in the PR is maintaining the code going forward. Itt requires someone to be aware of the dem-dev(think external contributors to an OSS project) changes to understand why all of the branching was introduced and how to work around it. I put together a quick POC(https://github.com/grafana/synthetic-monitoring-agent/pull/1918/changes) that is intended to be fed as context that introduces a stamper interface that allows external programs to provide an implementation that handles how to handle the timestamp. This simplifies the logic in the collectData method while providing a mechanism to introduce other ways of generating seeded data.

internal/scraper/scraper.go
```diff
@@ -521,7 +521,77 @@ func tickWithOffset(
...
+}
+
+func (l logicalStamper) endLogFields(duration float64) []any {
+	eventEnd := l.eventTime.Add(time.Duration(duration * float64(time.Second)))
+	return []any{"time", eventEnd.UTC().Format(time.RFC3339Nano)}
+}
+
+func (logicalStamper) fallbackDuration(_, wallStart time.Time) time.Duration {
+	return time.Since(wallStart)
+}
```
Pokom: My PR wasn't really in shape to be merged in as is, and was hoping to see it used it as a basis in another session as the input context as there's some code smells with it. I have two suggestions
* Move the interface and implementations to a dedicated file, such as `stamper.go`. and add a corresponding set of unit tests
* Update the interface so that both stamper's have the same fallbackDuration method call

internal/scraper/scraper.go
```diff
@@ -521,7 +521,77 @@ func tickWithOffset(
...
+	// extractLogs turns that field into each Loki entry's timestamp, so it
+	// governs where logs without their own `time` field land.
+	wrapperLogTimestamp() kitlog.Valuer
+	// endLogFields returns any extra fields appended to the terminal
+	// (succeeded/failed) wrapper log, e.g. an event-end `time` that overrides
+	// `ts` for that line.
+	endLogFields(duration float64) []any
+	// fallbackDuration computes the probe duration when the prober reports
+	// none, given the scrape's nominal time and the wall-clock start.
+	fallbackDuration(scheduledAt, wallStart time.Time) time.Duration
```
Pokom: Suggestion: change this method interface to simply be 
```suggestion
	fallbackDuration(wallStart time.Time) time.Duration
```

The methods would change to 

```go
func (w wallClockStamper) fallbackDuration(time.Time) time.Duration {
      return time.Since(w.scheduledAt)
}

func (logicalStamper) fallbackDuration(wallStart time.Time) time.Duration {
      return time.Since(wallStart)
}
```

And the future calls need to be updated.

Pokom (commented):
We're pretty close. I left a few suggestions on how to clean up the code just a bit. Having the implementation in `scraper.go` is the blocker for me at this point as the intent for the interface is for `scraper.go` to not need to be aware of the underlying implementation. If you hoist the interface and implementation out and refactor the interface slightly, I think I'm good to see this merged in. 

And as always, naming is hard(:tm:). I think the interface could be generalized to be named `scrapeClock`, but that's nit picking at this point.

internal/scraper/scraper.go
```diff
@@ -836,11 +874,21 @@ func runProber(
...
 	} else {
 		probeSuccessGauge.Set(0)
 
-		_ = level.Error(logger).Log("msg", "Check failed", "duration_seconds", duration, "walltime_seconds", wallDuration)
+		_ = level.Error(logger).Log(
+			"msg", "Check failed",
+			"time", eventEnd.UTC().Format(time.RFC3339Nano),
+			"duration_seconds", duration,
+			"walltime_seconds", wallDuration,
+		)
```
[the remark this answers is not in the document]
ckbedwell: Doesn't apply to the scheduled path -- `wallClock.endLogFields` returns nil, so nothing writes a `time` field and the terminal log keeps its wall-clock `ts`, in order with the probe's own lines. The event-relative `time` only comes from `logicalClock`, and that's only ever selected by `CollectData`. `TestHTTPCheckLogTimestampsAreNonDecreasing` covers it.

internal/scraper/scraper.go
ckbedwell: Thanks -- both in. `scrapeClock` and both implementations now live in `internal/scraper/scrape_clock.go` with unit tests per policy, so `scraper.go` doesn't know they exist, and `fallbackDuration` is down to just `wallStart`. Took the `scrapeClock` name while I was there. The only comment I kept is on `wallClock.fallbackDuration` ignoring that argument -- measuring from the tick is the existing scheduled behaviour, and it's not obvious from a method that takes a `time.Time` and never reads it.

Also fixed both Bugbot findings: `Collector.Close()` cancels the context we hand the scraper, and `WithLabelMode` replaces the hard-coded unprefixed mode (defaults to `LABEL_MODE_PREFIXED` so user labels match a normal scrape). dem-dev#43 closes the generators.

Pokom (approved):

PR #1922, opened 2026-07-24
feat: emit region as sm_region for tenants in label migration

The-9880:
Relates to https://github.com/grafana/synthetic-monitoring/issues/549

Gives the `region` label back to users who perform the `label_`-prefix migration - `region` can be highly context/user-specific, so distinguish the SM-defined region with a `sm_` prefix (`sm_region`).
* Ensure `sm_region` is always present on `sm_check_info` - defaults to `unset` if empty.

Note that this has ramifications on per-check alerting expressions, which current exclude the `region` label and must start excluding `sm_region` instead. A corresponding change in the API must update the alerting rule expressions for tenants who switch away from the initial `PREFIXED` state.

For tenants that have started the label migration (label mode DUAL_WRITE or UNPREFIXED), the agent now emits its region label as `sm_region`.

### DIFF
--- a/docs/architecture/scraper.md
+++ b/docs/architecture/scraper.md
@@ -121,7 +121,7 @@ you need to change republishing behaviour.
 
 For each probe run:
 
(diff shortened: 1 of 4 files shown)


mem (approved):
This makes sense to me.

PR #1926, opened 2026-07-27
feat: Export k6 grace time metric

### LINKED ISSUE #987: Export grace time as a metric so that it can be used in dashboards
From the comment in the code review:

              This will have to wait for another time, but you can create a metric with that information (grace time), and you can use that metric in a dashboard to show where the indicator is crossing that value instead of refering to a particular line in the source code.

_Originally posted by @mem in https://github.com/grafana/deployment_tools/pull/182269#discussion_r1834369200_

Guflly:
Export the k6 grace-time values from the agent's Prometheus registry:

- `sm_adhoc_k6_grace_time_seconds` for the ad-hoc outer timeout
- `sm_agent_k6runner_grace_time_seconds` for the remote k6 runner used by scheduled browser, scripted, and multi-HTTP checks

This lets dashboards show the effective timeout threshold without duplicating source constants.

Tests cover metric registration and values.

Closes #987.

### DIFF
--- a/internal/adhoc/adhoc.go
+++ b/internal/adhoc/adhoc.go
@@ -85,7 +85,8 @@ type apiInfo struct {
 }
 
 type metrics struct {
-	opsCounter *prometheus.CounterVec
+	opsCounter  *prometheus.CounterVec
+	k6GraceTime prometheus.Gauge
 }
 
 // Backoffer defines an interface to provide backoff durations.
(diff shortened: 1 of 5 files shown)


Pokom:
Thanks for PR @Guflly! This seems like a reasonable implementation and something we'd like to support. Before moving forward, can you sign the CLA so the rest of our CI pipelines can execute.

Guflly:
Signed. The CLA check is green now; could you approve the pending validation workflow? Thanks.

Pokom:
@Guflly I've approved the workflow and there are some linter's failing, let's address those.

I did have a few clarifying questions. Is there a specific use case for the metrics that you're addressing? https://github.com/grafana/synthetic-monitoring-agent/issues/987 is fairly limited in it's context, so I'm curious if you're actively running into a situation that requires the metric. The reason I ask is this PR is specifically for ad-hoc checks, which is fairly limited in terms of usage. Would you consider extending this to the other types of checks as well, which I think has a higher value?

Guflly:
I picked this up from #987 rather than an active deployment issue. The goal is to show the timeout threshold in dashboards without copying a source constant. I've extended it to the remote k6 runner too, covering scheduled browser, scripted, and multi-HTTP checks. The ad-hoc gauge stays separate because it is a different outer timeout. The linter failure is fixed.

ka3de (approved):
Thank you.

PR #1974, opened 2026-08-31
fix: Redact check credentials in logs

### LINKED ISSUE #508: Security issue - check basic auth credentials written in plain text in log files
I created a check in an on-prem agent.  That check uses basic auth in the configuration.

In the log files of the synthetic-monitoring agent pod, the username and password are written out in plain text rather than being redacted.

Plain text passwords should never be written out, even in debug mode.

`{"level":"debug","program":"synthetic-monitoring-agent","subsystem":"updater","check change":{"check":{"id":1565289,"tenantId":00000,"frequency":120000,"offset":0,"timeout":10000,"enabled":true,"labels":null,"settings":{"http":{"ipVersion":"V4","method":"GET","noFollowRedirects":false,"tlsConfig":{"insecureSkipVerify":true},"basicAuth":{"username":"grafana","password":"randompassowrdinplaintext"},"failIfSSL":false,"failIfNotSSL":false,"validStatusCodes":[200],"failIfBodyNotMatchesRegexp":[".*Showing.*"]}},"probes":[7193],"target":"https://intranet.domain.blah/alerts","job":"Intranet - Alerts","basicMetricsOnly":true,"alertSensitivity":"none","created":1692822282.9243417,"modified":1694447326.0806525}},"time":1694447528688,"caller":"github.com/grafana/synthetic-monitoring-agent/internal/checks/checks.go:669","message":"got check change"}            `

digitalcrab:
Implement MarshalZerologObject interface for check and tenant types, so that they do not leak the secret information into logs.

Small change: wrap one debug line in the API server so it also gets the structure with implemented interface.

Fixes #508

### DIFF
--- a/pkg/pb/synthetic_monitoring/checks_extra.go
+++ b/pkg/pb/synthetic_monitoring/checks_extra.go
@@ -324,36 +324,48 @@ func CheckTypeFromString(in string) (CheckType, bool) {
 }
 
 func (c Check) Type() CheckType {
+	checkType, found := c.Settings.checkType()
+	if !found {
+		panic("unhandled check type")
+	}
+
+	return checkType
+}
+
+// checkType reports the type of check the settings describe. Unlike
+// Check.Type it does not panic on settings that have nothing set, which is
+// what a check attached to a delete operation looks like.
+func (s CheckSettings) checkType() (CheckType, bool) {
 	switch {
-	case c.Settings.Dns != nil:
-		return CheckTypeDns
+	case s.Dns != nil:
+		return CheckTypeDns, true
 
-	case c.Settings.Http != nil:
-		return CheckTypeHttp
+	case s.Http != nil:
+		return CheckTypeHttp, true
 
-	case c.Settings.Ping != nil:
-		return CheckTypePing
+	case s.Ping != nil:
+		return CheckTypePing, true
 
-	case c.Settings.Tcp != nil:
-		return CheckTypeTcp
+	case s.Tcp != nil:
+		return CheckTypeTcp, true
 
-	case c.Settings.Traceroute != nil:
-		return CheckTypeTraceroute
+	case s.Traceroute != nil:
+		return CheckTypeTraceroute, true
 
-	case c.Settings.Scripted != nil:
-		return CheckTypeScripted
+	case s.Scripted != nil:
+		return CheckTypeScripted, true
 
-	case c.Settings.Multihttp != nil:
-		return CheckTypeMultiHttp
+	case s.Multihttp != nil:
+		return CheckTypeMultiHttp, true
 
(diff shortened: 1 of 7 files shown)


pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -324,36 +322,48 @@ func CheckTypeFromString(in string) (CheckType, bool) {
...
 
-	case c.Settings.Browser != nil:
-		return CheckTypeBrowser
+	case s.Browser != nil:
+		return CheckTypeBrowser, true
 
 	default:
-		panic("unhandled check type")
+		return 0, false
 	}
```
Pokom, line 369: I don't know the answer to this: Does this change any expectations for downstream uses?

Pokom: Heh, answered my own question. Looks like the panic is handled in the `Type()` method.

pkg/pb/synthetic_monitoring/checks_extra.go
```diff
@@ -1697,20 +1682,3 @@ func GetCheckInstance(checkType CheckType) Check {
...
-
-func (ri RemoteInfo) MarshalJSON() ([]byte, error) {
-	type T RemoteInfo
-
-	tmp := T(ri)
-
-	tmp.Password = `<encrypted>`
-
-	return json.Marshal(tmp)
-}
```
Pokom, line 1716: What's the intent of removing the MarshalJSON method? My understanding is that removing this would leak the password over the wire when the api makes requests for RemoteInfo.

Pokom (commented):
I have one question about removing `RemoteInfo`'s  `MarshalJSON` method, which I believe could leak the password within the API codebase. Is that an intentional removal?

pkg/pb/synthetic_monitoring/checks_extra.go
Pokom: IE, the api pulls in RemoteInfo and calls json.Marshall on it, and not the Password is potentially leaked into logs or payloads. You can validate this behavior by pulling the change into the API(go get $$$) and then running tests. There will be failures, and it would be easy to update the test to pass without realizing that passwords could be leaked.

digitalcrab: Good catch, thank you. You were right, and I was not aware of all the parts.

Pokom: All good. I think the downstream affect is much harder to be aware of. CI in the api would have eventually caught it, but it wouldn't have been obvious the impact.

Pokom (approved):

pauldenisevich:
Hi Max :)

PR #1981, opened 2026-09-09
fix: Use the configured port for DNS checks

digitalcrab:
## What I was doing

Building a local fake DNS server to test DNS checks against, running it on a non-standard port.

## How I found it

The check failed straight away and the log of the fake server stayed empty, so nothing ever reached it. Network path and check configuration were both fine, which left the agent: the port from the check settings was never used, so the query always went to the default port.

## The fix

The prober now builds the server address from the server and the port together. An unset port keeps its current meaning, so existing checks are not affected, and tests cover an explicit port, an IPv6 server and no port at all. I also left a TODO where DNS over TLS would be wired up, as it is not reachable from the check settings today.

Worth keeping in mind on rollout: a check configured with a non-default port has been quietly querying the default one and may look healthy today. After this it queries what it asked for, so it can start failing - the correct result, just no longer a silent one.

### DIFF
--- a/internal/prober/dns/dns.go
+++ b/internal/prober/dns/dns.go
@@ -3,6 +3,8 @@ package dns
 import (
 	"context"
 	"errors"
+	"net"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


Pokom (approved):

Pokom:
I was hoping to assess the impact of this bug, but the DNS configuration data does not propagate to our analytics systems. The fix appears benign enough and should only impact customers that had DNS checks setup with custom ports defined. What I'm not sure about is what will happen to probers that are executing and are (luckily) working. Customers in that scenario potentially could start seeing checks failing and will need to fix their checks.

PR #1991, opened 2026-09-14
fix: Spread check start times over the whole frequency

digitalcrab:
## Problem

Checks wait a random delay before their first run, so that checks sharing a frequency do not all start at once. That delay was capped at `maxPublishInterval` (2m) - a constant about republishing stale data, not about scheduling.

So every check running less often than every 2m started inside the same 2m window, and kept that slot until its scraper restarted.

Example: 1000 checks on a 10m frequency put every start in the first fifth of the period.

```
starts per 2m slice of a 10m period
before   [1000    0    0    0    0]   5.00x the average
after    [ 192  168  170  193  160]   1.09x the average
```

## Fix

Draw the delay from the full frequency instead of the cap. `maxPublishInterval` keeps its real job, which is the republish interval passed to `tickWithOffset`.

## Trade-off

A first sample can now arrive up to one frequency after the scraper starts, where before it arrived within 2m. If that matters for long frequencies, the alternative is to scrape once immediately and then settle onto the random phase, which means changing `tickWithOffset`.

## Related

Ref: https://github.com/grafana/support-escalations/issues/24134


The-9880 (commented):
Hey, let's discuss this further - you can see some current discussion about this [here](https://docs.google.com/document/d/1ZAjcAuYeVPJKUWsBHVbQuhuXETndWcVY0cAQfXBtqwc/edit?disco=AAACDhppfOU).

We can't simply increase the interval - the `2m` limitation is in order to keep the timeseries related to the check alive.

Keeping the cadence ongoing in the situation of agent restarts, etc. requires some state persistence across those boundaries - e.g. caching.

I have a couple of branches for this @digitalcrab, to set up and use caching for this goal - if you have bandwidth to push on this, let's sync and do a handoff. I have a design doc to this effect as well which I can hand over.

digitalcrab:
Closing this.

The `2m` cap I removed is not a leftover constant. It is there to keep a check's metrics alive, and removing it breaks that. @The-9880 spelled this out in the review above.

A scraper keeps no state of its own. When a check moves to a different agent instance - during scaling, a restart or a deploy - the last result goes away with the old instance, so nothing is published again until the check actually runs. A series that gets no new data point for `5m` is treated as stale, and alerting and recording rules stop evaluating it correctly.

The `2m` cap means a check always runs within 2m of landing on a new instance, which leaves about 1m to run and publish before that 5m limit. Take the delay from the full frequency instead and the guarantee is gone: any check that runs less often than every 2m can stay silent for longer, and the longer the frequency, the longer the silence.

The problem in the description above is still real. Every check slower than 2m gets a start time inside the same 2m window and keeps that slot until its scraper restarts - on a 10m frequency that puts every start in the first fifth of the period.

So this is the right fix in the wrong order. One thing has to come first: the agent needs to remember the last result and the last run time of each check, so a check that moves to another instance continues where it stopped. That work is already in progress, and the spread fix belongs on top of it.

The branch stays as is. `initialOffset` and `TestInitialOffsetSpread` are reusable once that lands.

PR #1992, opened 2026-09-15
fix: Kill the whole k6 process group on cancel

digitalcrab:
## Problem

`exec.CommandContext` signals k6 alone on timeout or shutdown. Processes k6 started, such as a Chromium tree, keep running and PID 1 adopts them.

## Solution

`Setpgid` puts k6 in its own group, `cmd.Cancel` kills the group, `WaitDelay` stops a stuck child from blocking the agent.

k6 no longer gets signals sent to the agent process group. Nothing depends on that: systemd signals the cgroup, everything else the main PID.

## Testing

Two new tests, both red without the change.

Real Chromium in the browser image: a wedged browser leaked 10 processes, now 0. A healthy browser exits on its own either way, so this is hardening, not an incident fix.


internal/k6runner/local_orphan_unix_test.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// TestLocalRunStopsProcessTreeOnTimeout checks that a timed out check leaves no
+// processes behind. Without Setpgid and cmd.Cancel, exec.CommandContext signals k6
+// alone and PID 1 adopts whatever k6 started. The fake acts as k6: it starts a
+// grandchild, then waits past the timeout.
+func TestLocalRunStopsProcessTreeOnTimeout(t *testing.T) {
+	runner, pidFile := newSpawnerRunner(t)
+
+	const checkTimeout = 1000 * time.Millisecond
```
ka3de, line 26: nit: Why 1s check timeout? If I'm understanding this correctly, we will effectively wait for this TO first as part of `runner.Run()`, and then perform the verification. During that time we expect the `k6-fake-spawner` script to be executed at least until `echo $! > "ORPHAN_PIDFILE"` , which we'll use for verification.
Are we accounting for possible [worse] CI performance?

ka3de (approved):

internal/k6runner/local_orphan_unix_test.go
digitalcrab: 1 sec here is just to be sure we timeout script ourselves (it will run for 3 sec + children 10 sec). I think 1 sec here is already overkill, I hope we not gonna have CI that won't be able to run simple bash command in less than 1 sec ;)

ka3de: > I think 1 sec here is already overkill

Yeah, not critical, but this was more my intention, on whether this could be reduced since it can not run under `t.Parallel()` due to `t.Setenv()` usage.

PR #1995, opened 2026-09-16
feat: Enable protocol secrets by default

digitalcrab:
The `protocol-secrets` feature flag was a transitional gate. The feature is released in all environments and every proxy advertises the capability, so the agent now always reports `SupportsProtocolSecrets` at registration. Passing the flag logs the deprecation notice.

The flag only ever controlled advertising. Secret resolution is gated per check by `secret_manager_enabled`, so a check that did not opt in is not affected.

Closes grafana/synthetic-monitoring#631


mem:
> The protocol-secrets feature flag was a transitional gate. The feature is released in all environments and every proxy advertises the capability, so the agent now always reports SupportsProtocolSecrets at registration.

For public probes (the ones we operate).

What about private probes?

digitalcrab:
@mem private old probes keep running without secrets support, the ones who updates will have secrets and will report capability, same as with "traceroute" and "k6". So old probes still won/t be able to run checks with secrets because API verifies what they advertise. 

Not sure if that answers the question, as the question is too broad, please specify what exactly would you like to know about private probes

PR #1996, opened 2026-09-17
fix: Run the full Go test suite on macOS

digitalcrab:
## Why

The goal was to be able to run the full Go test suite on a Mac. That was not possible: two packages
failed for reasons unrelated to the code under test.

`internal/scraper` did not build at all - `scraper_test.go` imported libcap, which is Linux-only,
and that took down all 45 tests in the package for the sake of one capability probe used by one of
them. Separately, `TestSettingsToScript` needs an sm-k6 binary, and grafana/xk6-sm publishes
linux-amd64 and linux-arm64 only, so `make sm-k6-native` cannot satisfy it on darwin.

## What changed

The capability probe moved into a Linux-only file with a `!linux` stub that skips, and
`testhelper.K6Paths` now skips on darwin when no sm-k6 binary is present.

Linux behaviour is unchanged in both cases. A missing k6 binary there still fails, because there it
means the download step did not run, and that is worth catching.

## Result

`go test ./...` on darwin now exits 0 - 262 passing tests, up from 220, and 9 skips. Almost the full
suite: the traceroute subtests and the k6 execution half of `TestSettingsToScript` cannot run on a
Mac, and they now skip with a reason instead of failing.

## How to test

```
go test ./...
GOOS=linux GOARCH=arm64 CGO_ENABLED=1 ./scripts/docker-run go test -race ./...
```

Both are green, and `golangci-lint` reports 0 issues for `GOOS=linux` and `GOOS=darwin`.


Pokom (commented):
I do not have a mac, I do not think I can test this fully. @The-9880 do you run MacOS?

The-9880 (approved):

PR #1997, opened 2026-09-17
test: Harden interpolation coverage for HTTP and multihttp checks

digitalcrab:
## Why

I am going to bring secrets to more check types and to more fields. Before I start moving things around, I want the current interpolation behaviour pinned down, so a regression shows up as a failing test rather than as a credential that silently goes out wrong.

No production code is touched here.

## What

**`internal/prober/http/interpolation_wire_test.go`** - goes through `NewProber` and `Probe` against a recording target, and asserts the exact credential bytes that reach it for every combination of credential value and `secretManagerEnabled`. The existing tests cover `resolveSecretValue` and `buildPrometheusHTTPClientConfig` in isolation, so the wiring between them was untested. Covers resolution, passthrough, references embedded in larger values, malformed and nested placeholders, values holding a literal `$`, a reference in a custom header, and the failure paths - where it also asserts that nothing reaches the target at all.

The three TLS fields are covered by reading back the file `tls.SMtoProm` writes. `TestBuildTLSConfig_WithSecrets` asserts that a path was produced but not what landed in it.

**`internal/prober/multihttp/script_golden_test.go`** plus `testdata/*.js` - pins the generated k6 script for a check using `${variable}` interpolation. A variable has no value when the script is built, so there is nothing substituted to observe and the script itself is the artifact. Four fixtures: every site that expands a variable, every extraction type, the JavaScript escaping, and the login-then-use-the-token shape the feature exists for.

## How to test

```
go test ./internal/prober/http/ -run TestInterpolation
go test ./internal/prober/multihttp/ -run TestScriptGolden
```

Regenerate the golden scripts with `-update` on the second one, and please read the diff before committing it.


internal/prober/multihttp/testdata/escaping.js
```diff
@@ -0,0 +1,108 @@
...
+	body=body.replaceAll('${tenant}', vars['tenant']);
+	
+	response = http.request('GET', url.toString(), body, {
+		// TODO(mem): build params out of options for the check
+		tags: {
+		  name: '0', // TODO(mem): give the user some control over this?
+		  __raw_url__: 'http://example.org/héllo/${tenant}',
+		},
+		redirects: 0,
+		headers: {'Content-Type':"text/plain","X-Script":'\u003C/script\u003E '+vars['tenant'],"X-Emoji":'emoji 🎉 '+vars['tenant'],"X-Hyphen":'${my-var}',"X-Quote":'it\'s a '+vars['tenant'],"X-Backslash":'back\\slash '+vars['tenant']}
```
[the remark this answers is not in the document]
digitalcrab: Not valid, the test passes.

I ran it on `f05bcbf1`, the same commit this review is against:

```
--- PASS: TestScriptGolden/escaping (0.00s)
```

`validate` is green on `ubuntu-latest` and on `github-hosted-ubuntu-arm64` as well, so this is not a local-only pass.

The mechanism in the report is also not what `text/template.JSEscape` does. For a rune at or above `utf8.RuneSelf` it decodes the rune and writes the original bytes through untouched when `unicode.IsPrint(r)` holds, and only otherwise emits `\uXXXX` with the codepoint. No UTF-16 surrogate pairs anywhere in it:

```
"héllo"      -> "héllo"
"emoji 🎉 "  -> "emoji 🎉 "
"</script> " -> "\u003C/script\u003E "
```

That description fits the `html/template` escaper, not the `text/template` one that `script.go:45` calls.

One part of this does land, for a different reason than the one given. My comment says JSEscape "passes multi-byte UTF-8 through unchanged", and that is broader than the truth - non-printable runes do get escaped, `\u00A0` for a non-breaking space, `\u200B` for a zero width space. The missing word there is "printable".

The fixture itself is correct, nothing to regenerate.

internal/prober/multihttp/testdata/escaping.js
```diff
@@ -0,0 +1,108 @@
+import http from 'k6/http';
+import { check, fail } from 'k6';
+import { test } from 'k6/execution';
+// TODO(mem): conditionally import these modules
```
mem, line 4: are you giving me tasks?

mem: oh, I see, you took this from the multihttp generated scripts...

mem (approved):

